PR #2, opened 2021-09-18
Fix documentation of --utf-force in README and --help.

### LINKED ISSUE #1: [BUG] Help/docs reference to non-existence --utf-foce option
**Describe the bug**

The documentation (README.md) and --help output both reference an option `--utf-foce` ("force start even if no UTF-8 locale was detected").

But this option doesn't exist.  The actual option is `--utf-force`.

**To Reproduce**

Run `btop --utf-foce` and notice you get

    kde@midna ~/tmp/btop $ btop --utf-foce
     Unknown argument: --utf-foce
     Use -h or --help for help.

**Expected behavior**

I expected btop to start up without issue (as it did when using `--utf-force`). I didn't not test on a non-UTF-8 locale yet but I would expect the option to work even in that case.

 - btop++ version: btop version: 1.0.0   (*PS* recommend adjusting the template here to read `btop -v` not `bpytop -v`)
 - Architecture: x86_64
 - Platform: Linux
 - (Linux) Kernel: 5.14.0
 - Terminal used: Konsole 21.11.70
 - Font used: JetBrains Mono

**Additional context**

Let me know if you actually need this

purinchu:
This seems like a simple enough issue that I figured I could submit a patch instead of just reporting it.

Just in case I did run `make` and verify that the new binary runs properly and outputs correct `--help` output as well (great job on making it single Makefile without needing CMake or anything else btw).

Fixes #1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,7 +476,7 @@ log_level = "DEBUG"
 #### Command line options
 
 ```text
-usage: btop [-h] [-v] [-/+t] [--utf-foce] [--debug]
+usage: btop [-h] [-v] [-/+t] [--utf-force] [--debug]
 
 optional arguments:
   -h, --help            show this help message and exit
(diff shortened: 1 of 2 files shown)


aristocratos:
Oops, noticed I missed a couple of other things too.

> great job on making it single Makefile without needing CMake or anything else btw)

Thanks :) Wanted to make it as easy possible to compile (probably spent way too much time making it look pretty though...)

Thansk for the fix!

PR #17, opened 2021-09-21
Add make install missing reset escape sequence.

### LINKED ISSUE #16: [BUG] Terminal keeps text formatting after make install
**Describe the bug**

After install, the terminal keeps the text setting of white/bold

**To Reproduce**

I downloaded the linux x86-64 tbz file - 

```
mkdir btop
wget https://github.com/aristocratos/btop/releases/download/v1.0.4/btop-1.0.4-linux-x86-64.tbz
PREFIX=test make install
```

My term is still bold white.

**Expected behavior**

I expect my term to reset.

I looked in the makefile and it's missing a `\033[0m` on the last printf statement in the install.

**Screenshots**

First run is before makefile modification, second is after.

image

davised:
Fixes #16 

I see you must make each release independently, but maybe this helps a little.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,7 @@ install:
 	@printf "\033[1;92mInstalling doc to: \033[1;97m$(DESTDIR)$(PREFIX)/share/btop\n"
 	@mkdir -p $(DESTDIR)$(PREFIX)/share/btop
 	@cp -p README.md $(DESTDIR)$(PREFIX)/share/btop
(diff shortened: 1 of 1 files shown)


aristocratos:
@davised 
> I see you must make each release independently, but maybe this helps a little.

I have script that handles all version bumping, changelog, compiling and file compression for the different architectures, takes about 1 minute and a half to run :)

PR #23, opened 2021-09-23
Ignore hidden directories like .vscode

CoryM:
Set git to ignore linux hidden directories (.vscode, .config, ...) while still allowing hidden files (.gitignore, ...) to be tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@
 build
 bin
 btop
+.*/


aristocratos:
@CoryM 
Have a bunch of stuff in `.git/info/exclude` should probably move over some of the more common ones.

PR #42, opened 2021-09-26
Snap is published and running

kz6fittycent:
This PR is just for the sake of keeping the code in one house. I can continue to maintain the snap but it would make sense to me for you to keep this in your repo. I can do more PR's to the `.yaml` as needed so things stay up to date between our repos. 

Thanks for you help getting this going!


aristocratos:
Sure :)

PR #44, opened 2021-09-26
Update snapcraft.yaml

kz6fittycent:
Set `STATIC=true`


aristocratos:
Just pushed the same :P

PR #68, opened 2021-10-03
macOS support

joske:
Hi,

I've started implementing macOS support. It's not complete at all, I just create this PR to raise awareness, solicit feedback and discussion. If you think this is noise or not appropriate, please delete the PR. 

You have to use gcc 11 from homebrew to compile, as macOS clang does not support a few of the c++20 features you use. Make sure it finds c++-11 as c++ in your PATH (I've symlinked it to c++ in /usr/local/bin and put /usr/local/bin at start of PATH).

Right now it shows some memory info, CPU name, processes (but no sorting etc) so lots to do still.

I'm not a c++ (and for sure not in c++20) nor macOS expert, so please comment kindly ;-).

I know the files are formatted wrongly (vscode defaults). Is there a simple way I can import a style file (like in eclipse) to have your code style (which I prefer ;-)).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,8 @@ OBJEXT		:= o
 #? Flags, Libraries and Includes
 override REQFLAGS   := -std=c++20
 WARNFLAGS			:= -Wall -Wextra -pedantic
-OPTFLAGS			?= -O2 -ftree-loop-vectorize -flto=$(THREADS)
-LDCXXFLAGS			:= -pthread -D_FORTIFY_SOURCE=2 -D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector -fstack-clash-protection $(ADDFLAGS)
+OPTFLAGS			?= -O2 #-ftree-loop-vectorize -flto=$(THREADS)
+LDCXXFLAGS			:= -pthread -D_FORTIFY_SOURCE=2 -D_GLIBCXX_ASSERTIONS -fexceptions $(ADDFLAGS)
 override CXXFLAGS	+= $(REQFLAGS) $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)
 override LDFLAGS	+= $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)
 INC					:= -I$(INCDIR) -I$(SRCDIR)
--- a/Makefile
+++ b/Makefile
@@ -81,13 +81,17 @@ OBJEXT		:= o
 #? Flags, Libraries and Includes
 override REQFLAGS   := -std=c++20
 WARNFLAGS			:= -Wall -Wextra -pedantic
-OPTFLAGS			?= -O2 #-ftree-loop-vectorize -flto=$(THREADS)
+OPTFLAGS			?= -O2
 LDCXXFLAGS			:= -pthread -D_FORTIFY_SOURCE=2 -D_GLIBCXX_ASSERTIONS -fexceptions $(ADDFLAGS)
 override CXXFLAGS	+= $(REQFLAGS) $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)
 override LDFLAGS	+= $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)
 INC					:= -I$(INCDIR) -I$(SRCDIR)
 SU_USER				:= root
 SU_GROUP			:= root
+
+ifeq ($(ARCH),x86_64)
+	override OPTFLAGS += -ftree-loop-vectorize -flto=$(THREADS)
+endif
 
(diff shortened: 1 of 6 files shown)


aristocratos:
Nice, great initiative :)

Will look through it later this week.

Regarding the formatting, I'm also using vscode, but it doesn't auto format any placement of brackets and so on for me (it only autoadds the closing bracket after the cursor), I think this is likely an option somewhere. I don't have any style file or similar.

joske:
For formatting: put this line in your settings.json (or adjust to your liking and let me know ;-)):

    "C_Cpp.clang_format_style": "{BasedOnStyle: Google, IndentWidth: 4, IndentCaseLabels: true, TabWidth: 4, UseTab: ForIndentation, ColumnLimit: 0, NamespaceIndentation: All, FixNamespaceComments: true}"

Docs about this: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

I have a M1 macbook air, so arm64 like you, and a i5 11600 hackintosh (so not really real, but for most purposes fine). This is a desktop so can't test battery, but should be able to test temperature.

ShrirajHegde:
@joske disable `-fstack-protector -fstack-clash-protection` only if platform is detected as macos arm64 (setuid)

and also maybe disable suid option for the platform.

Makefile
```diff
@@ -81,7 +81,7 @@ OBJEXT		:= o
 #? Flags, Libraries and Includes
 override REQFLAGS   := -std=c++20
 WARNFLAGS			:= -Wall -Wextra -pedantic
-OPTFLAGS			?= -O2
+OPTFLAGS			?= -O0 -g
```
aristocratos, line 84: Should maybe setup a compile script to set OPTFLAGS="-O0 -g" instead of hardcoding it in the Makefile :)

joske: oops, and also the 2 ifneq is not right. But there's no way to do AND :-/

Makefile
```diff
@@ -93,6 +93,11 @@ SU_GROUP			:= root
 ifeq ($(ARCH),x86_64)
 	override OPTFLAGS += -ftree-loop-vectorize -flto=$(THREADS)
```
aristocratos, line 94: What compile errors do you get with `-flto`, it's usually caused by errors in the code that aren't caught unless all objects are optimized as one.

aristocratos:
Also regarding setuid, from https://superuser.com/questions/457868/making-a-program-always-run-as-root-in-os-x/1316546

> The setuid flag, which would allow a program to be run with its owners permissions. What seems to be a restriction in OS X, and apparently is not documented, is that the setuid bit on an executable has an effect only if the executable is in a directory that is owned by root (and not open for writing by others), etc, up to the root directory. Otherwise it is ignored for security reasons though.

Is this is saying that `setuid` works in OsX if under a superuser owned directory, don't have a mac at hand right now, but if someone want to test that...

joske:
setuid does not work indeed.

aristocratos:
@joske 
Can you try `make setuid SU_USER=root SU_GROUP=wheel` and see if that makes a difference?

src/osx/btop_collect.cpp
```diff
@@ -324,8 +324,29 @@ namespace Cpu {
...
 	auto get_battery() -> tuple<int, long, string> {
-		// if (not has_battery)
-		return {0, 0, ""};
+		if (not has_battery) return {0, 0, ""};
+
+		int percent = -1;
+		long seconds = -1;
+		string status = "discharging";
+
+		FILE *bat = popen("pmset -g batt", "r");
```
aristocratos, line 330: Yeah, this is gonna be a security risk, especially on OsX where admin privileges will be required to see stats about non user owned processes.

joske: Yes, it's a quick hack to get something working. I'll continue searching for an API. (I think this call can be done by all users on the system).

joske:
Great, fine for me. 

Regarding the battery stuff, I found the code for pmset, but the problem is that the APIs it uses are private :-/. The header files are missing on macos (available on opensource.apple.com though, but not sure if that's useful). Doesn't seem like anyone has code to programmatically get the battery stats...

aristocratos:
@joske 
Could be worth taking a look at how https://github.com/giampaolo/psutil handles it for OSX.

Added branch OSX and sent collaboration invite.

lbibass:
Unfortunately, this doesn't seem to want to compile on my system. (M1). 

Even though I'm compiling with GCC 11, I still get an 

```ld: symbol(s) not found for architecture arm64
collect2: error: ld returned 1 exit status
make: *** [btop] Error 1
``` 
error.  most of the issues appear to be in the btop.o file. Not sure what's happening here.

aristocratos:
@joske 
You also don't get cpu usage, threads and mem for non user owned processes.
Start with regular `btop` and then `sudo btop` and look at the difference in the processes.
So you are essentially limited to only seeing useful stats for your own processes without superuser rights for btop.

PR #80, opened 2021-10-09
Add github workflow for MacOS

ShrirajHegde:
Static compilation doesn't seem to work.


ShrirajHegde:
I am not sure about that since the workflow uses 10.x (macos-latest uses Catalina currently) and I tested it in a 10.x VM. Maybe someone with Big Sur can test it.

If it isn't compatible, workflow offers macos-11, so two binaries can be generated.

Issue #33, opened 2021-09-23
[REQUEST] homebrew formula


lucmichalski:
Hi @aristocratos ,

Hope you are all well !

I was wondering if it would be possible to create an homebrew formula for your great project :-)

Ps. apt package also ^^

Cheers,
Luc Michalski

aristocratos:
@lucmichalski 
All good and swell :)

I've said this before but I will not be doing any other packaging (than the static binaries + installer script), I'll leave that to those who are good at it.

Hopefully some homebrew enthusiast will notice this this request and create a PR with the needed homebrew scripts.

martin-braun:
Yes, I'd love to see this as well, for Intel and M1 chips.

XuehaiPan:
`btop` is now available in Homebrew (Homebrew/homebrew-core#90061).

martin-braun:
@XuehaiPan Thanks, I can confirm the installation on Intel devices works flawlessly.

brerx:
ARM on M1 works as well. like a charm. ❤️

aristocratos:
Will add install instructions to the readme.

nevack:
Happy to help, thanks for testing it! @aristocratos Can we close the issue now?

Issue #28, opened 2021-09-23
[REQUEST] Add Cmake support


WSLUser:
**Is your feature request related to a problem? Please describe.**
There are many benefits from using Cmake. Check out the documentation. Biggest one is portability. On platforms where `Make` isn't available, Cmake can still be made use of.
**Describe the solution you'd like**
I'd like to see this project built with llvm, which requires Cmake support. Compiler optimizations are much better using llvm.
**Describe alternatives you've considered**
Do nothing.
**Additional context**

aristocratos:
@WSLUser 
I really don't see a use for Cmake. This is a small project with few files. The only platform dependent compiler flag is `-fcf-protection` and is easily handled by logic in the Makefile.
`llvm-clang` is currently not supported becasue `clang` currently don't support C++20 Ranges. When `clang` has complete C++20 support (or as complete as `gcc`), there is no problem to again handle the correct flags for the compiler with logic in the Makefile.

> Biggest one is portability. On platforms where Make isn't available, Cmake can still be made use of.

First off, which platform are you referring to that doesn't support `Make` ?!?
Second, `CMake` can't replace `Make`. It's a build system generator, you still need a build system like `Make` to actually build the software.

WSLUser:
> First off, which platform are you referring to that doesn't support Make ?!?

Windows for one, but I think there are others based on arch or other reasons. I'm not sure how Visual Studio handles cmake, maybe they use the Msys2 or WSL make in the backend. Right now point is, I want to see the project available for as many platform and architectures as possible and that requires doing things like adding cmake support.

aristocratos:
Windows is very unlikely to be supported unless someone else takes on that pretty big challenge. Btop++ is written with very much POSIX in mind, things like threading, signal handling and input are very much POSIX dependent.

Changing the build system for something that isn't even looked at being implemented and if it was, likely would take a long time to finish would be extremely premature.

jan-guenter:
There was a lot of discussion in #82 but the TL:DR is that there will be no cmake in this repo unless it becomes necessary.
So I decided to merge my cmake branch into the main of my fork and maintain a cmake version there.
I've now setup a job scheduled to run every 5min merging the upstream master. If a new version tag was added upstream it will create a corresponding version tag with the suffix -cmake on my fork and run my build and release workflow.
This creates static musl builds for all available 51 targets and dynamically linked ubuntu packages.
https://github.com/jan-guenter/btop/releases

[...]

PR #82, opened 2021-10-14
cmake and automated release packaging with static musl builds

jan-guenter:
### Why cmake instead of a Makefile?
- it offers a more flexible build process which comes in handy when adapting to different systems and architectures
- cpack make packaging trivial and might speed up adoption by distro maintainers
- ctest will come in handy when test will be added
- the CMakeList.txt is more readable than a Makefile imho
- my cmake file is half the size and has all the features of the make file, minus the colorful output, plus packaging instructions and more compiler/linker checks

also see comments by @WSLUser in #28 

### Why static musl builds?
- in contrast to libc you can use `getpwuid` in static builds without hidden dynamic dependencies
- it has a lower minimum kernel requirement of 2.6.39 (earlier 2.6 versions might also work - needs to be tested)

### Automated release builds
I added a GitHub workflow that is triggered whenever a tag matching 'v*.*.*' is pushed.
This workflow will build static musl versions with the following toolchains:
- x86_64-linux-musl
- i686-linux-musl
- aarch64-linux-musl
- arm-linux-musleabi
- arm-linux-musleabihf
There are more pre-built toolchains available but I lack the hardware to test the produced binaries, so I excluded them for now.

It will also build dynamically linked amd64 DEB packages for Ubuntu 20.04 (focal) and 21.04 (hirsute). (requested by @lucmichalski  in #33 )
I'll add other distro packages and architectures shortly. My priorities are CentOS 8, Alpine Linux and Fedora.

After the packages have been built, the workflow will automatically create a GitHub release attaching all the packages.
It extracts the topmost part of the CHANGELOG.md and uses it as the release notes.

### Notes
This is still a work in progress and any input/feedback is welcome.

I also incorporated the Makefile changes from the OSX branch into the cmake file, but I don't have Apple hardware to test it with.

If the `STATIC` option is set in the cmake configuration step it checks if `getpwuid` can be used in static builds and does NOT set the preprocessor define `STATIC_BUILD` accordingly. Perhaps a renaming of this define or an additional define might be sensible.

Currently the workflow immediately publishes the release, but it could also be configured to only create a draft that can be reviewed and edited before final release. Let me what you prefer.

This is actually the first time I ever used GitHub workflows. I could draw a bit from my experience with gitlab CI, but I'm very unsure if I utilized all the workflow features exactly as indented. I'm certain it can be improved a lot and I would appreciate if someone would have a look and review my workflow.

I added a `*.yml` entry to the `.editorconfig` for easier workflow editing. The indentation style corresponds to the existing `continuous-build.yml`.

To demonstrate the workflow I added a version tag `v1.0.15-1` to my fork with runner and job debug logging enabled:
https://github.com/jan-guenter/btop/actions/runs/
https://github.com/jan-guenter/btop/releases/tag/v1.0.15-1

### Todo list
- [ ] add install script, readme etc. to archive packages like the static builds
- [ ] add more static builds for additional architectures (see the [list of available pre-built musl toochain images](https://hub.docker.com/r/muslcc/x86_64/tags))
- [ ] add more distro packages (especially CentOS 8, Alpine Linux and Fedora)
- [ ] add architecture variants to the Ubuntu builds
- [ ] update build and installation instructions in README.md
- [ ] update `snapcraft.yml` to use cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ target_link_options_if_supported(btop PRIVATE -fexceptions -fstack-protector -fs
 
 if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
 	target_compile_options_if_supported(btop PRIVATE -ftree-loop-vectorize -flto)
-	target_link_options_if_supported(btop PRIVATE -flto)
+	target_link_options_if_supported(btop PRIVATE -flto -s)
 endif()
 
 target_compile_definitions(btop PRIVATE _FORTIFY_SOURCE=2 _GLIBCXX_ASSERTIONS)
@@ -84,7 +84,7 @@ if(STATIC)
 	if(NOT STATIC_GETPWUID)
 		target_compile_definitions(btop PRIVATE STATIC_BUILD)
 	endif()
-	target_link_libraries(btop PRIVATE -static)
+	target_link_options_if_supported(btop PRIVATE -static)
 endif()
 
 set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Nobody <nobody@nowhere.no>")


WSLUser:
I think it's great to add musl support but most distros utilize libc. I'd like to see a libc++ option. It should of had a decent amount of improvements ever since Microsoft open sourced STL under same license so some code could be re-used wherever it made sense to do so.

joske:
Well I do not agree, it's his project, he gets to decide. This is how open source works. Anyone is free to fork if they disagree. But the people doing the work get to decide.

I see no reason to use CMake, or to change to any other build system when the current one works just fine. Just because it *may* be better for other projects doesn't mean everyone needs to switch. Make works fine for simple projects with only a handful of files.

WSLUser:
`Make works fine for simple projects with only a handful of files.`

Sure, until you run into a system that Make doesn't support. Make doesn't support every archecture or Operating System and I think from a user perspective, if you're going to make this available for people to use, then you should make it available to everyone everywhere. Make won't do that. Cmake will however.

Issue #87, opened 2021-10-16
when vim keys are active [k]ill should be [K]ill


jlopezcur:
**Describe the bug**

When we activate the vim_keys new option the label for kill a process should be capitalized as "Kill" for match the shortcut Shift+K

**To Reproduce**

- Set `vim_keys = True`

**Expected behavior**

See kill capitalized

**Screenshots**

With vim_keys active

![20211016_03h17m18s_grim](https://user-images.githubusercontent.com/4610342/137568032-319fce27-70bc-4f98-aa5e-8dff679b965d.png)

**Info (please complete the following information):**
 - btop++ version: 1.0.15
 - Platform: Linux

PR #88, opened 2021-10-16
fix: when vim keys [k]ill should be [K]ill

jlopezcur:
When we activate the vim_keys new option the label for kill a process
should be capitalized as "Kill" for match the shortcut Shift+K

fix: #87


aristocratos (approved):
Great, thanks for contributing :)

PR #92, opened 2021-10-16
static musl builds

jan-guenter:
### Description of changes

_for more details on the reasoning behind these changes see #84_

- added yaml definitions to `.editorconfig` for comfortable workflow editing
- changed static build behavior  to restrict `getpwuid` usage only in static builds with glibc so that static musl build have full functionality
- changed the continuous build workflow to create static musl binaries
- updated Makefile
  - added cf-protection for i*86 builds
  - added platform and arch detection from CXX machine tuple
  - add `PLATFORM_LC`, a lowercase version of `PLATFORM`, for case insensitive comparison
  - added bool flag `STRIP` to force `-s` linker flag
  - turned linker warnings into errors in case of `STATIC` to fail in case of glibc static linking issues
  - bumped version to 1.3
- updated README.md to reflect Makefile changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,6 @@ ifeq ($(STRIP),true)
 	override ADDFLAGS += -s
 endif
 
-#? Make sure PLATFORM Darwin is OSX and not Darwin
-ifeq ($(PLATFORM_LC),darwin)
-	ifeq ($(shell sw_vers >/dev/null 2>&1; echo $$?),0)
-		PLATFORM := OSX
-	endif
-endif
-
 #? Compiler and Linker
 CXX ?= g++
 override CXX_VERSION := $(shell $(CXX) -dumpfullversion -dumpversion || echo 0)
@@ -62,7 +55,7 @@ ifeq ($(PLATFORM_LC),linux)
 else ifeq ($(PLATFORM_LC),freebsd)
 	PLATFORM_DIR := freebsd
 	THREADS	:= $(shell getconf NPROCESSORS_ONLN 2>/dev/null || echo 1)
-else ifeq ($(PLATFORM_LC),osx)
+else ifeq ($(PLATFORM_LC),apple)
 	PLATFORM_DIR := osx
 	THREADS	:= $(shell sysctl -n hw.ncpu || echo 1)
 else
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ ARCH ?= $(shell $(CXX) -dumpmachine | cut -d "-" -f 1)
 override PLATFORM_LC := $(shell echo $(PLATFORM) | tr '[:upper:]' '[:lower:]')
 
 #? Only enable fcf-protection if on x86_64
-ifeq ($(ARCH),x86_64)
+ifneq ($(filter x86_64 i%86, $(ARCH)),)
 	override ADDFLAGS += -fcf-protection
 endif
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -14,21 +14,27 @@ endif
 
 PREFIX ?= /usr/local
 
-#? NOTICE! Manually set PLATFORM and ARCH if not compiling for host system
-PLATFORM ?= $(shell uname -s || echo unknown)
-ARCH ?= $(shell uname -m || echo unknown)
+#? Detect PLATFORM and ARCH from gcc if not set
+PLATFORM ?= $(shell $(CXX) -dumpmachine | awk -F"-" '{ print (NF==4) ? $$3 : $$2 }')
```
aristocratos, line 18: This will give you "apple" on OSX it seems like. So line 36 to 41 shouldn't be needed anymore.

Makefile
```diff
@@ -50,13 +56,13 @@ ifeq ($(CXX),g++)
...
-ifeq ($(PLATFORM),Linux)
+ifeq ($(PLATFORM_LC),linux)
 	PLATFORM_DIR := linux
 	THREADS	:= $(shell getconf _NPROCESSORS_ONLN 2>/dev/null || echo 1)
-else ifeq ($(PLATFORM),FreeBSD)
+else ifeq ($(PLATFORM_LC),freebsd)
 	PLATFORM_DIR := freebsd
 	THREADS	:= $(shell getconf NPROCESSORS_ONLN 2>/dev/null || echo 1)
-else ifeq ($(PLATFORM),OSX)
+else ifeq ($(PLATFORM_LC),osx)
```
aristocratos, line 65: This should then be "apple" instead of "osx"

README.md
```diff
@@ -259,7 +259,10 @@ Also needs a UTF8 locale and a font that covers:
 
    Append `QUIET=true` for less verbose output.
 
-   Notice! Manually set `$ARCH` variable if cross-compiling
+   Append `STRIP=true` to force stripping of debug symbols (adds `-s` linker flag).
+
+   Append `ARCH=<architecture>` to manually set the target architectur.
```
aristocratos, line 264: Just a missing "e" :)

src/linux/btop_collect.cpp
```diff
@@ -27,7 +27,7 @@ tab-size = 4
 #include <ifaddrs.h>
 #include <net/if.h>
 
-#ifndef STATIC_BUILD
+#if !defined(STATIC_BUILD) || !defined(__GLIBC__)
```
aristocratos, line 30: ~~Won't this also exclude normal builds (dynamic builds using glibc)?~~

Disregard that, the double negatives got me...

aristocratos (changes requested):

src/linux/btop_collect.cpp
jan-guenter: Should I perhaps make it read:
```cpp
#if !(defined(STATIC_BUILD) && defined(__GLIBC__)
```

Makefile
jan-guenter: changed to apple

README.md
jan-guenter: fixed spelling

src/linux/btop_collect.cpp
aristocratos: Sure, I would say that's easier to parse.

Issue #94, opened 2021-10-17
 [BUG] btop reports utter garbage in contrast to bpytop


bionade24:
Lots of metrics on my system are broken when using btop , while working fine with bpytop.

**To Reproduce**
I have a Dell XPS 9310 running Arch Linux. Haven't had time to check my other devices yet.

**Expected behavior**
btop looking super similar to bpytop.
**Screenshots**
Bpytop:  
![bpytop](https://user-images.githubusercontent.com/42534613/137622506-e652452f-c611-4f82-a57e-c13b2774407c.jpg)
Btop:  
![btop](https://user-images.githubusercontent.com/42534613/137623308-a5a77aed-03d6-4080-9351-13c6ae64b953.jpg)

**Info (please complete the following information):**
 - btop++ version: `btop version: 1.0.15`
 - (Linux) Kernel: `Linux Dell-Oskar 5.12.15-arch1-1 #1 SMP PREEMPT Wed, 07 Jul 2021 23:35:29 +0000 x86_64 GNU/Linux`
 - Terminal used: Tmux in Alacritty
 
**Additional context**
```

2021/10/17 (11:59:21) | ===> btop++ v.1.0.15
2021/10/17 (11:59:21) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS

2021/10/17 (11:59:31) | ===> btop++ v.1.0.15
2021/10/17 (11:59:31) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS
2021/10/17 (11:59:31) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS
2021/10/17 (11:59:33) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS
2021/10/17 (11:59:35) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS
2021/10/17 (11:59:37) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS
2021/10/17 (11:59:39) | WARNING: Failed to get disk/partition stats with statvfs() for: /run/media/NAS

2021/10/17 (12:02:55) | ===> btop++ v.1.0.15
2021/10/17 (12:02:55) | DEBUG: Starting in DEBUG mode!
2021/10/17 (12:02:55) | INFO: Logger set to DEBUG
2021/10/17 (12:02:55) | DEBUG: Using locale de_DE.UTF-8
[...]

PR #97, opened 2021-10-17
Improved battery detection

jan-guenter:
### Improvements to the battery detection:
 
- now considering all power supplies (instead of path name filtering)
- only consider power supplies that are currently present
- only consider power supplies of type Batter or UPS

_All changes made are in accordants to the definitions of `/sys/class/power_supply` from the [kernel docs](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power)_

---

This change was made to address the battery issue part of issue #94

This issue is still draft status, since I'm still awaiting feedback by the issue reporter @bionade24 on the test build I provided.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -480,15 +480,20 @@ namespace Cpu {
 					//? Only consider online power supplies of type Battery or UPS
 					//? see kernel docs for details on the file structure and contents
 					//? https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power
-					if (not d.is_directory()
-						or not fs::exists(d.path() / "type") 
-						or not fs::exists(d.path() / "present")
-						or stoi(readfile(d.path() / "present")) != 1)
+					try {
+						if (not d.is_directory()
+							or not fs::exists(d.path() / "type") 
+							or not fs::exists(d.path() / "present")
+							or stoi(readfile(d.path() / "present")) != 1)
+							continue;
+						string type = readfile(d.path() / "type");
+						if (type == "Battery" or type == "UPS") {
+							bat_dir = d.path();
+							break;
+						}
+					} catch (...) {
+						//? skip power supplies not conforming to the kernel standard
 						continue;
-					string type = readfile(d.path() / "type");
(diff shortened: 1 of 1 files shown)


src/linux/btop_collect.cpp
```diff
@@ -477,7 +477,16 @@ namespace Cpu {
 		if (bat_dir.empty() and has_battery) {
```
aristocratos, line 477: Looking at label of this section I believe my intention was to also check for validity (which I guess I never got around to doing).

src/linux/btop_collect.cpp
```diff
@@ -477,7 +477,16 @@ namespace Cpu {
...
 			if (fs::exists("/sys/class/power_supply")) {
 				for (const auto& d : fs::directory_iterator("/sys/class/power_supply")) {
-					if (const string dir_name = d.path().filename(); d.is_directory() and (dir_name.starts_with("BAT") or s_contains(str_to_lower(dir_name), "battery"))) {
+					//? Only consider online power supplies of type Battery or UPS
+					//? see kernel docs for details on the file structure and contents
+					//? https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power
+					if (not d.is_directory()
```
aristocratos, line 483: For safety I would like the the if statement to be wrapped in a try block to catch `std::invalid_argument` and `std::out_of_range` from the `stoi`.

aristocratos (approved):

aristocratos:
@jan-guenter 
Isn't the path `'src/**'` recursive for the workflow trigger? Noticed it didn't trigger when merging this.

jan-guenter:
> Isn't the path `'src/**'` recursive for the workflow trigger? Noticed it didn't trigger when merging this.

According to the docs it should https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet.
This actually is my first time working with GitHub Actions myself, so I'm not too familiar with them, but don't see why it didn't trigger. Perhaps a good time to use the [GitHub support](https://support.github.com/) you are paying for 😄
Please let me know if you find out why it didn't trigger.

PR #98, opened 2021-10-17
SVG logo

jan-guenter:
I created an SVG version of the logo.
It is an exact match to the existing logo.
The SVG XML was constructed by hand and is very optimized.
All shapes of each the letter, shadow and the shadows double border are solid shape layered on top of each other. The striped colors are created using a pattern fill of the shapes.

To demonstrate the pixel-perfect matching I also recreated the logo.png form the SVG using ImageMagick:
```bash
magick convert -background transparent logo.svg -extent 349x98 -roll +3+2 logo.png
```
The resulting pixels are identical but the size is 1,012b compared to the 11.5kb before.

_(the `-extent` and `-roll` parameters are only there to preserve the offset from the original logo.png)_

---

Why? you might ask... I'm preparing a presentation which contains a few slides showing btop++ and PowerPoints image scaling made the original pixel version look atrocious and I had some fun doing it 😄 

It can come in handy for various things, like printing, merch, websites, animations or when designing an app icon (e.g. KDE or Gnome icons use/accept SVG).


aristocratos (approved):
Cool :)

PR #127, opened 2021-10-29
Add support for building using meson

er888kh:
I thought it would be a good idea to have a build system that makes porting to other platforms easier. Another good point is that the build system is a lot more understandable to someone with no shell and Makefile experience, and also a lot more standardized. Meson specifically has testing and benchmarking features (which I haven't used here) but will be helpful in the future. In addition, dependency management is much easier in meson, which will be handy when many external dependencies are added (inevitable with GPU support).

This currently only has support for linux with a basic installation (with themes and readme), and notably setsid feature missing. I found out that a change I made in src/btop.cpp breaks reading version in the Makefile, though it builds successfully and the resulting binary works as intended.

To use meson run:
```bash
cd btop # Go to the root of the project
meson builddir # set builddir as the build directory
cd builddir
ninja # to build 
ninja install # to install or can be skipped if not needed
```
I didn't change default build flags, but it may be useful to do so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ stage/
 *.app
 
 build
+builddir
 bin
 btop
 .*/
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,40 @@
+project('btop', 'cpp',
+  version : '1.0.20',
+  default_options : ['cpp_std=c++20', 'warning_level=3'])
+
+src_files = files([
+  'src/btop_config.cpp',
+  'src/btop_draw.cpp',
+  'src/btop_input.cpp',
+  'src/btop_menu.cpp',
+  'src/btop_theme.cpp',
+  'src/btop_tools.cpp',
+  'src/linux/btop_collect.cpp',
+  'src/btop.cpp'
+])
+
+add_global_arguments('-DMESON_BUILD', language : ['c', 'cpp'])
+
+config_data = configuration_data()
+config_data.set_quoted('CONFIG_BTOP_VERSION', meson.project_version())
+
+configure_file(
+  output : 'config.h',
+  configuration: config_data,
+)
+
+src_inc = include_directories(['src'])
+thirdparty_inc = include_directories(['include'])
+
+threads_dep = dependency('threads')
+
+install_data('README.md', install_dir : get_option('datadir') / 'btop')
+install_subdir('themes', install_dir : get_option('datadir') / 'btop')
+
+exe = executable(
+  'btop',
+  sources: src_files,
+  dependencies: [threads_dep],
+  include_directories: [src_inc, thirdparty_inc],
+  install : true,
+)
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -37,6 +37,10 @@ tab-size = 4
 #include <btop_draw.hpp>
 #include <btop_menu.hpp>
 
+#ifdef MESON_BUILD
+	#include "config.h"
+#endif
+
 using std::string, std::string_view, std::vector, std::atomic, std::endl, std::cout, std::min, std::flush, std::endl;
 using std::string_literals::operator""s, std::to_string;
 namespace fs = std::filesystem;
@@ -53,7 +57,11 @@ namespace Global {
 		{"#801414", "██████╔╝   ██║   ╚██████╔╝██║        ╚═╝    ╚═╝"},
 		{"#000000", "╚═════╝    ╚═╝    ╚═════╝ ╚═╝"},
 	};
+#ifdef MESON_BUILD
+	const string Version = CONFIG_BTOP_VERSION;
+#else
 	const string Version = "1.0.20";
+#endif
 
 	int coreCount;
 	string overlay;


aristocratos:
@er888kh 
See #82 for previous discussion and my thoughts on the build system.

In short, the build system won't change unless actual issues with it arise. And if it does, @jan-guenter has a `cmake` based fork that would be the likely candidate to change to.

Thanks for the effort, but for now, the current Makefile based build system will stay.

jan-guenter:
@er888kh If you like feel free to play around with the cmake version. I've setup a notification in case the `Makefile` changes and keep the cmake file in sync. It also automatically merges the upstream changes when they occur (unless the Makefile changes).
https://github.com/jan-guenter/btop/tree/cmake

kloczek:
@er888kh could you polease keep updated that PR? 🤔 

Looks like maintainer completly do not understand that he is not only person building this project.
Despite that lack of understanding it would be good to have some civilised to building btop.

Current custom Makefile has several issue so even poorly maintained meson support is better than nothing.
BTW: it woild begood to move meson.build to root directory. In such simple cases like this project there is no need to overcomplicate meson support.

kloczek:
If you don't wnat do that I'll try to keep up-to-date PR with meson support.
Just please let me know.

aristocratos:
@kloczek 
> installs stripped binary

Sure, if you compile with `STRIP=true` which you manually would have to type...

> do not allow build in verbose mode which will print exact executed commands and all params

Below is a make dry-run with sed filtering out non essential information.
```bash
make QUIET=true -n |  sed '/^[printf|sleep|TSTAMP]/d'
```

> it does not alllow build btop using all CC, CXX, CFLAGS, LDFLAGS. LD env variables (without patching your Makefiule it is not possoble to use mold as linker)

Only CXX and LDFLAGS are ignored from environment, but they can still be declared as a make parameters.

> Inteasd it offers fancy colored outpout with hardcoded terminal sequences (try to have look on tput(1) command man page and think how to write such colouring query terminfo database for exact terminal sequences).

ANSI escape sequences aren't general enough for you? You'll be shocked if you look at the btop sources and notice that it's all hardcoded terminal sequences.

> I'm not compleining. Above are just raw fact.

No, I would say you are complaining. And most of it seem to be fuelled by your misconceptions of how `make` works. Since your "raw fact" seem to be mostly self-assured ignorance.

kloczek:
> @kloczek
> 
> > installs stripped binary
> 
> Sure, if you compile with `STRIP=true` which you manually would have to type...

And why did you decide to hardcode that as default?
 
> > do not allow build in verbose mode which will print exact executed commands and all params
> 
> Below is a make dry-run with sed filtering out non essential information.
> 
> ```shell
> make QUIET=true -n |  sed '/^[printf|sleep|TSTAMP]/d'
> ```

Yep ..
Easy to gueas and the same patter is used in how many other projects?

> > it does not alllow build btop using all CC, CXX, CFLAGS, LDFLAGS. LD env variables (without patching your Makefiule it is not possoble to use mold as linker)
> 
> Only CXX and LDFLAGS are ignored from environment, but they can still be declared as a make parameters.

Thank you for the confirmation that you are aware that you are blearing come standrds.

> > Inteasd it offers fancy colored outpout with hardcoded terminal sequences (try to have look on tput(1) command man page and think how to write such colouring query terminfo database for exact terminal sequences).
> 
> ANSI escape sequences aren't general enough for you? You'll be shocked if you look at the btop sources and notice that it's all hardcoded terminal sequences.

So you are not aware that "ansi" terminal handles only 16 colors and for example vt100 no at all?
```
[tkloczko@devel-g2v SPECS]$ TERM=ansi tput colors
8
[tkloczko@devel-g2v SPECS]$ TERM=vt100 tput colors
-1
```
Relaly at least please try to read what is termonfo/termcap database is ..

> > I'm not compleining. Above are just raw fact.
> 
> No, I would say you are complaining. And most of it seem to be fuelled by your misconceptions of how `make` works. Since your "raw fact" seem to be mostly self-assured ignorance.

Please enlight me ..

aristocratos:
@kloczek 
> And why did you decide to hardcode that as default?

It's not, look at the Makefile:
https://github.com/aristocratos/btop/blob/102b6dbc9d33906319e7c710cf58ba2893d6ec49/Makefile#L52-L54

> Yep ..
Easy to gueas and the same patter is used in how many other projects?


Huh? `make -n` is the operative term here, perfectly understable just reading `man make`:
```
-n, --just-print, --dry-run, --recon
            Print the commands that would be executed, but do not execute them (except in certain circumstances).
```

The sed filter is just basic shell knowledge, is still readable without it.

> Thank you for the confirmation that you are aware that you are blearing come standrds.

You are under the misconception that I'm following some standard. CXX from environment is ignored to make clear that `gcc` is the only compatible compiler for C++20 with ranges (right now). To override it you explicitly have to define `make CXX=<whatever-compiler>` so that for example macos don't try use `clang` which is aliased to `gcc` by default.

> So you are not aware that "ansi" terminal handles only 16 colors and for example vt100 no at all?

And why exactly should I care about that? If someone for some weird reason would be compiling btop on a only vt100 compatible terminal (lol...) it would still compile, just without... you know, colors...

> Please enlight me ..

The resources to enlighten yourself is endless, you can start with: https://www.gnu.org/software/make/manual/


And please, install a spellchecker in your browser, some of your sentences are seriously hard to parse.

kloczek:
> I just updated my main branch (not the one used in this PR)

Just checked https://github.com/er888kh/btop and I don't see v1.2.8 tag in in that repo.
Could you please rediff your repo?

PR #135, opened 2021-11-03
Update to Readme.md

lu4:
There seems to be a problem with `osx => OSX` branch casing, the command doesn't work by just copy and pasting on osx


aristocratos (approved):
Oops!

Thanks :)

Issue #153, opened 2021-11-15
[BUG] SISEGV on Mojave


mgradowski:
Running the binary results in `Segmentation fault: 11` both on the provided binary and the one I compiled myself.

**To Reproduce**
Run the binary.

**Expected behavior**
The binary runs.

**Screenshots**

![image](https://user-images.githubusercontent.com/24899416/141810184-c8bdaa04-262e-41d5-bf26-f45e49a75e89.png)

**Info (please complete the following information):**
 - btop++ version: 1.1.0
 - Binary: both release and self compiled
 - Compiler and version: g++-11 (Homebrew GCC 11.1.0_1) 11.1.0
 - Architecture: x86_64
 - Platform: OSX
 - (OSX/FreeBSD) Os release version: MacOS Mojave 10.14.6
 - Terminal used: Alacritty, ITerm2, MacOS Terminal
 - Font used: Default on all terminals

**Additional context**

contents of `~/.config/btop/btop.log`

2021/11/15 (16:37:47) | ===> btop++ v.1.1.0
2021/11/15 (16:37:47) | DEBUG: Starting in DEBUG mode!
2021/11/15 (16:37:47) | INFO: Logger set to DEBUG
2021/11/15 (16:37:47) | DEBUG: Using locale en_US.UTF-8
2021/11/15 (16:37:47) | INFO: Running on /dev/ttys003

**GDB Backtrace**

Can't get GDB to work due to system integrity protection.

aristocratos:
@mgradowski 
Hey, can you try the binary from the [Continuous Build MacOS](https://github.com/aristocratos/btop/actions/runs/1457692975)
It's built under Mojave so should be compatible with your system.

Also don't really have anything to go on right now, but you can try running with `lldb` instead of `gdb` and see if you can get a backtrace.

mgradowski:
Okay I'll try my best when I get home :)

mgradowski:
Only one thread is running when the program crashes:

```lldb
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x00007fff659c06a9 libobjc.A.dylib`objc_msgSend + 41
    frame #1: 0x00007fff659c80c8 libobjc.A.dylib`objc_object::sidetable_release(bool) + 268
    frame #2: 0x000000010010f9eb btop`::matching(page=65280, usage=5) at sensors.cpp:40:11
[...]

PR #154, opened 2021-11-16
fix SISEGV on Mojave

mgradowski:
My understanding of the issue:
 - CFStringRef is a pointer to __CFString.
 - CFRelease takes a pointer to a CF object.
 - We give it a pointer to a __CFString pointer.
 - This couldn't possibly go right!
 
So anyway, it's my first PR. I just changed the CFRelease call to free each CFStringRef separately. My understanding of CF is minimal so a review is welcome!


aristocratos:
@mgradowski 
Looks good, since I noticed the same in #153 :)

But I would ask @joske to review it first since he has the most experience with these weird macos system functions :P

joske (approved):
Makes sense to me, I didn't know *anything* about CF before I started the work on btop ;-).

Issue #151, opened 2021-11-14
[BUG] Core temperature not shown on Intel CPUs supported by coretemp before SandyBridge on Linux


ChUnterberger:
Btop does not show CPU core temperatures on Intel CPUs supported by coretemp before SandyBridge because these platforms do not support the package temperature sensor file **"temp1_input"** in the hwmon folder.

This is documented in the coretemp documentation found here.
[https://www.kernel.org/doc/html/v5.11/hwmon/coretemp.html](https://www.kernel.org/doc/html/v5.11/hwmon/coretemp.html)
> "The per-package sensor is new; as of now, it is present only in the SandyBridge platform."

The details are found in the source code.
[https://elixir.bootlin.com/linux/latest/source/drivers/hwmon/coretemp.c#L460](https://elixir.bootlin.com/linux/latest/source/drivers/hwmon/coretemp.c#L460)
```c
/*
 * Find attr number for sysfs:
 * We map the attr number to core id of the CPU
 * The attr number is always core id + 2
 * The Pkgtemp will always show up as temp1_*, if available
 */
```
This means that on platforms before SandyBridge the coretemp entries actually start with core 0 temperature at **"temp2_input"** and not with package temperature at **"temp1_input"**.

Here a sample of an Intel Core i7 860 (codename Lynnfield)
[https://ark.intel.com/content/www/us/en/ark/products/41316/intel-core-i7860-processor-8m-cache-2-80-ghz.html](https://ark.intel.com/content/www/us/en/ark/products/41316/intel-core-i7860-processor-8m-cache-2-80-ghz.html)
```shell
> sensors -u
coretemp-isa-0000
Adapter: ISA adapter
Core 0:
  temp2_input: 40.000
  temp2_max: 83.000
  temp2_crit: 99.000
  temp2_crit_alarm: 0.000
Core 1:
  temp3_input: 40.000
  temp3_max: 83.000
  temp3_crit: 99.000
  temp3_crit_alarm: 0.000
Core 2:
  temp4_input: 42.000
  temp4_max: 83.000
  temp4_crit: 99.000
  temp4_crit_alarm: 0.000
Core 3:
  temp5_input: 40.000
  temp5_max: 83.000
  temp5_crit: 99.000
  temp5_crit_alarm: 0.000
```
The btop code assumes that all temperature sensors have to start with **"temp1_input"** and will skip the folders where this is not the case.

[...]

PR #158, opened 2021-11-16
Fix coretemp iteration due to missing temp9_input (i9-9940X)

KFilipek:
Hi all,
This PR introduces small change I made accidentally - I've downloaded snap package with older version of btop and shows temperatures always 0'C. I didn't compile master branch before I've started fixing, don't ask.
#### Story behind changes
Anyway, after some reconnaissance I realized that I can only set temperature sensor for first few cores and package.
I have Intel i9-9940X (14 cores, 28 threads), so I should have `temp(1-15)_input`, but...
```sh
$ ls /sys/devices/platform/coretemp.0/hwmon/hwmon*/temp*_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp10_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp11_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp12_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp13_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp14_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp15_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp16_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp1_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp2_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp3_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp4_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp5_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp6_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp7_input
/sys/devices/platform/coretemp.0/hwmon/hwmon1/temp8_input
```
Surprise, CPU topology, I have `temp(1-8 and 10-16)_input`.
It's matches:
```sh
$ cat /sys/devices/system/cpu/cpu7/topology/core_id
8
$ cat /sys/devices/system/cpu/cpu6/topology/core_id
6
```

#### Proposed solution:
Iteration through files matching `tempX_input` pattern, it should cover my case and also bug: #151 

#### Commit description:
There are cases when some files are missing:
1. Due to specific CPU topology (e.g. {cpu6: core_id=6; but cpu7: core_id=8})
2. On architectures older than SandyBridge (missing per-package sensor)

P.S.
Great app, I love it.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -240,20 +240,36 @@ namespace Cpu {
 					if (s_contains(add_path, "coretemp"))
 						got_coretemp = true;
 
-					if (fs::exists(add_path / "temp1_input")) {
-						search_paths.push_back(add_path);
+					for (const auto & file : fs::directory_iterator(add_path)) {
+						if (string(file.path().filename()) == "device") {
+							for (const auto & dev_file : fs::directory_iterator(file.path())) {
+								string dev_filename = dev_file.path().filename();
+								if (dev_filename.starts_with("temp") and dev_filename.ends_with("_input")) {
+									search_paths.push_back(file.path());
+									break;
+								}
+							}
+						}
+
+						string filename = file.path().filename();
+						if (filename.starts_with("temp") and filename.ends_with("_input")) {
+							search_paths.push_back(add_path);
+							break;
+						}
 					}
-					else if (fs::exists(add_path / "device/temp1_input"))
-						search_paths.push_back(add_path / "device");
 				}
 			}
 			if (not got_coretemp and fs::exists(fs::path("/sys/devices/platform/coretemp.0/hwmon"))) {
 				for (auto& d : fs::directory_iterator(fs::path("/sys/devices/platform/coretemp.0/hwmon"))) {
 					fs::path add_path = fs::canonical(d.path());
 
-					if (fs::exists(d.path() / "temp1_input") and not v_contains(search_paths, add_path)) {
-						search_paths.push_back(add_path);
-						got_coretemp = true;
+					for (const auto & file : fs::directory_iterator(add_path)) {
+						string filename = file.path().filename();
+						if (filename.starts_with("temp") and filename.ends_with("_input") and not v_contains(search_paths, add_path)) {
+								search_paths.push_back(add_path);
+								got_coretemp = true;
+								break;
+						}
 					}
 				}
 			}
@@ -261,9 +277,17 @@ namespace Cpu {
 			if (not search_paths.empty()) {
 				for (const auto& path : search_paths) {
 					const string pname = readfile(path / "name", path.filename());
-					for (int i = 1; fs::exists(path / string("temp" + to_string(i) + "_input")); i++) {
-						const string basepath = path / string("temp" + to_string(i) + "_");
-						const string label = readfile(fs::path(basepath + "label"), "temp" + to_string(i));
+					for (const auto & file : fs::directory_iterator(path)) {
+						const string file_suffix = "input";
+						const int file_id = atoi(file.path().filename().c_str() + 4); // skip "temp" prefix
+						string file_path = file.path();
+
+						if (!s_contains(file_path, file_suffix)) {
+							continue;
+						}
+
+						const string basepath = file_path.erase(file_path.find(file_suffix), file_suffix.length());
+						const string label = readfile(fs::path(basepath + "label"), "temp" + to_string(file_id));
 						const string sensor_name = pname + "/" + label;
 						const int64_t temp = stol(readfile(fs::path(basepath + "input"), "0")) / 1000;
 						const int64_t high = stol(readfile(fs::path(basepath + "max"), "80000")) / 1000;


KFilipek:
According to structure I got on Intel(R) Core(TM) i5-6600 CPU @ 3.30GHz:
```
/sys/class/hwmon/hwmon0/temp1_input
/sys/class/hwmon/hwmon1/temp1_crit
/sys/class/hwmon/hwmon1/temp1_crit_alarm
/sys/class/hwmon/hwmon1/temp1_input
/sys/class/hwmon/hwmon1/temp1_label
/sys/class/hwmon/hwmon1/temp1_max
/sys/class/hwmon/hwmon1/temp2_crit
/sys/class/hwmon/hwmon1/temp2_crit_alarm
/sys/class/hwmon/hwmon1/temp2_input
/sys/class/hwmon/hwmon1/temp2_label
/sys/class/hwmon/hwmon1/temp2_max
/sys/class/hwmon/hwmon1/temp3_crit
/sys/class/hwmon/hwmon1/temp3_crit_alarm
/sys/class/hwmon/hwmon1/temp3_input
/sys/class/hwmon/hwmon1/temp3_label
/sys/class/hwmon/hwmon1/temp3_max
/sys/class/hwmon/hwmon1/temp4_crit
/sys/class/hwmon/hwmon1/temp4_crit_alarm
/sys/class/hwmon/hwmon1/temp4_input
/sys/class/hwmon/hwmon1/temp4_label
/sys/class/hwmon/hwmon1/temp4_max
/sys/class/hwmon/hwmon1/temp5_crit
/sys/class/hwmon/hwmon1/temp5_crit_alarm
/sys/class/hwmon/hwmon1/temp5_input
/sys/class/hwmon/hwmon1/temp5_label
/sys/class/hwmon/hwmon1/temp5_max
```
I changed the following:
1. Find a file that starts with "temp" and ends with "_label" to mitigate case like above (only `/sys/class/hwmon/hwmon0/temp1_input` without any other files like `*_crit`, `*_label` etc).
2. Missing files like `temp1_label` (PR #151) or `temp9_label` on i9-9940X are not problem.

I can test on 2 socket machine also to see CPU topology.

Best regards,
Krzysztof Filipek

aristocratos (commented):
Looks good, great work!
If you're done with it I will merge in next release.

aristocratos (approved):

ChUnterberger:
I think this PR needs another round. While it finds all the sensors on all the machines i tested, i noticed a new bug inroduced with this PR which messes up the core mapping.

You can reproduce the issue by stressing a single core and checking the core temperature.
You can see that on some machines this will not match with the core which has the 100% load.
I tested this with the stress tool.

`stress -c 1` for single core load.

The reason for this is that the `fs::directory_iterator` at [https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L280](https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L280) does not return the results sorted and `core_sensors` is expected sorted because at [https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L372](https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L372) the index of the core is used as a index to the vector position.
```
core_sensors
[0] "coretemp/Core 1"
[1] "coretemp/Core 2"
[2] "coretemp/Core 3"
[3] "coretemp/Core 0"
```
For this reason the code will show all core temperatures offset by 1.

I think we either need to sort before we insert or we need to replace the vector with a map and use the core id as index.

BTW:  [https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L312](https://github.com/KFilipek/btop/blob/fix-iterate_through_coretemp/src/linux/btop_collect.cpp#L312) still uses the old style for loop. I think for consistency we would want the same style.

aristocratos:
@ChUnterberger @KFilipek 
> I think we either need to sort before we insert or we need to replace the vector with a map and use the core id as index.

Since the function is only run once the easiest way to fix this would probably be to just add:
```c++
else rng::sort(core_sensors, rng::less{});
```
after line 336.

ChUnterberger:
@aristocratos This fixes the issue for me but would not work on machines with more than 10 cores.
This is because the sort would order the cores

core_sensors
[0] "coretemp/Core 0"
[1] "coretemp/Core 1"
[2] "coretemp/Core 10"
[3] "coretemp/Core 2"
```

**Comment by aristocratos** [comment]
@ChUnterberger @KFilipek 
Oops :)

Should be fixed by:
```c++
else {
	rng::sort(core_sensors, rng::less{});
	rng::stable_sort(core_sensors, [](const auto& a, const auto& b){
		return a.size() < b.size();
	});
}
```

aristocratos:
Yeah, should work:
https://godbolt.org/z/necEnv945

As long as the assumption that the prefix before the number is the same length holds true, this will work.

ChUnterberger:
Yes either this or this should do the trick.

```c
else rng::sort(core_sensors, [](const auto a, const auto b) { 
   return std::stoi(a.substr(a.find_first_of("0123456789"))) < std::stoi(b.substr(b.find_first_of("0123456789")));
});
```

aristocratos:
@ChUnterberger 
Yeah, that's neat, didn't realize `string::find_first_of()` would match _any_ of the characters specified.

aristocratos:
@ChUnterberger
However the `stoi` opens up for a `std::out_of_range` exception if for some reason the name doesn't include a number, so would be slightly less safe.

ChUnterberger:
@aristocratos 
True, the code makes some assumptions about the format of the strings which if violated causes a runtime exception. 

The first code would fail to sort correctly if numbers are padded with 0. E.g. `coretemp/Core 009` while the second code could cause a runtime exception if the passed string does not contain a number, is in hex or oct or if the value is out of range.

Both are not likely to occur but to be on the safe side a implementation which cannot throw unhandled runtime errors is preferred even if it silently hides a potential core mapping issue.

aristocratos:
Merged in v1.1.2

@KFilipek thanks for contributing!
@ChUnterberger thanks for the input!

PR #180, opened 2021-12-02
Create gruvbox_dark_v2.theme

pietryszak:
I create a theme with colors used only from original scheme https://github.com/morhetz/gruvbox and with 3 colors gradient.

image
image


aristocratos (approved):
Looks great, thanks for contributing!

pietryszak:
> Looks great, thanks for contributing!

I think this theme can be added to bashtop and bpytop too.

PR #186, opened 2021-12-05
Remove duplicated fstype

GuillaumeGomez:
Very simple nit where a fstype check is duplicated.


aristocratos:
Good catch and thanks for contributing!

PR #218, opened 2021-12-19
Create dependency files in build directory when compiling

stwnt:
The ```-MMD``` flag tells ```gcc``` to create prerequisites next to the object files and it is aware of the output directory.


aristocratos (approved):
Nice, thanks for contributing!

PR #222, opened 2021-12-21
fix CPU temp fallback on macOS

joske:
@aristocratos You can remove the debug lines if you want. I've left the ones in get_sensors() as this is only ran once.


aristocratos (approved):

aristocratos:
@joske 
The debug logging should be fine, could be useful for debugging :)
Awesome work as always!

joske:
@aristocratos will the original code wasn't so awesome apparently 😬

aristocratos:
@joske Considering you were writing it blind without being able to test it and the poster of the original issue never complained and that it didn't break anything else, I would say it was a partial success ;)

PR #227, opened 2021-12-29
in_avail() can always be zero, it is an optimization opportunity only

pg83:
in_avail() can always return zero, so libc++'s streams does. 

So, one can not use such a technique to async cin poll.

One can use poll/select for zero fd, but it will not work for file inputs, like in 'btop < some_file'.

Best available option - just read input in separate thread. Also, after this commit, we will have an opportunity to remove sleep's from code, and use condition_variable timed_wait.

Also we should always try quick_exit, instead of plain exit() call, cause:

1) why bother with threads? let them die fast.
2) should not overcomplicate shutdown process

And I do not personally understand, why quick_exit disabled on Darwin... One can always use _Exit()/_exit() instead.

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -237,15 +237,15 @@ void clean_quit(int sig) {
 		std::cerr << Global::fg_red << "ERROR: " << Global::fg_white << Global::exit_error_msg << Fx::reset << endl;
 	}
 	Logger::info("Quitting! Runtime: " + sec_to_dhms(time_s() - Global::start_time));
-
-	const auto excode = (sig != -1 ? sig : 0);
-
+	close(0);
 	//? Assume error if still not cleaned up and call quick_exit to avoid a segfault from Tools::atomic_lock destructor
 #ifndef __APPLE__
-	quick_exit(excode);
+	if (Tools::active_locks > 0) {
+		quick_exit((sig != -1 ? sig : 0));
+	}
 #endif
 
-	exit(excode);
+	if (sig != -1) exit(sig);
 }
 
 //* Handler for SIGTSTP; stops threads, restores terminal and sends SIGSTOP
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -17,6 +17,7 @@ tab-size = 4
 */
 
 #include <csignal>
+#include <clocale>
 #include <pthread.h>
 #include <thread>
 #include <numeric>
@@ -237,14 +238,14 @@ void clean_quit(int sig) {
 	}
 	Logger::info("Quitting! Runtime: " + sec_to_dhms(time_s() - Global::start_time));
 
+	const auto excode = (sig != -1 ? sig : 0);
+
 	//? Assume error if still not cleaned up and call quick_exit to avoid a segfault from Tools::atomic_lock destructor
 #ifndef __APPLE__
-	if (Tools::active_locks > 0) {
-		quick_exit((sig != -1 ? sig : 0));
-	}
+	quick_exit(excode);
 #endif
 
-	if (sig != -1) exit(sig);
+	exit(excode);
 }
 
 //* Handler for SIGTSTP; stops threads, restores terminal and sends SIGSTOP
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -241,11 +241,11 @@ void clean_quit(int sig) {
 	const auto excode = (sig != -1 ? sig : 0);
 
 	//? Assume error if still not cleaned up and call quick_exit to avoid a segfault from Tools::atomic_lock destructor
-#ifndef __APPLE__
+#ifdef __APPLE__
+	_Exit(excode);
+#else
 	quick_exit(excode);
 #endif
-
-	exit(excode);
 }
 
 //* Handler for SIGTSTP; stops threads, restores terminal and sends SIGSTOP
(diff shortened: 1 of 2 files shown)


aristocratos:
@pg83 
> in_avail() can always be zero [...]

Do you have any examples or sources for this? Have not gotten any bug reports that input doesn't work, which would be the case if this happened.

> Also we should always try quick_exit, instead of plain exit() call, cause:
>
>   1. why bother with threads? let them die fast.
>   2. should not overcomplicate shutdown process

Not sure I agree that cleaning up would be to overcomplicate things? The "runner" thread can be writing to the logfile so generally a good idea to let it finish and exit cleanly. 
And also in also in answer to your last question, `quick_exit()` isn't available on all platforms, like on macos.

Also from the PR:
`// intentional memory leak, to simplify shutdown process`
Can you expand on this?

aristocratos:
Honestly, I can't say I see what the "optimization" part of this is?
It's creating more threads and intentionally leaking, that sounds a lot like the opposite of optimization.

pg83:
>Do you have any examples or sources for this

https://github.com/llvm/llvm-project/blob/main/libcxx/include/streambuf#L172
https://github.com/llvm/llvm-project/blob/main/libcxx/include/streambuf#L396

>And also in answer to your last question, quick_exit() isn't available on all platforms, like on macos.

One can always use _exit()/_Exit(), with same semantics

>Can you expand on this?

If we do not leak memory here, and call destructor of InputThr, running thread can damage already destroyed instance of InputThr. To implement graceful shutdown, one must stop running thread, join it, and destroy InputThr instance. It is too complex(and require more extra housekeeping, like stop condition variable, and so on) for a such simple task.

pg83:
> Honestly, I can't say I see what the "optimization" part of this is?
> It's creating more threads and intentionally leaking, that sounds a lot like the opposite of optimization.

It is not an optimization, it is fix of incorrect usage of standart c++ classes. in_avail() for cin can be always zero for a conforming implementation, like in libc++, and current btop code never try to actually read from stream in this case.

pg83:
https://www.cplusplus.com/reference/streambuf/streambuf/in_avail/
https://www.cplusplus.com/reference/streambuf/streambuf/showmanyc/

0 | no information | Further calls may either retrieve more characters or return traits_type::eof().

aristocratos:
But why would this always return zero?

`showmanyc()` always returns zero for the currently used gcc libs also:
https://github.com/gcc-mirror/gcc/blob/52a6b06c26346cfb6972252ef547720b431844e6/libstdc%2B%2B-v3/include/std/streambuf#L654

Doesn't the `if` clause return if there are characters there?
```c++
if (__ninp_ < __einp_)
    return static_cast<streamsize>(__einp_ - __ninp_);
```

> It is not an optimization, it is fix of incorrect usage of standart c++ classes.

Well you titled it "it is an optimization opportunity only" :)

> One can always use _exit()/_Exit(), with same semantics

Sure, but would be even less "clean".

aristocratos:
> 0 | no information | Further calls may either retrieve more characters or return traits_type::eof().

Wouldn't it be possible to `peek()` and check for `EOF` instead? Or would that also block?

pg83:
> > 0 | no information | Further calls may either retrieve more characters or return traits_type::eof().
> 
> Wouldn't it be possible to `peek()` and check for `EOF` instead? Or would that also block?

Actually, I answer this in my first message :)

It will work for sockets/pipes. but not for files. Only one safe way to achive this - by using threads.

>Well you titled it "it is an optimization opportunity only" :)
My english is not very good. What I mean - "in_avail() can be used only as hint/optimization opportunity, but not as signal about available data"

pg83:
>Sure, but would be even less "clean".

How I convince you, that _Exit()/_exit()/quick_exit() is safest possible way to do shutdown in case of threads?

pg83:
>Doesn't the if clause return if there are characters there?

Yes, but only in case if you already call (possibly blocking) read() before, and current code never call read()/get()/whatever iff in_avail() == 0.

aristocratos:
> How I convince you, that _Exit()/_exit()/quick_exit() is safest possible way to do shutdown in case of threads?

I'm convinced, macos will stall waiting for input without it with the proposed changes.
Add them and I'll merge it :)

> Yes, but only in case if you already call (possibly blocking) read() before, and current code never call read()/get()/whatever iff in_avail() == 0.

Ah, got it.

pg83:
>Add them and I'll merge it :)

Like this?

aristocratos:
But why did you revert the "no memory leak" commit?

pg83:
> But why did you revert the "no memory leak" commit?

I can return it, but:

1) It is more(and complex) code to execute on shutdown path, and bigger probability of error in code. My expirience says that shutdown must be as simple as possible

2) it is rather common to not free resources on shutdown, to achive faster shutdown.

Anyway, I will do as you say, I have no some kind of personal preference here :)

aristocratos:
If the threads don't leave any unreleased resources it doesn't really matter.

pg83:
> If the threads don't leave any unreleased resources it doesn't really matter.

So I should return cleanup phase?

aristocratos:
> So I should return cleanup phase?

I'm no way any expert on threading, so if the cleanup isn't needed and there is no "persistent" leakage of memory, then I don't see any reason for it.

So I'll leave it up to you.

pg83:
>So I'll leave it up to you.

I prefer not to clean up, cause it is much safer(debug threading deadlock in cleanup phase is hard, belive me)

>I'm in no way any expert on threading
Well, I am, no doubt, cause my day work is developing some hi-load multi-threading daemons :)

aristocratos (approved):
> my day work is developing some hi-load multi-threading daemons :)

I'll defer to your knowledge then :)

Great fix and nice work, thanks for contributing!

PR #253, opened 2022-01-25
Just keeping our repos synced.

kz6fittycent:


snap/snapcraft.yaml
```diff
@@ -36,8 +36,9 @@ apps:
       - hardware-observe
       - network
       - network-observe
-      - physical-memory-observe
+    #  - physical-memory-observe
```
T3sT3ro, line 39: this also needs an update in README snap section

kz6fittycent: Thanks for the reminder, the README is updated. The snap now has autoconnect setup for all but `removable-media`.

aristocratos: Oh, great!

aristocratos (approved):

PR #254, opened 2022-01-31
README: Update openSUSE install section

sebix:
The package is available in the official default repos for Tumbleweed. For Leap, the package is available in the development repository. Linking to the generic page, which has the current information - independent of distro releases and package submissions. The user's repository is obsolete/not necessary any more.


aristocratos (approved):
Thanks!

PR #266, opened 2022-02-13
Sort input file list

bmwiedemann:
Sort input file list
so that `btop` builds in a reproducible way
in spite of indeterministic filesystem readdir order. (linking order matters)

See https://reproducible-builds.org/ for why this is good.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,9 +129,9 @@ ifdef DEBUG
 	override OPTFLAGS := -O0 -g
 endif
 
-SOURCES	:= $(shell find $(SRCDIR) -maxdepth 1 -type f -name *.$(SRCEXT))
+SOURCES	:= $(sort $(shell find $(SRCDIR) -maxdepth 1 -type f -name *.$(SRCEXT)))
 
-SOURCES += $(shell find $(SRCDIR)/$(PLATFORM_DIR) -type f -name *.$(SRCEXT))
+SOURCES += $(sort $(shell find $(SRCDIR)/$(PLATFORM_DIR) -type f -name *.$(SRCEXT)))
 
 SOURCE_COUNT := $(words $(SOURCES))
 


aristocratos:
@bmwiedemann 
Does this matter when compiling with `-flto` (used by default for btop) since no linking is happening until all object files are already compiled?

bmwiedemann:
It matters both with and without `-flto`

aristocratos:
@bmwiedemann 
Looks like there is more changes needed to make it deterministic.
From: https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html

> Randomness created by the compiler
> 
> This problem arises for example in gcc when Link-Time Optimizations are activated (with the -flto flag). This option introduces randomly generated names in the binary files. The only way to avoid this problem is to use -frandom-seed flag. This option provides a seed that gcc uses when it would otherwise use random numbers. It is used to generate certain symbol names that have to be different in every compiled file. It is also used to place unique stamps in coverage data files and the object files that produce them. This setting has to be different for each source file. One option would be to set it to the checksum of the file so the probability of collision is very low. For example in CMake it could be made with a function like this [...]

bmwiedemann:
This PR was the only patch needed to make openSUSE `btop` package builds reproducible (even with LTO).
Probably gcc got some toolchain fix in the meantime.

aristocratos:
@bmwiedemann 
I don't doubt your claim, but I'm curious, what was the setup for replicating the result?

aristocratos (approved):

bmwiedemann:
> @bmwiedemann I don't doubt your claim, but I'm curious, what was the setup for replicating the result?

on openSUSE, I use 
```bash
osc checkout openSUSE:Factory/btop && cd $_
rbk
```
where the `rbk` script comes from
https://github.com/bmwiedemann/reproducibleopensuse/

aristocratos:
Alright, thanks for contributing!

PR #273, opened 2022-02-21
add "g" and "G" to vim keys

mohi001:
add "g" (equivalent to "home") and "G" (equivalent to "end") to vim keys


aristocratos (approved):

PR #274, opened 2022-02-23
Fixed a string in the options menu

NMrocks:
(Copy-pasted from commit description)
The "Proc Sorting" option in the proc section of the options menu has "cpu direct" as the 7th option but the description of the option mentioned it using its older name, "cpu responsive". Changed this to "cpu direct" in the description.

Sorry for making such a minor PR but this bugged me so I fixed it :)


aristocratos (approved):
Oh right, missed that :)

While you're at it do you wanna fix the same description in btop_config.cpp?

NMrocks:
Sure lol

aristocratos:
Great :)
Thanks for contributing!

PR #284, opened 2022-03-07
Fixed: Wrong memory unit when shorten and size is less than 10

### LINKED ISSUE #283: [BUG] Wrong memory unit
Unit of memory in proc section is wrong

**To Reproduce**

Call floating_humanizer() with a size less than 10 when shortened and slightly less than a integer when converted to MB
 e.g(8388608 - 10)
this will result 8G
expected
8.0M

Sc_14001216_20220307_15:51:40

 btop++ version:1.2.5
 Commit `2fa800e21e2a74f3b12f52178a75e067ce349648` caused this

mohi001:
Fix #283

### DIFF
--- a/src/btop_tools.cpp
+++ b/src/btop_tools.cpp
@@ -372,7 +372,7 @@ namespace Tools {
 		}
 		if (shorten) {
 			auto f_pos = out.find('.');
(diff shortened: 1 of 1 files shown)


aristocratos (approved):
Great catch!

Thanks for contributing!

Issue #159, opened 2021-11-17
[REQUEST] Non-case-sensitive Process Filtering


RVRX:
**Is your feature request related to a problem? Please describe.**
When I want to filter/search for a process I have to get the capitalization correct to find it.

**Describe the solution you'd like**
The ability to toggle off case-sensitive searches

**Additional context**
All lower-case:
<img width="698" alt="image" src="https://user-images.githubusercontent.com/45409688/142261992-ba6c05b8-f971-4683-ae37-274c9e76cccb.png">
Matching-case:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/45409688/142262022-8eaa8150-5686-4d4e-8435-f18ce36f78ae.png">
_I would like the processes that show up when I type `iTerm` to all show up when I type `iterm`._

lucatrv:
I support this request, it is very annoying to filter as `xorg` and not find `Xorg`...

c02y:
+1

PR #326, opened 2022-04-30
Added: Case insensitive process filtering

abrasumente233:
As pointed out in #159, when filtering for a process, we have to get the capitalization correct to find it.

This patch solves it.

### DIFF
--- a/src/btop_tools.hpp
+++ b/src/btop_tools.hpp
@@ -187,6 +187,16 @@ namespace Tools {
 		return str.find(find_val) != string::npos;
 	}
 
+	//* Check if string <str> contains string <find_val>, while ignoring case
+	inline bool s_contains_ic(const string& str, const string& find_val) {
+        auto it = std::search(
+            str.begin(), str.end(),
+            find_val.begin(), find_val.end(),
+            [](char ch1, char ch2) { return std::toupper(ch1) == std::toupper(ch2); }
+        );
+		return it != str.end();
+	}
+
 	//* Return index of <find_val> from vector <vec>, returns size of <vec> if <find_val> is not present
 	template <typename T>
 	inline size_t v_index(const vector<T>& vec, const T& find_val) {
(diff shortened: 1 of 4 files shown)


aristocratos:
@abrasumente233 
Great initiative!
However, I don't have access to my development machine for a couple of days, would you mind doing some benchmarking since this code will be processing alot of strings at every update when a filter is active.
Just to make sure it doesn't have too much of an performance impact.

abrasumente233:
> @abrasumente233 Great initiative! However, I don't have access to my development machine for a couple of days, would you mind doing some benchmarking since this code will be processing alot of strings at every update when a filter is active. Just to make sure it doesn't have too much of an performance impact.

Sorry for the late reply, as I was busy these days lol.

I put together a small benchmark comparing the original case-sensitive and the new case-sensitive findstr functions. Here are some numbers on Apple M1:

```
2022-05-12T22:39:33+08:00
Running ./findstr
Run on (8 X 24.1216 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 2.20, 2.70, 2.81
-------------------------------------------------------------------------------
Benchmark                                     Time             CPU   Iterations
-------------------------------------------------------------------------------
bench_findstr/sensitive/cmd_filter         26.9 ns         26.9 ns     25513111
bench_findstr/insensitive/cmd_filter       66.3 ns         66.3 ns     10522676
```

The new findstr function is 2.46x slower than the original one. I don't know if that's an acceptable slowdown, but let's assume there are 10000 processes and we need to filter 5 fields on each frame. And if we have 30 fps, we must compute a frame within 33.33 ms, and the findstr takes 10000 processes * 5 fields * 66.3 ns = 3.315 ms, which is ~10% of the time budget.

Does it look good to you or we have to further imporve the findstr algorithm?

aristocratos (approved):
@abrasumente233 
> The new findstr function is 2.46x slower than the original one.
> Does it look good to you or we have to further imporve the findstr algorithm?

No that looks fine!
Thanks for contributing :)

PR #327, opened 2022-04-30
Include ZFS ARC in cached/available memory on Linux

### LINKED ISSUE #150: [REQUEST] Option to count ZFS ARC as cached memory instead of used?

When using ZFS, the Adaptive Replacement Cache (ARC) is counted by Linux as used memory, because it is separate from the normal disk cache. However, my understanding is that it functions more like cached memory, in that it will be freed if the system is under memory pressure.

It would be nice if btop had an option to count the ZFS ARC as cached memory instead of used, by getting the size of the ARC and subtracting it from used memory and adding it to cached.

The only alternative I see would be to not implement this.

**Additional context**

If implemented, this probably needs to be a user option, since some people will definitely want the memory stats exactly as reported by the kernel.

mattico:
Fixes #150


aristocratos (approved):
Awesome, thanks for contributing!

intrigus:
@mattico isn't https://github.com/aristocratos/btop/pull/327/files#diff-ee570ba04f2b49d02794ec9f8001902d40c7cabf47397dcc87e8f5f37254cb50R813 incorrect?
Shouldn't this line be:
```c
mem.stats.at("available") += arc_size - arc_min_size;
```
?
Have a look at this similar issue for htop:
https://github.com/htop-dev/htop/pull/1003/files#diff-0fcd76c22533048f3cae2ee4216d4f5e97f84c63faa06e0713ee4f2091d0ca47

mattico:
I don't have time to look at this right now, but from what I remember the first version of the code was more similar to the htop version but I changed it because there was some problem related to the surrounding code that made it not work in certain situations.

Have you noticed any issues? I've been using this version regularly and haven't noticed anything that looked wrong.

intrigus:
I didn't notice any issues per se, but I think `btop` will show more available memory than is available, because ZFS will not release `arc_min_size` worth of memory.
(I looked for ZFS support in bpytop and saw that it wasn't supported, so I searched how htop does it.
So I became aware of `arc_min_size` and after that found `btop` and saw that it probably does not account for `arc_min_size`)

PR #336, opened 2022-05-20
Add option for filtering kernel processes.

0xJoeMama:
It's not ready, however I think it's worth creating a PR for.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,4 @@
 BasedOnStyle: Google
-IndentWidth: 4
+UseTab: ForIndentation
+TabWidth: 4
 PointerAlignment: Left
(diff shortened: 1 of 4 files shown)


0xJoeMama:
Also I still need to fix the formatting changes.

aristocratos:
@0xJoeMama 
You should remove the `.clang-format` file and submit as a separate PR with a good explanation why it should be considered, as per https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md

`.gitignore` should also be reverted, if you have files you want to ignore in your local copy use `.git/info/exclude` file which is private.

As for the other files, all formatting need to be reset. Easiest might be to start clean from `main` and copy paste your changes.

aristocratos (changes requested):
Needed changes in comment above.

0xJoeMama:
I will close this and promptly open one with the branch with the correct fixes.

Issue #332, opened 2022-05-15
[REQUEST] Add kernel process filtering.


0xJoeMama:
**Is your feature request related to a problem? Please describe.**
Personally I think that kernel processes aren't needed to be rendered, with some rare exceptions. It is not a problem however they really pollute the process list with a whole bunch of junk.

**Describe the solution you'd like**
I have added a config option to disable kernel processes from being rendered on my fork. 

**Additional context**
I have made my fork of the repo just to hold my solution so that I can use it in the future. However I think it is worth pushing to the public release because I can see a lot of people missing this feature after switching from htop. 

Thank you for your time!

0xJoeMama:
> since you changed the indentation in the files you edited.
Yes I am aware that is an issue and I am working on a solution with a clang-format configuration file that I would like to merge with upstream if possible.

0xJoeMama:
Also regarding osx and bsd. I am not sure how processes are structured on those systems. I need to research that before I create a PR, at least that would make me feel better than just creating a platform specific feature.

If you have any useful *pointers* I'd love to hear them out! Thank you for your time!

PR #337, opened 2022-05-20
Proper kernel process filtering.

0xJoeMama:
Properly add the functionality mentioned in #332.

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -1,3 +1,4 @@
+// clang-format off
 /* Copyright 2021 Aristocratos (jakob@qvantnet.com)
 
    Licensed under the Apache License, Version 2.0 (the "License");
@@ -95,6 +96,8 @@ namespace Config {
 		{"proc_info_smaps",		"#* Use /proc/[pid]/smaps for memory information in the process info box (very slow but more accurate)"},
 
 		{"proc_left",			"#* Show proc box on left side of screen instead of right."},
+
+        {"proc_filter_kernel",  "#* Filter processes tied to the Linux kernel(similar behavior to htop)."},
 
 		{"cpu_graph_upper", 	"#* Sets the CPU stat shown in upper half of the CPU graph, \"total\" is always available.\n"
 								"#* Select from a list of detected attributes from the options menu."},
@@ -222,6 +225,7 @@ namespace Config {
 		{"proc_mem_bytes", true},
 		{"proc_info_smaps", false},
 		{"proc_left", false},
+        {"proc_filter_kernel", false},
 		{"cpu_invert_lower", true},
 		{"cpu_single_graph", false},
 		{"cpu_bottom", false},
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -1,4 +1,3 @@
-// clang-format off
 /* Copyright 2021 Aristocratos (jakob@qvantnet.com)
 
    Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -96,7 +96,7 @@ namespace Config {
 
 		{"proc_left",			"#* Show proc box on left side of screen instead of right."},
 
-        {"proc_filter_kernel",  "#* Filter processes tied to the Linux kernel(similar behavior to htop)."},
+        {"proc_filter_kernel",  "#* (Linux) Filter processes tied to the Linux kernel(similar behavior to htop)."},
 
 		{"cpu_graph_upper", 	"#* Sets the CPU stat shown in upper half of the CPU graph, \"total\" is always available.\n"
 								"#* Select from a list of detected attributes from the options menu."},
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -1,3 +1,4 @@
+// clang-format off
 /* Copyright 2021 Aristocratos (jakob@qvantnet.com)
 
    Licensed under the Apache License, Version 2.0 (the "License");
@@ -598,6 +599,10 @@ namespace Menu {
 				" ",
 				"Will show percentage of total memory",
 				"if False."},
+            {"proc_filter_kernel",
+                "Filter kernel processes from output",
+                "", 
+                "Removes processes most users don't care about."},
 		}
 	};
 
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -599,9 +599,10 @@ namespace Menu {
 				"Will show percentage of total memory",
 				"if False."},
             {"proc_filter_kernel",
-                "Filter kernel processes from output",
-                "", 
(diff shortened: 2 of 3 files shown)


0xJoeMama:
I think everything is done now!

src/btop_config.cpp
```diff
@@ -1,3 +1,4 @@
+// clang-format off
```
aristocratos, line 1: Should be removed.

src/btop_menu.cpp
```diff
@@ -1,3 +1,4 @@
+// clang-format off
```
aristocratos, line 1: Should be removed.

src/linux/btop_collect.cpp
```diff
@@ -1,10 +1,11 @@
+// clang-format off
```
aristocratos, line 1: Should be removed.

src/btop_menu.cpp
```diff
@@ -598,6 +599,10 @@ namespace Menu {
 				" ",
 				"Will show percentage of total memory",
 				"if False."},
+            {"proc_filter_kernel",
+                "Filter kernel processes from output",
+                "", 
+                "Removes processes most users don't care about."},
```
aristocratos, line 605: This line is too long to fit in the options box, needs to be split with "about." on a new line.

src/btop_menu.cpp
```diff
@@ -598,6 +599,10 @@ namespace Menu {
 				" ",
 				"Will show percentage of total memory",
 				"if False."},
+            {"proc_filter_kernel",
+                "Filter kernel processes from output",
```
aristocratos, line 603: Should add a (Linux) tag unless you are gonna add support for FreeBSD and macos now also.

aristocratos (changes requested):
As far as I know we can safely assume that the pid of `kthreadd` is always gonna be `2`.

I think a more efficient way to go about this would be to add a:
 `static vector<size_t> kernel_pids = {2}` at the begging of `collect()`.

Then after line 1515 `const size_t pid = stoul(pid_str);`:
 `if (should_filter_kernel and v_contains(kernel_pids, pid)) continue;`

And below 1649 `pread.close();`:
`if (should_filter_kernel and new_proc.ppid == 2) {
	kernel_pids.push_back(new_proc.pid);
	current_procs.erase(rng::find(current_procs, new_proc.pid, &proc_info::pid));
	found.pop_back();
	continue;
}`

That way we aren't constantly adding and removing the kernel processes at every update.
Note that the code above is just an example, I haven't tested it.

src/btop_menu.cpp
0xJoeMama: I am aware of this. I actually implemented a fix locally but didn't have time to push.

src/btop_menu.cpp
0xJoeMama: I am not sure of how FreeBSD and OSX handle processes so I will keep it Linux only on my part. However, if anyone in the future wishes to add that functionality, they can just go ahead and do so. 

Also, I suppose by '(Linux) tag' you mean added '(Linux)' in the beginning of the line. Correct?

0xJoeMama:
> I think a more efficient way to go about this would be to add a...

I agree and I was actually thinking of ways to make this more efficient. However the making sure kthreadd is always pid 2 part is something I wasn't aware of. It is gonna help a lot.

0xJoeMama:
I don't think caching would work here since if we cache a kernel process that is restarted with a different id, that would mean any process that acquires the old kernel process id would be filtered out. That is why I just pop back the process vector in my version. Let me know if you have any other ideas.

src/linux/btop_collect.cpp
```diff
@@ -1629,6 +1635,11 @@ namespace Proc {
 				catch (const std::out_of_range&) { continue; }
 
 				pread.close();
+				
+				if (should_filter_kernel && new_proc.ppid == KTHREADD) {
+					current_procs.pop_back();
```
aristocratos, line 1640: `pop_back()` won't work if `should_filter_kernel` is set to true at a later time, since `new_proc` is only guaranteed to be the last added element when it doesn't already exist in `current_procs`. So a search is needed.

You also need a `continue` after `found.pop_back()` else you have operations being done on a dangling reference below.

aristocratos (changes requested):
I don't think a cache would be an issue, since PIDs are always incrementing a wrap around is very rare considering how high PID_MAX_LIMIT is on most systems. To make sure it doesn't happen you could add a counter that clears out the cache every 100 updates and all should be good.
Will still save alot of cpu by not collecting stats for processes that we will immediately discard.

src/linux/btop_collect.cpp
0xJoeMama: > So a search is needed.

Uhm, isn't it the case that a kernel process cannot be in the current_procs list already when we check for it?
Wouldn't it have been already filtered out?

0xJoeMama: The dangling reference thing, yeah it's true. I forgot the continue there.

aristocratos: Not when `should_filter_kernel` is a configurable value that can be toggled at any time in the options.
In the case that the config value is false and then later toggled to true the current code will restart removing the last element every time it finds a kernel process and it's very likely that the last element isn't the process we want and really weird behavior will result.

0xJoeMama: However I do agree with the saving of CPU time by skipping calculations. Therefore it seems a cached implementation would improve runtime. However, because of constant access time an unordered_set would be better in my opinion(assuming they are implemented with a backing array and not a linked_list). I will look into an implementation however I have started preparing for an exam, so it's gonna take some time.

aristocratos: I doubt you'll see much difference considering the size of the cache, but there is `robin_hood::unordered_set` already included you can use.

From https://github.com/martinus/robin-hood-hashing:
> Two memory layouts. Data is either stored in a flat array, or with node indirection. Access for unordered_flat_map is extremely fast due to no indirection, but references to elements are not stable. It also causes allocation spikes when the map resizes, and will need plenty of memory for large objects. Node based map has stable references & pointers (NOT iterators! Similar to [std::unordered_map](https://en.cppreference.com/w/cpp/container/unordered_map)) and uses `const Key` in the pair. It is a bit slower due to indirection. The choice is yours; you can either use `robin_hood::unordered_flat_map` or `robin_hood::unordered_node_map` directly. If you use `robin_hood::unordered_map` It tries to choose the layout that seems appropriate for your data.

0xJoeMama: The new changes should handle this case.

0xJoeMama:
I think the requested changes are done. Take a look and tell me if you need anything.

aristocratos:
Merged with some added optimizations.

Thanks for contributing!

Issue #352, opened 2022-06-04
[REQUEST] Makefile: Respect $(CXX) if set


NexAdn:
**Is your feature request related to a problem? Please describe.**

The current makefile doesn't respect a custom `CXX` variable.

**Describe the solution you'd like**

I'd like for the build system to respect `CXX` if set.

**Describe alternatives you've considered**

None. The alternative is patching the makefile for every custom case which isn't really a nice solution.

**Additional context**

The fix is quite easy (just replace `:=` with `?=` when setting `CXX` in the makefile). It might be useful when wanting to switch between compiler versions (e.g. to test other compilers) or when trying to do cross-compilation.
Support for this can also make testing clang compatibility for #324 easier.

aristocratos:
@NexAdn 
It does respect CXX, but not from the environment only when you use `CXX=` as a parameter when calling `make`.
This was a deliberate choice in the Makefile.
This is because GCC is currently the only compiler with the C++20 support needed to compile btop++.
The CXX flag for make is also referenced as a working flag in https://github.com/aristocratos/btop#compilation-linux

> Support for this can also make testing clang compatibility for https://github.com/aristocratos/btop/issues/324 easier.

`make CXX=clang` however, the settings for the C++ library also need to be changed to compile with clang.
Regardless, when clang has support for C++20 Ranges, I will add support for clang with the correct flags switched in the Makefile.

aristocratos:
@FederAndInk 
You can easily override this by just calling `make CXX="${CXX}"` if you need to use the environment variable.

But I would argue that convenience is better left for the 90%+ of people that just want to type `make` and have a working binary.
Rather than the minority that are for example cross compiling (and also a lot more likely to know how to customize the compilation).

I mean, the inconvenience here is that you have to specify what compiler to use if you don't want the Makefile to find the best match.

FederAndInk:
Sure no worries, thanks for the answer :)

It still took me some time before figuring it out, as I was working in an environment with CXX set up. I had to read the Makefile to see why it didn't work and then came across this issue.

Maybe this could be added to the Makefile? As a colored note when showing the summary e.g.

```bash

██████╗ ████████╗ ██████╗ ██████╗
██╔══██╗╚══██╔══╝██╔═══██╗██╔══██╗   ██╗    ██╗
██████╔╝   ██║   ██║   ██║██████╔╝ ██████╗██████╗
[...]

PR #353, opened 2022-06-04
Makefile: Don't override user-provided CXX var

NexAdn:
Allows running the build with compilers other than the g++ executables
found in $PATH.

Closes: #352


aristocratos:
As noted in Issue #352 this is not correct.
User provided CXX parameters to make are respected, but not CXX set in environment.

PR #359, opened 2022-06-07
Add everforest-dark-hard.theme

iambeingtracked:
Add an everforest dark hard theme


aristocratos (approved):
Looks great, thanks for contributing!

PR #385, opened 2022-07-04
Implement new ZFS pool io monitoring and the option to show ZFS pools only

### LINKED ISSUE #380: show ZFS pools instead of dataset in the disk usage listing
It seems like we show ZFS datasets in the per-filesystem disk usage right now. Datasets can be quite numerous on a given system. In my case, I have *21* datasets:

```
anarcat@curie:~$ zfs list
NAME                                                                                    USED  AVAIL     REFER  MOUNTPOINT
bpool                                                                                   358M   473M       96K  /boot
bpool/BOOT                                                                              358M   473M       96K  none
bpool/BOOT/debian                                                                       358M   473M      132M  /boot
rpool                                                                                   268G   631G      192K  /
rpool/ROOT                                                                             18.1G   631G      192K  none
rpool/ROOT/debian                                                                      18.1G   631G     16.5G  /
rpool/home                                                                              245G   631G      232G  /home
rpool/home/root                                                                        2.18G   631G     1.81G  /root
rpool/var                                                                              4.70G   631G     1.82G  /var
rpool/var/03a8e9cd3579e3d73ecc3f884b6d7dd1ec54cc5a542abcaadde7a51ae619a6f9              114M   631G      118M  legacy
rpool/var/5216338b40a7b96416b8b9858974bbe4acc3096ee60acbc4dfb1ee02aecceb10             5.09M   631G     5.08M  legacy
rpool/var/a014e33c4e6e136dfa23824a52e4af69ef9356e7bdd41325da43d774dc960426              264K   631G      118M  legacy
rpool/var/c4261d7173a1d4c1f85b680ecdddd9f41cb18544d08e700d7ff3aedf90d3f187              224K   631G      118M  legacy
rpool/var/cache                                                                         565M   631G      254M  /var/cache
rpool/var/lib                                                                          1.59G   631G      192K  /var/lib
rpool/var/lib/docker                                                                    354M   631G      173M  /var/lib/docker
rpool/var/lib/docker/8888b2f382e279201ac9b4cbebcd1ec7c0ff0c815d4bdc205d51cdc5a46b0427   114M   631G      118M  legacy
rpool/var/lib/docker/90689560f8398b2bf97e4355acf19abe4b8dbadde9f7a1a985919ba9738a7748  5.09M   631G     5.08M  legacy
rpool/var/lib/flatpak                                                                  1.24G   631G      900M  /var/lib/flatpak
rpool/var/tmp                                                                           217M   631G      216M  /var/tmp
```

And that's probably on the low end of things. You can basically create as many datasets as you want, and it's common to have dozens of them on a busy workstation.

Wouldn't it be better to show pool usage instead? For example, here's the listing for my workstation:

```
anarcat@curie:~$ zpool list
NAME    SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
bpool   960M   359M   601M        -         -     3%    37%  1.00x    ONLINE  -
rpool   928G   268G   660G        -         -    13%    28%  1.00x    ONLINE  -
```

nice and clean. It also shows the health of the pool, critically.

_That this wasn't supported already was clarified by @simplepad in https://github.com/aristocratos/btop/issues/374#issuecomment-1167564626_

### LINKED ISSUE #383: [REQUEST] zpool stats monitoring method change.
Per #374 which is monitoring ZFS file system via `/proc/spl/kstat/zfs/*pool_name*/io` this will not work for long term monitoring.  This has already been removed from current builds of ZFS:

```
$ zfs version
zfs-2.1.5-1
zfs-kmod-2.1.5-1

$ ls -l /proc/spl/kstat/zfs/vault
total 0
-rw-r--r-- 1 root root 0 Jun 30 17:40 dmu_tx_assign
-rw-r--r-- 1 root root 0 Jun 30 17:40 iostats
-rw-r--r-- 1 root root 0 Jun 30 17:40 multihost
-rw-r--r-- 1 root root 0 Jun 30 17:40 objset-0x183
-rw-r--r-- 1 root root 0 Jun 30 17:40 objset-0x195b1
-rw-r--r-- 1 root root 0 Jun 30 17:40 objset-0x36
-rw-r--r-- 1 root root 0 Jun 30 17:40 objset-0x3699
-rw-r--r-- 1 root root 0 Jun 30 17:40 objset-0x86
-rw------- 1 root root 0 Jun 30 17:40 reads
-rw-r--r-- 1 root root 0 Jun 30 17:40 state
-rw-r--r-- 1 root root 0 Jun 30 17:40 txgs
```

These links appear relevant to its removal:
[https://github.com/openzfs/zfs/pull/12212](https://github.com/openzfs/zfs/pull/12212)
[https://github.com/openzfs/zfs/commit/371f88d96fe0aeb46a72fec78f90e1d777493ee5](https://github.com/openzfs/zfs/commit/371f88d96fe0aeb46a72fec78f90e1d777493ee5)

Prometheus Node Exporter for ZFS was using this as well:
[https://github.com/prometheus/node_exporter/issues/2068](https://github.com/prometheus/node_exporter/issues/2068)

simplepad:
This PR implements new ZFS pool io monitoring, compatible with newer versions of ZFS, where `/proc/spl/kstat/zfs/*pool_name*/io` files are not present anymore. Instead, it uses `/proc/spl/kstat/zfs/*pool_name*/objset-*` files, which allows for per dataset io stats. 

Also, this PR adds the `zfs_pools_only` option, which if enabled will hide all datasets and only show pools in disk listings, and will count io stats differently.
In more detail:
When `zfs_pools_only` is set to true:
- In disk listings, only ZFS pools are shown
- IO stats are calculated per pool by combining the io of all objects in the pool (all `objset-*` files in the pool directory, pool root included)

When `zfs_pools_only` is set to false:
- In disk listings, both ZFS pools and datasets are shown
- IO stats are calculated per object, that is separately for ZFS pool root and each dataset (operations on datasets are not calculated towards pool root io)

Closes #383 
Partially closes #380, pools health monitoring isn't implemented


src/linux/btop_collect.cpp
```diff
@@ -26,6 +26,7 @@ tab-size = 4
 #include <netdb.h>
 #include <ifaddrs.h>
 #include <net/if.h>
+#include <dirent.h>
```
aristocratos, line 29: I take it you're coming from C programming :)

We already got `filesystem` included which has `fs::directory_iterator()` and `fs::recursive_directory_iterator()` which I believe will do what you want to achieve with `dirent`.

Search for "fs::directory_iterator" to see some examples in btop_collect.cpp
The `fs::path` class also has some really useful functions for stripping filenames, basepaths and so on.

src/btop_shared.hpp
```diff
@@ -160,6 +160,12 @@ namespace Mem {
 
 	//* Draw contents of mem box using <mem> as source
 	string draw(const mem_info& mem, const bool force_redraw=false, const bool data_same=false);
+
+	//?* Find the filepath to the specified ZFS object's stat file
```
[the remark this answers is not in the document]
simplepad: Agreed, fixed.

src/linux/btop_collect.cpp
simplepad: Re-implemented using fs::path.

trentbuck:
FYI

I tested `Debian 11` / `Linux 5.16.0-0.bpo.4-amd64` / `ZFS 2.1.4-1~bpo11+1`.
With `Use fstab = False`,
1. https://github.com/simplepad/btop/commit/4969dd8 shows I/O sparklines for ZFS datasets
2. https://github.com/aristocratos/btop/commit/6b1b9f8 does not show I/O sparklines for ZFS datasets

With `Zfs pools only = True`, my datasets disappear, but my pool doesn't appear.
I saw this on three hosts, all set up roughly along the lines of https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Bullseye%20Root%20on%20ZFS.html
They import the pool from inside the initramfs-tools initrd (before zed is running).
They run refind (not grub), and do not have a bpool.  Instead /boot is rsynced into /boot/efi/EFI/debian via this crap https://github.com/trentbuck/flash-kernel-efi/tree/ansible

simplepad:
@trentbuck Does it show your pool when `Zfs pools only = False`?

simplepad:
Also, does anything come up in your `~/.config/btop/btop.log`?

trentbuck:
@simplepad I don't ever see the pool in btop; both `zfs_pools_only = False` and `zfs_pools_only = True`  in `~/.config/btop/btop.conf` just show/hide the datasets.  Earlier, I wrongly thought it meant "show/hide the pool" rather than "show/hide datasets".  I did not edit btop.conf directly; I used `m > Options > mem` from inside btop.

For the avoidance of doubt, I *do* see the pool in `zpool list`, and I see the datasets in `zfs list`.

My btop.log has some entries but they look boring.
[btop.log](https://github.com/aristocratos/btop/files/9055691/btop.log)

I tested both as my regular user, and as root - I did not see any difference in behaviour.  My regular user normal Debian/GNOME defaults: sudo and polkit privileges, but no zfs-allow(8) privileges.
I can't think of anything else "weird" that would cause a pool to not show up.
I haven't actually straced btop or read the code, yet.

trentbuck:
PS: here is `grep -rnH ^ /proc/spl/kstat/zfs` run as the same regular user that runs btop:
[grep.typescript.txt](https://github.com/aristocratos/btop/files/9055756/grep.typescript.txt)
The pool's name is `hera`, and the objset files are clearly readable.

I wonder if the issue is that my pool's root dataset is not mounted?

```
$ zfs list
NAME                                    USED  AVAIL     REFER  MOUNTPOINT
hera                                    152G  72.5G      192K  none
hera/goll                              9.43G  6.57G     9.43G  /srv/backup/goll
hera/hera                               108G  72.5G     8.31G  /
hera/hera/home                         87.4G  72.5G      320K  /home
hera/hera/home/twb                     87.4G  12.9G     51.1G  /home/twb
hera/hera/var                          2.36G  72.5G      320K  /var
hera/hera/var/cache                    1.99G  72.5G     1.18G  /var/cache
hera/hera/var/tmp                       380M  7.71G      294M  /var/tmp
hera/not-backed-up                     34.7G  72.5G      192K  none
hera/not-backed-up/apt-cacher-ng       9.59G  2.41G     9.59G  /var/cache/apt-cacher-ng
hera/not-backed-up/tweak.prisonpc.com  25.1G  22.9G     4.60G  /var/lib/libvirt/images/twoke.d

```

simplepad:
@trentbuck Can you confirm that your pool appears when using btop 1.2.8?

trentbuck:
Ah, sorry - I was unclear again.
I have *never* seen btop show my pool on any host (I tried 3 different hosts, all Debian 11).
I (incorrectly?) thought that was expected on v1.2.8 because, prior to your PR, btop was looking in the wrong place?

With btop v1.2.8, I don't see "hera" (pool) listed under disks.  I see datasets with only Used/Free (no I/O)
Screenshot from 2022-07-07 06-58-01 - btop v1 2 8
With your commit, I see I/O tickers, but still no pool
Screenshot from 2022-07-07 07-00-08 btop v1 2 8-5-g4969dd8

trentbuck:
PS: I have `use_fstab = False` because otherwise I don't see any ZFS datasets at all.
My fstab is only has legacy crap in it -- /boot (ext4), /boot/efi (vfat), &c.
All my ZFS mounts are specified as attributes of the ZFS datasets and get mounted automatically via /usr/lib/systemd/system-generators/zfs-mount-generator, which creates systemd .mount units and bypasses /etc/fstab entirely.  They do show up in /proc/self/mountinfo and /proc/self/mounts.  Per /lib/tmpfiles.d/debian.conf, /etc/mtab is a symlink to ../proc/self/mounts.

```
$ grep -c zfs /etc/fstab /etc/mtab /proc/self/mountinfo /proc/self/mounts
/etc/fstab:0
/etc/mtab:6
/proc/self/mountinfo:6
/proc/self/mounts:6

```

simplepad:
I get what you want, but currently btop acts like this: no mountpoint == disk won't be detected and won't be shown. 

>The pool's name is hera, and the objset files are clearly readable.

I looked closely to your output of `grep -rnH ^ /proc/spl/kstat/zfs` command, and I couldn't find the hera root object file in there.

src/linux/btop_collect.cpp
```diff
@@ -1094,6 +1128,119 @@ namespace Mem {
...
+				}
+				filestream.close();
+			}
+		}
+
+		Logger::warning("Could not read directory: " + zfs_pool_stat_path.string());
+		return "";
+	}
+
+	bool zfs_collect_pool_total_stats(struct disk_info &disk) {
```
aristocratos, line 1182: What does this bool indicate? The function currently only returns true.

src/linux/btop_collect.cpp
```diff
@@ -1094,6 +1128,119 @@ namespace Mem {
...
+
+					// skip characters until '4' is reached, indicating data type 4, next value will be out target
+					diskread.ignore(numeric_limits<streamsize>::max(), '4');
+					diskread >> io_ticks;
+					io_ticks_total += io_ticks;
+
+					// skip characters until '4' is reached, indicating data type 4, next value will be out target
+					diskread.ignore(numeric_limits<streamsize>::max(), '4');
+					diskread >> bytes_read;
+					bytes_read_total += bytes_read;
```
aristocratos, line 1212: Should probably wrap this in a `try` block with a `continue`in the `catch`, so we don't crash if there is an exception thrown from reading after `ignore` or from the `int64_t`conversions.

src/linux/btop_collect.cpp
simplepad: This bool indicates whether or not `disk_ios` variable should be incremented indicating a successful ZFS pool io collection. I modified the functions so now it checks that at least one object was read before updating the io stats and returning true.

src/linux/btop_collect.cpp
simplepad: Done.

aristocratos:
@simplepad 
Nice work!

Do you feel this PR is ready or do you wanna do more testing?

The logging functions should be ok to keep if you change them from warning to debug, that way we can still get some useful output if someone is having issues.
If you want to keep them as warnings I would ask that you set some limiter on them so we don't spam the logfile in case of errors.

simplepad:
@aristocratos 
> The logging functions should be ok to keep if you change them from warning to debug, that way we can still get some useful output if someone is having issues.

We could replace them with debug calls, but users won't know something is wrong until they change their log level. On the other hand, it's not like they would randomly want to check the log file, since as far as I know, btop has no error indication in the UI, so I'm not against hiding it behind debug.

> Do you feel this PR is ready or do you wanna do more testing?

The only thing that's stopping me from marking this PR ready to review is that the `zfs_pools_only` option's name and description might be a little confusing and hard to understand, as shown by [trentbuck](https://github.com/trentbuck) in their comments above. Maybe we should change it to `zfs_hide_datasets`?
Also, this PR changes the ZFS io collection to be compatible with the new openzfs versions, but technically, if the older files we use in btop 1.2.8 are still present and the `zfs_pools_only` option is enabled, it's much faster to use them because then we only need to read one file (as opposed to the new version where we read all objects files that comprise the pool and calculate the sum of all of their stats). Should I keep the old way to read stats around for the odd occasion that someone with the older version of openzfs will be able to use it? I am not sure how much faster it is and how much of a benefit that is.

trentbuck:
> `zfs_pools_only` option's name and description might be a little confusing and hard to understand, as shown by [trentbuck](https://github.com/trentbuck) in their comments above. Maybe we should change it to `zfs_hide_datasets`?

+1 for `zfs_hide_datasets`, since that reflects the current semantics.
(Or, keep the variable name, but change the semantics to match :-)

> Also, this PR changes the ZFS io collection to be compatible with the new openzfs versions, but technically, if the older files we use in btop 1.2.8 are still present and the `zfs_pools_only` option is enabled, it's much faster to use them because then we only need to read one file (as opposed to the new version where we read all objects files that comprise the pool and calculate the sum of all of their stats). Should I keep the old way to read stats around for the odd occasion that someone with the older version of openzfs will be able to use it? I am not sure how much faster it is and how much of a benefit that is.

Do you know what version of OpenZFS removed the "old way" files?  From the below, I guess it's in 2.1.x.
- Debian 11 / ZFS 2.1.4 --- only `/proc/spl/kstat/zfs/*morpheus*/objset*`
- Debian 10 / ZFS 2.0.3 --- both
- Debian 9 / ZFS 0.7.12 --- only `/proc/spl/kstat/zfs/morpheus/io`

simplepad:
@trentbuck 
>Do you know what version of OpenZFS removed the "old way" files? From the below, I guess it's in 2.1.x.

Well, here's the commit that removes it https://github.com/openzfs/zfs/commit/371f88d96fe0aeb46a72fec78f90e1d777493ee5, so yeah, probably 2.1.0

>Debian 11 / ZFS 2.1.4 --- only /proc/spl/kstat/zfs/*morpheus*/objset*

Strange, I have a Debian 11 server with ZFS installed and the stable version there is 2.0.3

aristocratos:
@simplepad 
> We could replace them with debug calls, but users won't know something is wrong until they change their log level. On the other hand, it's not like they would randomly want to check the log file, since as far as I know, btop has no error indication in the UI, so I'm not against hiding it behind debug.

Well, an indication would be that the zfs io isn't showing up leading to them hopefully creating a bug report. And then if they haven't provided a debug logfile we can always ask for one.

I've been thinking about adding some more functionality to the logger, for example optional parameters "log_identifier" and "max_reported". To easily be able to mute log entries of the same "log_identifier".
That way warnings can still be generated for non critical issues without filling up the logfile and adding unnecessary io operations and processing for those who choose to ignore them.

> The only thing that's stopping me from marking this PR ready to review is that the zfs_pools_only option's name and description might be a little confusing and hard to understand, as shown by [trentbuck](https://github.com/trentbuck) in their comments above. Maybe we should change it to zfs_hide_datasets?

I haven't really used zfs so can't really tell what the best terminology for a regular user would be, but in general I would guess having "hide" in the name rather than "only" is probably better. There are several options already in the program that probably should be changed to better clarify what they actually do...

> Also, this PR changes the ZFS io collection to be compatible with the new openzfs versions, but technically, if the older files we use in btop 1.2.8 are still present and the zfs_pools_only option is enabled, it's much faster to use them because then we only need to read one file (as opposed to the new version where we read all objects files that comprise the pool and calculate the sum of all of their stats). Should I keep the old way to read stats around for the odd occasion that someone with the older version of openzfs will be able to use it? I am not sure how much faster it is and how much of a benefit that is.

Unless we are talking 100+ files that are being read compared to the "old" way, I don't think will be noticeable performance wise. But I would say it's up to you if you wanna keep both ways.
You could do some simple testing with just the timer shown for "mem" in debug mode and see if you notice any discernable difference between the two ways and if it's worth keeping both.

simplepad:
Although I don't really have that many datasets I did some testing and there wasn't any noticeable difference between the two ways. After all, the files we're reading are stored in `/proc`, which is a virtual filesystem stored in memory, so the reads are probably fast enough that we can leave the code as is. 

Marking this PR as Ready for review.

aristocratos (approved):
Great work man!

And thank you for the continued contributions!

trentbuck:
> Strange, I have a Debian 11 server with ZFS installed and the stable version there is 2.0.3

It's in bullseye-backports: https://packages.debian.org/search?keywords=zfs-dkms

You can put something like this in `/etc/apt/preferences.d/zfs-is-cool`:
```
Explanation: use Linux 5.14+ (not 5.10)
Package: src:linux src:linux-signed-amd64 src:zfs-linux
Pin: release a=bullseye-backports
Pin-Priority: 500
```

trentbuck:
> Unless we are talking 100+ files that are being read compared to the "old" way, I don't think will be noticeable performance wise.

FWIW, I think 10 is typical for a single-user laptop, and 100 to 1000 is typical for a multi-user server.

AFAICT you get one objset file for each mounted dataset.  On a server with 20 users (thus at least 40 datasets mounted -- one for each $HOME and one for each $MAIL), I see:
```
cyber@heavy:~$ zfs list | wc -l
109
cyber@heavy:~$ mount -t zfs | wc -l
95
cyber@heavy:~$ ls -l /proc/spl/kstat/zfs/*/*objset* | wc -l
95
```
On a single-user laptop, those numbers are all `6`.
On a backup server, `9` objsets/mounts, and `66` datasets (because ZFS doesn't need to decrypt or mount the incremental backups).

At the default refresh rate of 0.5Hz, I doubt 100 files is a big deal.
In any case, the "old way" will be gone from all ZFS systems eventually.
If I/you/we are worried about performance, we can talk to the ZFS developers -- maybe they will add an efficient path just for btop :-)

simplepad:
@trentbuck I see, we should be good then.

dm17:
I'm wondering if ZFS users would like to see ZFS errors in btop, or if that is useless as they monitor those another way... I figure this would be a decent place to ask.

Issue #333, opened 2022-05-17
[BUG] In case of zram swap, the used and free values go way above normal levels


Firsh:
I have a Raspberry Pi which has only 1GB RAM therefore I'm using a special kind of swap which is [zram](https://github.com/foundObjects/zram-swap) (compresses swap contents and is in memory not disk/sdcard). Maybe that makes reporting funny, when starting and stopping a lot of stuff so the memory usage goes up, the numbers go way above the configured size, in the multiple GiB territory which doesn't seem real. As soon as I press 2 twice to toggle the view, the numbers revert to something that you'd get by multiplying the percentage value with the total "Swap:" amount. As you can see I've reconfigured swap to be displayed under memory and not among disks.

**To Reproduce**

I run many things on the server to the point where CPU is at 100% and the combined memory almost runs out, then I close them (they are vscode remote instances, like 5 of them). The numbers are fine before the usage drops, but then they get weird:

![BMiB](https://i.justifiedgrid.com/rc1bvl.png)

**Expected behavior**

I'd like to see values in the `Used:` and `Free:` that are smaller than the `Total:`. After "cleaning" (pressing 2 twice, they revert):

![good](https://i.justifiedgrid.com/rc1bia.png)

**Screenshots**

![bad](https://i.justifiedgrid.com/rc1bbw.png)

**Info (please complete the following information):**
 - btop++ version: 1.2.6
 - Binary: [btop-aarch64-linux-musl.tbz](https://github.com/aristocratos/btop/releases/download/v1.2.6/btop-aarch64-linux-musl.tbz)
 - Architecture: aarch64
 - Platform: Linux
 - (Linux) Kernel: 5.15.0-1006-raspi
 - Terminal used: bash via ssh in Windows 11 PowerShell
 - Font used: Fira Code

**Additional context**

contents of `~/.config/btop/btop.log`

```
2022/05/17 (17:25:10) | ===> btop++ v.1.2.6
2022/05/17 (17:25:10) | INFO: Logger set to INFO

2022/05/17 (17:57:10) | ===> btop++ v.1.2.6
2022/05/17 (17:57:10) | ERROR: Stall in Runner thread, restarting!

2022/05/17 (18:15:49) | ===> btop++ v.1.2.6
2022/05/17 (18:15:49) | ERROR: Stall in Runner thread, restarting!
```

tie:
I think the issue here is that `btop` shows memory usage similar to what `free` reports (i.e. uncompressed size), e.g.

```
[...]

PR #393, opened 2022-07-16
Fix decreasing memory view mismatch

vanishacmp:
fix for bug
https://github.com/aristocratos/btop/issues/333
when memory decreasing it produce 
"1023.5 Mib " =>
"1003.4 Mib" instead "3.4 Mib"
"1003.Byte" instead "0 Byte"
and etc


aristocratos:
Fixed in v1.2.9

PR #432, opened 2022-09-27
Update README.md

franciosi:
s/OSX/macOS on where it makes sense


franciosi:
I will make another one.

PR #436, opened 2022-10-02
Code Cleanup

stefanos82:
There are lots of things that need cleanup but I think we can do this, bit by bit.


aristocratos (approved):
Looks good 👍🏼

PR #437, opened 2022-10-04
Fix variable initialization to 'auto' for getB()

stefanos82:
It does not make sense to return a `const bool&` as we are dealing with Boolean literals that are `prvalue`s of type bool.

The compiler is smart enough to do the necessary optimizations wherever is applicable.

@aristocratos can you double-check the whole code in case it breaks somewhere? I have updated the code for OS X and FreeBSD, but I don't use those two aforementioned OS-es to test them myself.

### DIFF
--- a/src/btop_config.hpp
+++ b/src/btop_config.hpp
@@ -113,10 +113,3 @@ namespace Config {
 	//* Write the config file to disk
 	void write();
 }
-
-
-
-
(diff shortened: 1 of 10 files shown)


aristocratos (approved):
Looks good. No issues on macOS or FreeBSD either.

PR #452, opened 2022-10-28
proc tree view: if there's more than 40 width left, print full cmd

Superty:


aristocratos:
https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md

Superty:
PR for #453

aristocratos (approved):
Looks good!
Thanks for contributing :)

stefanos82:
It does not produce a clean output on my monitor; it looks rather broken, just take a look at the left side of expanded process tree and the width of parentheses on the far right side.

image

stefanos82:
Another issue I have noticed is in the footer area; where is `select`, `info`, and the whole word for `terminate`?

image

Superty:
Thanks for reporting, should be fixed by #454.

stefanos82:
My maximum resolution is `1360x768` if it helps; it's presumably an HDTV monitor.

PR #454, opened 2022-10-28
proc tree: fix display bug when displaying full cmd

Superty:
Also don't show the cmd if it's empty, this seems to happen for kernel processes. Also update width threshold to 40.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1474,7 +1474,7 @@ namespace Proc {
 					width_left -= (ulen(p.name) + 1);
 				}
 				if (width_left > 7) {
-					const string& cmd = width_left > 40 ? p.cmd : p.short_cmd;
+					const string& cmd = width_left > 40 ? rtrim(p.cmd) : p.short_cmd;
 					if (not cmd.empty() and cmd != p.name) {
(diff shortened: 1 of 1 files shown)


src/btop_draw.cpp
```diff
@@ -1473,9 +1473,12 @@ namespace Proc {
 					out += c_color + uresize(p.name, width_left - 1) + end + ' ';
 					width_left -= (ulen(p.name) + 1);
 				}
-				if (width_left > 30 || (width_left > 7 && p.short_cmd != p.name)) {
-					out += g_color + '(' + uresize(width_left > 30 ? p.cmd : p.short_cmd, width_left - 3, p_wide_cmd[p.pid]) + ") ";
-					width_left -= (ulen(p.short_cmd, true) + 3);
+				if (width_left > 7) {
+					const string& cmd = width_left > 40 ? p.cmd : p.short_cmd;
```
aristocratos, line 1477: I would suggest changing `p.cmd` to `rtrim(p.cmd)` here to clean up the leftover whitespace many kernel processes have in the command line.

aristocratos (changes requested):

aristocratos (approved):

PR #457, opened 2022-11-04
Show the first IP of the interface in NET box [on Linux]

correabuscar:
... instead of the last.

Also, indented the `for` statement with tabs rather than spaces.

Closes #456

<details>
<summary>irrelevant details here</summary>

Notes:
I moved up the `if (not v_contains(interfaces, iface)) {` because I'm also able to use it to reset/empty the IP string value for the current interface in the `for` loop, thus allowing me to detect the first time this specific `iface` is seen in the IPv4/IPv6 context and thus assigning only the first encounter of the IP address for this `iface` while parsing getifaddrs() results, even if the results aren't grouped by the `iface`.

This PR, unlike the OP patch in #456, will continuously refresh the interfaces and their IPs, and thus is still sensitive to interface IP changes.
</details>

Side Note: I'll make another PR for the `inet_ntop` part of #456 ideally after this PR is merged(to avoid unnecessary conflicts), and it's okay if this PR closes it.

PR #458, opened 2022-11-04
Replace  getnameinfo  with  inet_ntop [on Linux]

correabuscar:
saves 979 bytes of reserved buffer because:
NI_MAXHOST is 1025 bytes
and
INET6_ADDRSTRLEN is 46

Depends on PR #457 being merged first.

I'm not sure how, having this PR created before PR #457 is merged, will work, but we'll see what I'll have to do(git-wise) to not mess things up. I'm not against making it a new PR if this one is breaking things too much (ie. because the commit from PR #457 is included in this PR, which can't be good)

<s>**EDIT:** I didn't physically test the IPv6 part (I've disabled ipv6 in too many places, but maybe I will do it in a virtualbox)</s>
**EDIT2:** I've now tested the IPv6 part too, `btop` shows IPv6 only if there's no IPv4 and only if the window is horizontally large enough to show the numeric IPv6. For some reason the previous behavior was different in that: 1. removing the IPv4 from the interface while `btop` is running, keeps showing the old IPv4, 2. the IPv6 shown has "%devnamehere" appended to it (eg. "%enp0s3"), but otherwise with this PR it behaves as I would normally expect it, fixing the above 2 issues. So, I guess, there's more good reasons to use `inet_ntop` than I had previously thought.

**EDIT3:** When using `gnome-control-center network` (eg. a shell command) to configure the interface, (manually-, only? probably not)assigning the interface an IP in the block 169.254.0.0/16 will cause this IP to be first when seen by `ip addr` (command) and thus `btop` shows this as the IP (due to PR #457) but old `btop` shows the last IP which would be the right one, in this case. This is because https://en.wikipedia.org/wiki/Private_network#Link-local_addresses and I guess this is why they're first in the list. So this only happens for the 169.254.0.0/16 IPs.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -26,6 +26,8 @@ tab-size = 4
 #include <netdb.h>
 #include <ifaddrs.h>
 #include <net/if.h>
+#include <arpa/inet.h> // for inet_ntop()
+
 
 #if !(defined(STATIC_BUILD) && defined(__GLIBC__))
 	#include <pwd.h>
@@ -1388,7 +1390,13 @@ namespace Net {
 				return empty_net;
 			}
 			int family = 0;
-			char ip[NI_MAXHOST];
+			static_assert(INET6_ADDRSTRLEN >= INET_ADDRSTRLEN); // 46 >= 16, compile-time assurance.
+#if defined(IPBUFFER_MAXSIZE)
+#error Overwriting a previous macro with the same name: IPBUFFER_MAXSIZE, this means you likely need to rename this one here!
+#else
+#define IPBUFFER_MAXSIZE (INET6_ADDRSTRLEN) // manually using the known biggest value, guarded by the above static_assert
+#endif
+			char ip[IPBUFFER_MAXSIZE];
 			interfaces.clear();
 			string ipv4, ipv6;
 
@@ -1410,19 +1418,35 @@ namespace Net {
 					net[iface].ipv6.clear();
 				}
 
+
 				//? Get IPv4 address
 				if (family == AF_INET) {
-					if (net[iface].ipv4.empty() and 
-						getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in), ip, NI_MAXHOST, NULL, 0, NI_NUMERICHOST) == 0)
-						net[iface].ipv4 = ip;
+					if (net[iface].ipv4.empty()) {
+						if (NULL != inet_ntop(family, &(reinterpret_cast<struct sockaddr_in*>(ifa->ifa_addr)->sin_addr), ip, IPBUFFER_MAXSIZE)) {
+							net[iface].ipv4 = ip;
+						} else {
+							int errsv = errno;
+							Logger::error("Net::collect() -> Failed to convert IPv4 to string for iface " + string(iface) + ", errno: " + strerror(errsv));
+						}
+					}
 				}
 				//? Get IPv6 address
 				else if (family == AF_INET6) {
-					if (net[iface].ipv6.empty() and 
-						getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in6), ip, NI_MAXHOST, NULL, 0, NI_NUMERICHOST) == 0)
-						net[iface].ipv6 = ip;
+					if (net[iface].ipv6.empty()) {
+						if (NULL != inet_ntop(family, &(reinterpret_cast<struct sockaddr_in6*>(ifa->ifa_addr)->sin6_addr), ip, IPBUFFER_MAXSIZE)) {
+							net[iface].ipv6 = ip;
+						} else {
+							int errsv = errno;
+							Logger::error("Net::collect() -> Failed to convert IPv6 to string for iface " + string(iface) + ", errno: " + strerror(errsv));
+						}
+					}
 				} //else, ignoring family==AF_PACKET (see man 3 getifaddrs) which is the first one in the `for` loop.
 			}
+#if defined(IPBUFFER_MAXSIZE)
+#undef IPBUFFER_MAXSIZE
+#else
+#error whoa, the programmer forgot something, eg. was this renamed?
+#endif
 
 			//? Get total recieved and transmitted bytes + device address if no ip was found
 			for (const auto& iface : interfaces) {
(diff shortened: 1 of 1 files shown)


src/linux/btop_collect.cpp
```diff
@@ -1388,33 +1390,63 @@ namespace Net {
 				return empty_net;
 			}
 			int family = 0;
-			char ip[NI_MAXHOST];
+			static_assert(INET6_ADDRSTRLEN >= INET_ADDRSTRLEN); // 46 >= 16, compile-time assurance.
+#if defined(IPBUFFER_MAXSIZE)
+#error Overwriting a previous macro with the same name: IPBUFFER_MAXSIZE, this means you likely need to rename this one here!
+#else
+#define IPBUFFER_MAXSIZE (INET6_ADDRSTRLEN) // manually using the known biggest value, guarded by the above static_assert
```
correabuscar, line 1397: All these `#` lines could be replaced with an a one line `enum` constant, like `enum { IPBUFFER_MAXSIZE = INET6_ADDRSTRLEN }` (not tested yet). I wanted something with the same effect of not using any storage and not allowing for other side-effects. I tried to find a good doc about it, but this is good enough: https://stackoverflow.com/a/33232410/19999437

aristocratos:
Oh right, if you rebase now after #457 is merged, I think it should be good and both could be reverted without reverting the other, I think..

aristocratos:
> EDIT3: When using gnome-control-center network (eg. a shell command) to configure the interface, (manually-, only? probably not)assigning the interface an IP in the block 169.254.0.0/16 will cause this IP to be first when seen by ip addr (command) and thus btop shows this as the IP (due to PR https://github.com/aristocratos/btop/pull/457) but old btop shows the last IP which would be the right one, in this case. This is because https://en.wikipedia.org/wiki/Private_network#Link-local_addresses and I guess this is why they're first in the list. So this only happens for the 169.254.0.0/16 IPs.

Huh, will have to implement a manual IP selection (from detected IP's in a list) or similar if people start complaining...

correabuscar:
> > EDIT3: When using `gnome-control-center network` (eg. a shell command) to configure the interface, (manually-, only? probably not)assigning the interface an IP in the block 169.254.0.0/16 will cause this IP to be first when seen by ip addr (command) and thus btop shows this as the IP (due to PR #457) but old btop shows the last IP which would be the right one, in this case. This is because https://en.wikipedia.org/wiki/Private_network#Link-local_addresses and I guess this is why they're first in the list. So this only happens for the 169.254.0.0/16 IPs.
> 
> Huh, will have to implement a manual IP selection (from detected IP's in a list) or similar if people start complaining...
> 
> Edit: Actually, we could probably add an exception to the `169.254.0.0` IP range if that's the only one that behave this way? Something like `if (net[iface].ipv4.empty() and ! net[iface].ipv4.starts_with("169.254"))` to continue to the next one.**

I think this shouldn't be an issue, because if it happens [automatically](https://serverfault.com/a/132716), then the 169.254.* IP is right to be the first, but if the IP is added manually by user(\*), then I think that maybe `gnome-control-center network` is wrong to set it in the `link-local` scope, thus it should set in the `global` scope, just like any other IP. In Gentoo, I set such an IP manually and it's like any other private network IP, in the `global` scope, and it's thus NOT first in the list no matter how many other private network IPs I've added.

\* (such as for when wanting to access a router directly (on eg. 169.254.100.100 (which QNAP NAS uses), or `169.254.1.250 is mainly used by brands such as Linksys and Strix Systems.`, and my router always had such an IP(can't remember which exactly) even when using the normal private network IPs  and access to the router through such an IP(eg. 169.254.100.100) would work even when a local IP wasn't set in the same 169.254.0.0/16 block, if I remember correctly) )

I think maybe wait until people complain, to see their use of it, then if still want to add an exception, doing it in a new PR.

** modded quote.

aristocratos (approved):
Looks good! Compiles fine for all release platforms and should be easily noticeable in the [build action](https://github.com/aristocratos/btop/actions) for Linux if there is any issue with any particular platform.

Thanks again for contributing :)

correabuscar:
Really Thank You for this fun, eye-candy program!

<details>

<summary></summary>

On another note, 



Considering that PR #457 fixes an issue where the last IPv4 seen by `btop` on the interface remains shown in NET box even after the IP is removed from the interface (instead of switching to the interface's MAC, or its next IP), and PR #458 fixes the appending of `%ifname` (where `ifname` is eg. `%enp0s3` is appended) after the numeric IP for IPv6 (apparently because `getnameinfo` is used), maybe it's a good idea to consider doing them for freebsd and osx too ... ?

The freebsd one:
https://github.com/aristocratos/btop/blob/c65c5e5d2233854dd9b46296210c15b487c9fbd7/src/freebsd/btop_collect.cpp#L833-L841

The MacOS one:
https://github.com/aristocratos/btop/blob/c65c5e5d2233854dd9b46296210c15b487c9fbd7/src/osx/btop_collect.cpp#L872-L880

I didn't do it for these, because:
1. I didn't even consider they existed as part of `btop` (ie. noticed it late, thank to `grep` for `getnameinfo`, didn't look at directory structure to see freebsd and osx)
2. I don't have a way to test them (well, I could do it in virtualbox, at least the freebsd one, not sure about osx...) 
3. I'm not that motivated since I don't use them.

</details>

aristocratos:
> On another note [...]

You are more than welcome to create the fixes for macos and freebsd if you want to. I can test any PR you create against freebsd and macos targets.

If not, no worries, I'll do it before next release.

correabuscar:
> > On another note [...]
> 
> You are more than welcome to create the fixes for macos and freebsd if you want to. I can test any PR you create against freebsd and macos targets.
> 
> If not, no worries, I'll do it before next release.

I'm interested in doing them, now that I see the new commits (like [this one](https://github.com/aristocratos/btop/commit/f05bc65c3a0b7660976523028c4843aaa224af08)) appear to be doing the builds automatically for each new pull request? that is, before the PR is merged. If this was [already](https://github.com/aristocratos/btop/pull/458#pullrequestreview-1169262302) happening before, I misunderstood and thought that only after the merge(or for any new commits pushed to main) was happening.

aristocratos:
> I'm interested in doing them

Awesome :)

> now that I see the new commits (like [this one](https://github.com/aristocratos/btop/commit/f05bc65c3a0b7660976523028c4843aaa224af08)) appear to be doing the builds automatically for each new pull request?

Yeah, I thought that might be helpful and saves me some time since I don't have to manually test bigger PR's myself.

> btw, you'd want them as 4 PRs, right? (rather than 1 PR made of 4 commits)

It's probably a good idea to split it up so we don't revert unrelated changes if there is an issue that needs it to be reverted.

PR #459, opened 2022-11-06
FreeBSD: use the first IP of the interface

correabuscar:
... instead of the last.

Side effect of this is that it also detects when the current IP gets removed from the interface, instead of keep displaying the old IP.

This is PR #457 but for FreeBSD, not Linux.


correabuscar:
I see it automatically [run](https://github.com/aristocratos/btop/actions) here:
https://github.com/aristocratos/btop/actions/runs/3403403144

aristocratos (approved):

PR #462, opened 2022-11-06
FreeBSD: replace getnameinfo with inet_ntop

correabuscar:
also make IPv6 work (previously it was commented out)

(re)tested it to work in freebsd 13.1 in virtualbox, not just to compile.

Note that, unlike in Linux, if all IPv4 and IPv6  IPs are missing, then the interface MAC isn't shown in their place (but this is a different issue)

### DIFF
--- a/src/freebsd/btop_collect.cpp
+++ b/src/freebsd/btop_collect.cpp
@@ -27,6 +27,7 @@ tab-size = 4
 #include <net/route.h>
 #include <netdb.h>
 #include <netinet/tcp_fsm.h>
+#include <netinet/in.h> // for inet_ntop stuff
 #include <pwd.h>
 #include <sys/_timeval.h>
 #include <sys/endian.h>
@@ -821,7 +822,9 @@ namespace Net {
 				return empty_net;
 			}
 			int family = 0;
-			char ip[NI_MAXHOST];
+			static_assert(INET6_ADDRSTRLEN >= INET_ADDRSTRLEN); // 46 >= 16, compile-time assurance.
+			enum { IPBUFFER_MAXSIZE = INET6_ADDRSTRLEN }; // manually using the known biggest value, guarded by the above static_assert
+			char ip[IPBUFFER_MAXSIZE];
 			interfaces.clear();
 			string ipv4, ipv6;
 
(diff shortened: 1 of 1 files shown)


src/freebsd/btop_collect.cpp
```diff
@@ -843,17 +846,25 @@ namespace Net {
...
+
+							net[iface].ipv4 = ip;
+						} else {
+							int errsv = errno;
+							Logger::error("Net::collect() -> Failed to convert IPv4 to string for iface " + string(iface) + ", errno: " + strerror(errsv));
+						}
+					}
 				}
 				//? Get IPv6 address
-				// else if (family == AF_INET6) {
```
correabuscar, line 851: whoops I forgot an `else`, even tho it works anyway

src/freebsd/btop_collect.cpp
```diff
@@ -843,17 +846,25 @@ namespace Net {
...
-				// }
-
+				if (net[iface].ipv6.empty()) {
+					if (NULL != inet_ntop(family, &(reinterpret_cast<struct sockaddr_in6*>(ifa->ifa_addr)->sin6_addr), ip, IPBUFFER_MAXSIZE)) {
+						net[iface].ipv6 = ip;
+					} else {
+						int errsv = errno;
+						Logger::error("Net::collect() -> Failed to convert IPv6 to string for iface " + string(iface) + ", errno: " + strerror(errsv));
+					}
+				}
```
correabuscar, line 867: and I need to check if AF_PACKET is still a thing to comment here.

correabuscar: not on freebsd

src/freebsd/btop_collect.cpp
correabuscar: force pushed the change.

src/freebsd/btop_collect.cpp
```diff
@@ -843,17 +846,25 @@ namespace Net {
...
+					}
 				}
 				//? Get IPv6 address
-				// else if (family == AF_INET6) {
-				// 	if (net[iface].ipv6.empty() and
-				// 		getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in6), ip, NI_MAXHOST, NULL, 0, NI_NUMERICHOST) == 0)
-				// 		net[iface].ipv6 = ip;
-				// }
-
+				else if (net[iface].ipv6.empty()) {
```
correabuscar, line 860: oh, I made a booboo here, missing AF_INET6

aristocratos (approved):
Looks good! 👍🏼 

Not sure why IPv6 addresses was previously disabled but looks to be working fine.

aristocratos:
> Note that, unlike in Linux, if all IPv4 and IPv6 IPs are missing, then the interface MAC isn't shown in their place (but this is a different issue)

Yeah, there is currently no method employed to get the MAC address in freebsd (or macos I think). In linux it's collected from `/sys/class/net/<iface>/address` if no IPv4 or IPv6 address has been found.

PR #464, opened 2022-11-06
osx: replace getnameinfo with inet_ntop

correabuscar:
this is like PR #462 for FreeBSD,
and like PR #458 for Linux.

--------------

I can't test the build and run for these, as I've been unable to get macOS in virtualbox (tried 4 iso\`s). But I'll look in github actions for the build, and just **hope** the runs work fine on bare metal.

<details>
<summary>outdated info</summary>

Unsure about those `#include`\`s: is `#include <arpa/inet.h>` the same as `#include <sys/arpa/inet.h>` ? 
I've got them from:
```
LEGACY SYNOPSIS
     #include <sys/types.h>
     #include <sys/socket.h>
     #include <sys/netinet/in.h>
     #include <sys/arpa/inet.h>

     These include files are necessary for all functions.
```
src: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/inet_ntop.3.html

</details>

PS: Thank you for this learning experience btw :) (with all the PRs, not just this one)

### DIFF
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -41,6 +41,7 @@ tab-size = 4
 #include <sys/statvfs.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
+#include <netinet/in.h> // for inet_ntop
 #include <unistd.h>
 #include <stdexcept>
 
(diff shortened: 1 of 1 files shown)


src/osx/btop_collect.cpp
```diff
@@ -884,16 +887,24 @@ namespace Net {
...
+							Logger::error("Net::collect() -> Failed to convert IPv4 to string for iface " + string(iface) + ", errno: " + strerror(errsv));
+						}
+					}
 				}
 				//? Get IPv6 address
 				else if (family == AF_INET6) {
-					if (net[iface].ipv6.empty() and
-						getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in6), ip, NI_MAXHOST, NULL, 0, NI_NUMERICHOST) == 0)
+					if (net[iface].ipv6.empty()) {
 						net[iface].ipv6 = ip;
```
correabuscar, line 902: uhm, I need some sleep, missing lines!!! the whole `inet_ntop` is missing here , brb fixing

aristocratos (approved):
Tested to work as expected on macos13, since it compiles fine on macos11 and macos12 I would assume the same is true there.

Great work 👍🏼 
Thanks for your contributions!

PR #478, opened 2022-12-15
macOS: Copy out the select process's command line 

porceCodes:
More often than not, one desires to inspect the exact (shell) command line that was used to invoke a process, especially when that process was invoked in a non-manual fashion.

macOS Activity Monitor does not offer this feature either, to the best of the author's knowledge.

**What this PR achieves**
Upon a keystroke, the full command line of the select process is copied to a clipboard.

**How to use**
* Select a process, and keypress `v`, or
* Show the process detail by `enter`

**Sample outcome**
```
/Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper.app/Contents/MacOS/Code Helper --type=utility --utility-sub-type=audio.mojom.AudioService --lang=en-US --service-sandbox-type=audio --message-loop-type-ui --user-data-dir=/Users/SOMEUSER/Library/Application Support/Code --standard-schemes=vscode-webview,vscode-file --secure-schemes=vscode-webview,vscode-file --bypasscsp-schemes --cors-schemes=vscode-webview,vscode-file --fetch-schemes=vscode-webview,vscode-file --service-worker-schemes=vscode-webview --streaming-schemes --shared-files --field-trial-handle=1718379636,r,5493000567189891038,3303168194434618476,131072 --enable-features=AutoDisableAccessibility --disable-features=CalculateNativeWinOcclusion,SpareRendererForSitePerProcess --seatbelt-client=64
```

**Enhancements**
* Supports the full length of the command line that the underlying platform supports at the time of build.
* Convert `sprintf` to `snprintf` to zap a compiler warning (g++-12)

**Limitation**
* Only macOS is supported. FreeBSD and Linux are TODO.

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -400,6 +400,7 @@ namespace Input {
 						Config::set("proc_last_selected", Config::getI("proc_selected"));
 						Config::set("proc_selected", 0);
 						Config::set("show_detailed", true);
+						process("copyout-shellcmd");
 					}
 					else if (Config::getB("show_detailed")) {
 						if (Config::getI("proc_last_selected") > 0) Config::set("proc_selected", Config::getI("proc_last_selected"));
@@ -436,6 +437,10 @@ namespace Input {
 						return;
 					else if (old_selected != new_selected and (old_selected == 0 or new_selected == 0))
 						redraw = true;
+				} else if (is_in(key, "copyout-shellcmd", "v")) {
+					auto procs = Proc::collect(false);
+					auto p_info = rng::find(procs, Config::getI("selected_pid"), &Proc::proc_info::pid);
+					copy_to_clipboard(p_info->cmd);
 				}
 				else keep_going = true;
 
--- a/src/btop_tools.cpp
+++ b/src/btop_tools.cpp
@@ -597,3 +597,13 @@ namespace Logger {
 		}
 	}
 }
+
+int copy_to_clipboard(const std::string& msg) {
+#ifdef __APPLE__
+	char buf[ARG_MAX];
+	snprintf(buf, ARG_MAX, "echo \"%s\" | pbcopy", msg.c_str());
+	return system(buf);
+#elif // __APPLE__
+	return 0;
+#endif // __APPLE__
+}
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -1224,7 +1224,7 @@ namespace Proc {
 								std::string_view proc_args(proc_chars.get(), argmax);
 								if (size_t null_pos = proc_args.find('\0', sizeof(argc)); null_pos != string::npos) {
 									if (size_t start_pos = proc_args.find_first_not_of('\0', null_pos); start_pos != string::npos) {
-										while (argc-- > 0 and null_pos != string::npos and cmp_less(new_proc.cmd.size(), 1000)) {
+										while (argc-- > 0 and null_pos != string::npos and cmp_less(new_proc.cmd.size(), ARG_MAX)) {
 											null_pos = proc_args.find('\0', start_pos);
 											new_proc.cmd += (string)proc_args.substr(start_pos, null_pos - start_pos) + ' ';
 											start_pos = null_pos + 1;
@@ -1235,8 +1235,8 @@ namespace Proc {
 							}
 						}
 						if (new_proc.cmd.empty()) new_proc.cmd = f_name;
-						if (new_proc.cmd.size() > 1000) {
-							new_proc.cmd.resize(1000);
+						if (new_proc.cmd.size() > ARG_MAX) {
+							new_proc.cmd.resize(ARG_MAX);
 							new_proc.cmd.shrink_to_fit();
 						}
 						new_proc.ppid = kproc.kp_eproc.e_ppid;
--- a/src/btop_tools.hpp
+++ b/src/btop_tools.hpp
@@ -353,3 +353,4 @@ namespace Logger {
 	inline void debug(const string msg) { log_write(4, msg); }
 }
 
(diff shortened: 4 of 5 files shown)


src/btop_input.cpp
```diff
@@ -436,6 +437,10 @@ namespace Input {
 						return;
 					else if (old_selected != new_selected and (old_selected == 0 or new_selected == 0))
 						redraw = true;
+				} else if (is_in(key, "copyout-shellcmd", "v")) {
+					auto procs = Proc::collect(false);
```
aristocratos, line 441: `atomic_wait(Runner::active);` should be called first to make sure we aren't pulling information from a list that's being updated.
And `Proc::collect()` Should be called with `true` here since we don't need to update the processes.

src/btop_input.cpp
```diff
@@ -400,6 +400,7 @@ namespace Input {
 						Config::set("proc_last_selected", Config::getI("proc_selected"));
 						Config::set("proc_selected", 0);
 						Config::set("show_detailed", true);
+						process("copyout-shellcmd");
```
aristocratos, line 403: I'm not sure if it's good practice to replace the users clipboard without warning anytime they open the information panel.

src/osx/btop_collect.cpp
```diff
@@ -1235,8 +1235,8 @@ namespace Proc {
 							}
 						}
 						if (new_proc.cmd.empty()) new_proc.cmd = f_name;
-						if (new_proc.cmd.size() > 1000) {
-							new_proc.cmd.resize(1000);
+						if (new_proc.cmd.size() > ARG_MAX) {
+							new_proc.cmd.resize(ARG_MAX);
```
aristocratos, line 1239: `argmax` retrieved in `Shared::init()` by `sysconf()` should be used instead of `ARG_MAX`.
Line 1220 `size_t argmax = Shared::arg_max;`

aristocratos (changes requested):
I like the initiative but there are some big security issues using `system()` to copy to clipboard.

A pure programmatic way to achieve is needed.
[NSPasteboard](https://developer.apple.com/documentation/appkit/nspasteboard?language=objc) looks like preferred way on MacOS, but might be tricky to get working with C++.

Another way could be to look at how QT does this in it's `QClipboard` classes, that way you could also gain insight in how to handle this on Linux and FreeBSD.

src/btop_tools.cpp
```diff
@@ -597,3 +597,13 @@ namespace Logger {
 		}
 	}
 }
+
+int copy_to_clipboard(const std::string& msg) {
+#ifdef __APPLE__
+	char buf[ARG_MAX];
+	snprintf(buf, ARG_MAX, "echo \"%s\" | pbcopy", msg.c_str());
+	return system(buf);
```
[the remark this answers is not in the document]
porceCodes: Point taken. Improved a bit, with an assumption that text file redirect is either mitigating or eliminating the attack front. PTAL.

src/btop_input.cpp
porceCodes: `atomic_wait` is added. But I request to evaluate if the new patch is appropriate; esp. the new patch uses atomic_wait not only for `get_cmd_line(pid)` but also for `copy_to_clipboard()`

Now `no_update` as `true` is used.

src/btop_input.cpp
porceCodes: Remove this behavior upon `enter` keystroke. Now `v` is the only keystroke. The help message clarifies and affirms the user's intention and corresponding behavior

src/osx/btop_collect.cpp
porceCodes: Applied wherever applicable. PTAL a new pragma directive; that is introduced due to a fact that `arg_max` is a variable not a constant.

porceCodes:
All points from the review are taken. The new patch does not yet adopt Swift library or QT library APIs. I think that would warrant a discussion in advance on how much new dependencies the maintainer desires, and when.

aristocratos (changes requested):
I would like to avoid external dependencies (not in system by default) if possible.

Regarding QT I was more thinking along the lines of looking at their clipboard implementation as inspiration for writing it natively, not pulling in QT libs, which would be a bit excessive for a single feature.

For example, QT clipboard implementation for MacOS:
https://github.com/qt/qtbase/blob/067b53864112c084587fa9a507eb4bde3d50a6e1/src/plugins/platforms/cocoa/qmacclipboard.h
https://github.com/qt/qtbase/blob/067b53864112c084587fa9a507eb4bde3d50a6e1/src/plugins/platforms/cocoa/qmacclipboard.mm

src/btop_tools.cpp
```diff
@@ -597,3 +597,36 @@ namespace Logger {
...
+	fp << msg;
+	fp.close();
+
+	// TODO(..): Explore an API-driven option not to invoke a shell command
+	// (1) Use Swift/Objective C++ binding and call NSPasteboard. Enhance Makefile to support.
+	//     See https://developer.apple.com/documentation/appkit/nspasteboard?language=objc
+	// (2) Explore to use QClipboard but be mindful on dependency
+	//     See https://doc.qt.io/qt-6/qclipboard.html
+	const auto argmax = Shared::arg_max;
+	#pragma GCC diagnostic push // Alternative to this is to use ARG_MAX directly from limits.h
```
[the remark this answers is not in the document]
joske: Sorry, have no idea anymore about ARG_MAX :shrug:

porceCodes:
> I would like to avoid external dependencies (not in system by default) if possible.

For macOS, this is how a text can be copied to a clipboard. It requires one `.mm` file and one line of API declaration. It will be great if someone can adjust `Makefile` to include this.

https://github.com/porceCodes/cl

Issue #479, opened 2022-12-15
[BUG] Possibly mssing target dependencies in Makefile


NexAdn:
**Describe the bug**

In [Gentoo Bug 884005](https://bugs.gentoo.org/884005) we noticed builds failing due to files not being created with e.g. this output from the Tinderbox:

> Assembler messages:
> Fatal error: can't create obj/btop_menu.o: No such file or directory
> Assembler messages:
> Fatal error: can't create obj/btop_draw.o: No such file or directory

As described by a user on the [gentoo-user mailing list](https://marc.info/?l=gentoo-user&m=166980746719287), the problem might be that the `mkdir` commands for the output directories didn't finish before some jobs attempted to open files in said directories, leading to write failures.

**To Reproduce**

The aforementioned user got this problem by building on a machine with many jobs in parallel (here it was `-j17`).

**Expected behavior**

The compiler should be able to open the files for writing.

**Screenshots**

Screenshots not applicable, but you might be interested in some of this output:

* [Tinderbox build log](https://884005.bugs.gentoo.org/attachment.cgi?id=838935)
* [First comment on Bug 884005](https://bugs.gentoo.org/884005#c0) (contains a condensed output with less noise)
* [Environment variables set during build](https://884005.bugs.gentoo.org/attachment.cgi?id=838931)
* [emerge --info](https://884005.bugs.gentoo.org/attachment.cgi?id=838927) (contains a bunch of system and package manager information from the build)

**Info (please complete the following information):**
 - btop++ version: `btop -v` – Not applicable, since the build failed. As can be deduced from the build log, a build of 1.2.13 was attempted.
 - Binary: self compiled
 - (If compiled) Compiler and version: From the build log: `CXX        ?| x86_64-pc-linux-gnu-g++ (12.2.1)`
 - Architecture: [x86_64, aarch64, etc.] `uname -m`: amd64
 - Platform: Gentoo Linux
 - (Linux) Kernel: `uname -r`: Linux-6.0.10-x86_64-AMD_Ryzen_9_5950X_16-Core_Processor-with-glibc2.36
 - (OSX/FreeBSD) Os release version: Not applicable
 - Terminal used: Not applicable
 - Font used: Not applicable

**Additional context**

contents of `~/.config/btop/btop.log` – Not applicable

**GDB Backtrace**

Not applicable.

NexAdn:
[...]

PR #481, opened 2022-12-15
Makefile: make build targets depend on directories

NexAdn:
Since directory creation is a requirement for writing the built artifacts, this commit introduces a dependency on the directories for all build targets (compile and link), so parallel builds can't fail because a build target is executed before the directories target.

Closes: https://github.com/aristocratos/btop/issues/479

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,7 +259,7 @@ uninstall:
 
 #? Link
 .ONESHELL:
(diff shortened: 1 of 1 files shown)


aristocratos (approved):
Thanks for the contribution and for fixing the issue before I've even had time to see #479 :)

PR #485, opened 2022-12-24
Create adwaita.theme

flipflop133:
Add the popular Adwaita theme for people who like a theme matching GTK applications.
Note: I used a 3 colors gradient while the start and mid-colors are the same, this is not a mistake, it's because using a two colors gradient it would create ugly mid-colors.


flipflop133:
Link to the color palette used:
https://developer.gnome.org/hig/reference/palette.html

aristocratos (approved):

Issue #346, opened 2022-05-30
[REQUEST] Include a method to copy process parameters


stevenxxiu:
**Is your feature request related to a problem? Please describe.**
I always wanted a way to copy process parameters in system monitors, to make a note of them before I close them. Other monitors e.g. *Glances*, also lacked this feature.

**Describe the solution you'd like**
Give the user the option to specify a shortcut key in config, say `Alt + C` or something, to copy this.

**Describe alternatives you've considered**
It's rather annoying to have to run `pgrep` in another terminal to be able to copy this.

Issue #490, opened 2023-01-03
[REQUEST] Add LLVM support


stwnt:
Due to missing implementations in llvm ```btop``` didn't support llvm based compilation for some time.
It's now more or less possible to compile ```btop```, with some tricks, with llvm and clang. I myself got a few llvm/musl system and I'd like to use ```btop``` on them aswell without a dependecy on gcc.

Ad support for llvm base compilation in the makefile.

When llvm's ```<ranges>``` implementation is complete we should rollback some workarounds.

Notes:
We need to use libc++ because clang cannot compile ```libstdc++'s``` ```<ranges>``` because it relies on concept which are not yet ready in clang.
We need to enable libc++'s experimental features for ```<ranges>``` to work properly.
We need to edit some flags passed to the compiler which are not known by clang.

I'll open a PR.

stwnt:
I think you saw my previous PR which was a bit hasty. I've read your opinion on adding another build system to the mix. I'm pretty sure I've read in those comments that you want to wait for better ```<ranges>``` support by llvm and don't want to implement a ton of hacks to support llvm prematurely.

This is just for llvm support and I don't care a lot what build system is used, however as you saw in my other PR I used cmake to do prepare this. The build file is [here](https://gist.github.com/stwnt/b1387180e96b9ea27b65468c3a50784c) for @jan-guenter and maybe you to have a look at. It doesn't check if a flag is supported but it adds flags based on the compiler in use with generator expressions. Neat feature. ;)

aristocratos:
Yeah sure, go for it. 👍🏼 

> When llvm's <ranges> implementation is complete we should rollback some workarounds.

I'm not sure I necessarily agree with this though, as far as I know only `ranges::view` is an issue right now, and it's very sparsely used in the code (isn't it just the `Tools::ssplit()` function?), so better to just replace those functions. As long as the replacement code is well written there shouldn't be any performance difference.

stwnt:
You're right, specifically it's just the ```rng::split``` function. The [Libc++ Ranges Status](https://libcxx.llvm.org/Status/Ranges.html#id3) page has info about the missing parts. I'll have a look at it over the next few months and then we can decide to roll back the changes or leave it as it is.

stwnt:
I've encountered two problems today on another machine:
- Currently I can't compile with LLVM's ```compiler-rt``` and ```llvm-libunwind```. The program segfaults on exit. (I've seen similar things with ```neovim``` and ```llvm-libunwind```.
[...]

PR #496, opened 2023-01-29
feat: copy selected process command to clipboard

stevenxxiu:
This is for [[REQUEST] Include a method to copy process parameters · Issue #346 · aristocratos/btop](https://github.com/aristocratos/btop/issues/346).

I looked at [macOS: Copy out the select process's command line by porceCodes · Pull Request #478 · aristocratos/btop](https://github.com/aristocratos/btop/pull/478) and took your comments into account on using other variables. I added `selected_cmd`.

Only tested on *X11*.


stevenxxiu:
Hmm I just realized that [dacap/clip: Cross-platform C++ library to copy/paste clipboard content](https://github.com/dacap/clip) is *X11* only. Is there some library perhaps that supports *Wayland* as well?

aristocratos:
@stevenxxiu @porceCodes
Before we get too deep down the rabbit hole of supporting the clipboards of every major windowing system :wink:
There might be a much simpler solution; 

A button called `full cmd` that shows the full command line for the selected process in a new box, pauses btop mouse control and pauses the background "drawing" so the text can be selected and copied manually. If it won't fit on one screen we can either split it in to multiple pages or switch back to "normal" terminal mode which can be scrolled and print it there and wait for "return to continue" or something like that.
This would also allow it to work in a TTY framebuffer with `tmux`, `screen`, etc.

And would avoid the mess of determining the current windowing system and adding external libraries on Linux.
The current code in #478 is also problematic because the Objective-C code used can only be compiled with clang, and even with clang 15 it's currently problematic to get btop to compile due to C++20 not fully being implemented, see #490.

What do you guys think of this solution? It would certainly simplify things a bit.

stevenxxiu:
That would be good if copying can still be done with one key. But I'm not sure if terminal emulators support this. I just tried selecting some region in the *Micro* text editor and using *WezTerm*'s copying key, but that didn't copy anything. Turns out *Micro* uses some clipboard library itself.

One goal for mine here is to do it quickly and more easily than the `pgrep` mess.

If I have to manually drag my mouse to copy the text, that just becomes a pain.

aristocratos:
@stevenxxiu 
Not one key, because the terminal needs to do the "selecting", but using `WezTerm` and it's default keybinds as an example:
1. `x` or whatever key used in `btop` to show the full command line.
2. `CTRL + SHIFT + X` to enter copy mode. The cursor should be left on the first character in the text by btop.
3. Select the region and `CTRL + SHIFT + c` to copy to clipboard.

My question here would be how many people have a need to regularly copy the command line of a running process, would just having the ability to see the full cmd be enough for most people and if need be manually select and copy it.

Is there enough people that would use this that it warrants a single key click and the extra includes and more involved compile process that comes with it?

If that is the case, another solution could be to add a config setting for `copy command` which is executed with the processes command line as argument (with dropped priviliges). It would be left empty by default but anyone that needs the functionality can manually add for example `xclip` as a `copy command`.

stevenxxiu:
@aristocratos

Well I think your idea of showing the full command line would be quite useful. Just to get a peek, whether it'll be copied not. I'm not sure how you would implement it, nor am I too fluent in *C++* so it would be great if you made that happen :+1:.

I think there are enough people who want to copy the command line, as there are already a few issues about it, and 2 PRs, including mine.

Using a `copy command` config setting seems quite ideal, as that'll support *Wayland* too. I'll give that a shot and code that. How do you envision dropping the privileges?

stevenxxiu:
I added a `copy_command` setting, and tested that it works, by setting it to:

```ini
copy_command = "/usr/bin/xclip -selection clipboard"
```

porceCodes:
> My question here would be how many people have a need to regularly copy the command line of a running process, would just having the ability to see the full cmd be enough for most people and if need be manually select and copy it.

I think being able to see the full command, however long it is, is more useful than being copying-out that text. By doing so, we can separate out pending concerns from a primary usefulness. Please note however, the maximum length for a command line characters can be large enough not to fit within a single screen, and may require scrolling when visual selection is required.

stevenxxiu:
If the command is so long you have to scroll to see it, it'll probably be hard to tell what's in it. This is another case where copying it becomes very useful, to look at it in a text editor.

stevenxxiu:
I changed the default key to `y`, as in "yank" in *Neovim*. This feels a bit more natural vs `x`, like in `Ctrl + x`.

stevenxxiu:
How does it open up for possible exploits? If you mean setting a binary to run, that's understandable, but it's set by the user. This isn't so different to configuring the shell, so it feels alright to me.

aristocratos:
Below is just one simple example:
User has installed btop manually and during install also ran `make setuid`, granting root priviligies to btop.

An attacker gains access to this user account, notices btop is available, changes the clipboard command to `bash #`, runs btop and presses the clipboard command, the attacker now has a root shell.

stevenxxiu:
If the attacker is able to change the clipboard command that's already too late. Chances are they can change any file in the home directory. They'll be able to change shell config, and also gain root access.

The alternative here is to support clipboards for every system. This seems like quite a bit of overhead. It'll also probably miss some system. I think the added safety is extremely marginal.

PR #510, opened 2023-03-22
Support compiling with LLVM

stwnt:
This is more of a draft, I appreciate feedback.

I tried my best to test all possbile configurations with my available hardware, which are x86_64/LLVM/GCC,Musl/Glibc/libstc++/libcxx/static. I'm pretty sure there will be hicups with Apple and BSD.


nobounce:
I don't know why the CI for FreeBSD is failing.

There is currently still an issue when using `llvm-libunwind` as the unwindlib when exiting the app and canceling all launched threads.

aristocratos:
@nobounce 
Seems to be a Makefile issue 
`Makefile:51: *** ERROR: c++: Version minimum is 16.0.0.  Stop.`

It stops because of wrong `clang` version detected ignoring that a compatible `gcc` version is available.

A solution would be to instead increment a error counter variable and if the version checks for both `clang` and `gcc` fails, then throw an error.

aristocratos:
@nobounce
I'm getting 8000 lines of `undefined reference` linker errors trying to compile with clang 16 on Ubuntu 22.04 with everything directly pulled from https://apt.llvm.org/

Installed packages:
```clang-16 clang-tools-16 clang-16-doc libclang-common-16-dev libclang-16-dev libclang1-16 clang-format-16 python3-clang-16 clangd-16 clang-tidy-16 libc++-16-dev libc++abi-16-dev libllvm-16-ocaml-dev libllvm16 llvm-16 llvm-16-dev llvm-16-doc llvm-16-examples llvm-16-runtime```

Am I missing something?

aristocratos:
@nobounce 
Turns out the problem was that I was passing `CXX=clang-16` instead of `CXX=clang++-16` :)
`lld` and `libunwind` got pulled in as dependencies by the earlier packages.

Not sure why, but the binary compiled by clang is 2.0MiB compared to 1.5MiB by gcc, with the same optimization flags?

Will take a closer look at the changes this week, but since it's passing compile for all platforms now it's looking good. Nice work! 👍🏼

nobounce:
For what I know, Clang does more aggressive loop unrolling with `-O2` than GCC with `-O3`. I'll have a look aswell.

nobounce:
@aristocratos

| Compiler | Bytes | Opt | Libc | Static |
|---------------|----------|------|--------|----------|
| Clang | 1,690,008 | -O2 | Glibc | false |
| Clang | 1,826,016 | -O3 | Glibc | false |
| GCC | 1,411,432 | -O2 | Glibc | false |
| GCC | 1,493,384 | -O3 | Glibc | false |
| Clang | 1,691,552 | -O2 | musl | false |
| Clang | 1,826,152 | -O3 | musl | false |
| Clang | 2,515,832 | -O2 | musl | true |
| Clang | 2,651,528 | -O3 | musl | true |

That seems reasonable to me.
My only concern is that Clang takes 1.5x the time for compilation (08s / 12s).

nobounce:
@aristocratos 
Did you have the time to look at this?
It would be great if that can land in main. Maybe coupled with a new release so (source-based-)distros can take advantage of this and remove gcc as a dependency :)

aristocratos (approved):
> Maybe coupled with a new release so (source-based-)distros can take advantage of this and remove gcc as a dependency :)

I can't imagine that many distros have clang16 included by default yet?

Gpu support should be done in a little while, and are planning to do a release with a minor version bump then.

vimproved:
FYI, Gentoo has clang 16, but only on the testing branch. LLVM 16 is not yet stabilized because of those various issues.

PR #525, opened 2023-05-11
Fix link to the themes folder

gwena:
Documentation: Fix the link to the `themes` folder (broken since the move from `base: master` to `base: main`)

Thanks a lot for btop!


aristocratos (approved):
Thanks for fix :)

PR #529, opened 2023-05-18
Add GPU monitoring support

romner-set:
Implements #426. Not ready for merge yet, still WIP.

Currently implemented features:
- GPU/VRAM/power usage & temp graphs in CPU panel
2023-05-18-161701_hyprshot
- Detailed GPU panels with 5,6,7,8,9,0 keys, one for each GPU
2023-05-19-164603_hyprshot

- Support for NVIDIA and AMD GPUs

Planned features:
- GPU process tracking + integration with proc panel?
- Intel GPU support
- macOS/FreeBSD support
- ???


aristocratos:
@romner-set 
I did some benchmarking since the collection times for nvidia was a bit high, below are result from one of the peaks.
As you can se the totals are 246 ms, which seems to be happening mostly when the gpu is in low power mode.

However, 98.5% of the time seems to be spent on the two calls to `nvmlDeviceGetPcieThroughput()`
When I disabled those calls the total time for gpu collection went down to between 1 - 3 ms.

The PCIE link speeds should probably have a toggle which is disabled by default, and that has a warning since those calls can cause significant lag.

``` logfile
2023/05/21 (03:09:47) | DEBUG: Timer [Nv utilization] took 260 μs
2023/05/21 (03:09:47) | DEBUG: Timer [Nv clock] took 97 μs
2023/05/21 (03:09:47) | DEBUG: Timer [Nv power] took 2'563 μs
2023/05/21 (03:09:47) | DEBUG: Timer [Nv temp] took 56 μs
2023/05/21 (03:09:47) | DEBUG: Timer [Nv mem] took 50 μs
2023/05/21 (03:09:48) | DEBUG: Timer [Nv pcie] took 242'536 μs
2023/05/21 (03:09:48) | DEBUG: Timer [Nvidia Total] took 246'184 μs
```

romner-set:
@aristocratos
Alright, I'll keep that in mind. Were you able to benchmark the ROCm SMI code too, or should I do that myself?

Also, is it really necessary to have all of the `fmt` header files in the project? I'm pretty sure nearly all distros package them in `/usr/include`.

aristocratos:
@romner-set 
Haven't done any benchmarks on the ROCm code yet, but I can push a commit with the `DebugTimer` benchmark class and leave the benchmarks for the nvidia code commented out for usage examples.

> Also, is it really necessary to have all of the `fmt` header files in the project? I'm pretty sure nearly all distros package them in `/usr/include`.

Why would they be included by default? Not something I've seen in any distro I've used so far.
But the goal is to limit the need to install dependencies and to increase compatibility by having the right version already included, it's not like they take up much space.

romner-set:
@aristocratos 
> Why would they be included by default? Not something I've seen in any distro I've used so far.
But the goal is to limit the need to install dependencies and to increase compatibility by having the right version already included, it's not like they take up much space.

I meant that there is a `fmt` package nearly everywhere that includes them, sorry. And if that's the case, should I also add `nvml.h` and the `rocm_smi` header files to the include dir? They're ~15k lines in total, which is around the same size as `fmt`.

aristocratos:
Huh, it's only 3 files: `nvml.h`, `rocm_smi.h` and `kfd_ioctl.h`?

As long as the licensing information from them is also included in the readme, I don't see any issue with that.
Do you know how often they are updated?

It would make it a lot smoother to compile with gpu support. I think there was around 1 GiB of dependencies automatically pulled down when installing AMD's gpu tools via their repository for ubuntu.

Then we could easily just add a flags for GPU_NVIDIA, GPU_AMD or GPU_ALL in the Makefile.
And would make the project free from external dependencies even if compiling with gpu support.

aristocratos:
Forgot about the libraries for them also, so including the headers might not really matter then, since you still need to install the libraries for the correct platform...

aristocratos:
We should also look into either static compilation with the Nvidia and Amd libraries (if even possible?) or preferably dynamic loading with `dlsym()` / `dlopen()` (see https://github.com/vertcoin-project/vertminer-nvidia/blob/master/nvml.cpp for a example of wrapping NVML with dlopen).

The alternative would be to have 4 binary releases (btop, btop-gpu-nvidia, btop-gpu-amd, btop-gpu-all) to avoid crashing on missing dynamic gpu libraries.

I don't have any previous experience with `dlsym()` and `dlopen()`, so have no clue how easy it would be to setup, but what are your thoughts about it?

romner-set:
> We should also look into either static compilation with the Nvidia and Amd libraries (if even possible?) or preferably dynamic loading with dlsym() / dlopen()

I've looked into static linking before, and I'm pretty sure it isn't possible in this case. I'm not sure about `librocm_smi64`, but (at least on my system) `libnvidia-ml.so` is symlinked to a driver version-specific `libnvidia-ml.so.530.41.03`, so it presumably couldn't work even if there was a .a file.

I don't have any experience with `dlsym()` either, but it doesn't seem to be too difficult to implement considering we don't have to deal with windows compatibility. And having multiple binary versions should really only be a last resort, especially when you also consider support for Intel GPUs (`btop`, `btop-nvidia`, `btop-amd`, `btop-intel`, `btop-amd-intel`, `btop-amd-nvidia`...)

romner-set:
@aristocratos 
NVML seems to be stable and well-supported enough on all distros that we can load it dynamically, but it's probably safer to compile and link ROCm SMI statically for the time being:

> The rocm_smi library api is new, and therefore subject to change either at the ABI or API level.
Instead of marking every function prototype as "unstable", we are instead saying the API is
unstable (i.e., changes are possible) while the major version remains 0. This means that if the
API/ABI changes, we will not increment the major version to 1. Once the ABI stabilizes, we will
increment the major version to 1, and thereafter increment it on all ABI breaks.

And since it's completely separate to the AMD driver, it's possible that not all distros have a package for it yet. I haven't checked too many, but Void doesn't seem to have one, for example. What are your thoughts on it?

aristocratos:
@romner-set 
Nice work on dynamic loading of NVML, seems to be working nicely :)

Static compilation of ROCm is fine in my opinion. But it should probably be set to not be a "opt in" option, or adding some logic for testing supported platforms for the drivers in the Makefile.

GandelXIV:
gaming

romner-set:
@aristocratos 
I made ROCm load dynamically by default and only statically compile with an `RSMI_STATIC` make flag, should work fine.

I've also checked how long the `nvmlDeviceGetPcieThroughput()` calls take on my GPU, and it was nowhere near as drastic:
```
2023/06/06 (19:04:38) | DEBUG: Timer [Nv utilization] took 1'889 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nv clock] took 4'653 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nv power] took 3'212 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nv temp] took 342 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nv mem] took 17 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nv pcie] took 41'280 μs
2023/06/06 (19:04:38) | DEBUG: Timer [Nvidia Total] took 51'805 μs
```
Still, by putting the calls on separate threads you can cut the time in half. The main thread can also do everything else in the meantime, so it ends up like this:
```
2023/06/06 (19:42:08) | DEBUG: Timer [Nv utilization] took 1'721 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nv clock] took 4'878 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nv power] took 2'903 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nv temp] took 298 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nv mem] took 14 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nv pcie thread join] took 15'143 μs
2023/06/06 (19:42:08) | DEBUG: Timer [Nvidia Total] took 25'128 μs
```
Still not ideal, but I'd say it's good enough. Could you benchmark this version on your system to see if it's any better?

aristocratos:
@romner-set 
> I made ROCm load dynamically by default and only statically compile with an RSMI_STATIC make flag, should work fine.

Good idea! Should probably remove the rocm submodule also and instead add `lib/rocm_smi_lib` to the `.gitignore` and we can add instructions in the readme for how to compile rocm statically.
Otherwise a lot of people will be pulling it down when most likely not needing it if pulling recursively to get the `libfmt` submodule.

> Could you benchmark this version on your system to see if it's any better?

The average for the two `nvmlDeviceGetPcieThroughput` calls looks a lot better now, around 20ms, but I still get some rare spikes up to over 200ms. Don't know if it could be related to that the card I'm testing with is an older GTX 970. But I'm a bit worried about what those spikes might look like with even older cards (or the lower numbered cards from the same generation).

romner-set:
> Good idea! Should probably remove the rocm submodule also and instead add `lib/rocm_smi_lib` to the `.gitignore` and we can then add instructions in the readme for how to compile rocm statically. Otherwise a lot of people will be pulling it down when most likely not needing it if pulling recursively to get the `libfmt` submodule.

Sounds good, I'd also add some sort of warning in the makefile if `RSMI_STATIC` is defined but the `lib/rocm_smi_lib` directory doesn't exist.

> The average for the two `nvmlDeviceGetPcieThroughput` calls looks a lot better now, around 20ms, but I still get some rare spikes up to over 200ms. Don't know if it could be related to that the card I'm testing with is an older GTX 970. But I'm a bit worried about what those spikes might look like with even older cards (or the lower numbered cards from the same generation).

The GTX 970 is Maxwell, i.e. the oldest generation supported by `nvmlDeviceGetPcieThroughput()`:
```
Retrieve PCIe utilization information. This function is querying a byte counter over a 20ms interval and thus is the PCIe throughput over that interval.

For Maxwell or newer fully supported devices.

This method is not supported in virtual machines running virtual GPU (vGPU).
```
I think it really depends on how common these spikes are, and if they're even noticeable as the lowest update period of 100ms. Either way, adding a config option to turn them off shouldn't be difficult, so we might as well.

nobounce:
It would be a bit cleaner if you would revert the last commit and rebase on main instead

romner-set:
> It would be a bit cleaner if you would revert the last commit and rebase on main instead

You're right, I thought that's what github does when it syncs forks, apparently not. Will fix now.

romner-set:
@aristocratos I've been trying to find a simple-ish way to get Intel GPU info, but there doesn't seem to be any. I managed to hook into [IGT](https://gitlab.freedesktop.org/drm/igt-gpu-tools) with a standalone program to display usage %, but that requires a ton of dependencies and only works for 1 GPU.

I think the best course of action would be to finish the PR without Intel support for the time being and implement it through generic DRM calls later on. What do you think? The only thing left to add after this is macOS/FreeBSD support, which I can't really do since I have neither of these OSs for testing.

aristocratos:
@romner-set 
> I think the best course of action would be to finish the PR without Intel support for the time being and implement it through generic DRM calls later on. What do you think?

That sounds good to me, hopefully more ways to monitor Intel gpu's becomes available now that they have started making discrete graphic cards.

There's still some UI issues to fix, for example when the gpu is shown in the cpu box and there is only 1 column:
image

Also should add some more obvious way to turn on the gpu monitoring (right now you have to select a gpu graph to enable it or trigger the 5-9 keys).
Maybe a "show_gpu" option in the CPU section that automatically sets the lower graph and shows the stats.
And then automatically change back the graphs in the cpu box when showing a dedicated gpu box with 5-9 keys.

I also noticed that musl doesn't support dynamic loading of libraries (maybe should have been obvious), so gonna do some experimenting with how to have gpu support included in the release binaries.
Maybe a separate release binary statically linked with glibc for x86 with gpu support or something similar.

Hopefully gonna have some time for this in the coming 2-3 weeks.

But hey, awesome work so far :)
GPU support wouldn't have been anywhere near this state of completion without your efforts!

romner-set:
> There's still some UI issues to fix, for example when the gpu is shown in the cpu box and there is only 1 column

Fixed by removing the temp graph when the box isn't wide enough... it would've been better to enlarge the box slightly instead, but the core column widths seem to be hardcoded in some weird way so this is a way simpler solution.

> Also should add some more obvious way to turn on the gpu monitoring (right now you have to select a gpu graph to enable it or trigger the 5-9 keys).
Maybe a "show_gpu" option in the CPU section that automatically sets the lower graph and shows the stats.
And then automatically change back the graphs in the cpu box when showing a dedicated gpu box with 5-9 keys.

If there's a GPU detected then the `gpu-totals` graph is selected automatically, it isn't only when there's something else set in the config already (e.g. when updating from a previous version).

I think it'd be fine to keep it as is, but make a default-on `show_gpu` option control the per-GPU brief info shown in the CPU box independently of GPU-related graphs. This way people with old configs from an older version not following changelogs will still be able to see atleast that.

I'm not sure about automatically switching GPU graphs to CPU when all detailed panels are open, but it could be implemented pretty easily (probably as another option?)

> I also noticed that musl doesn't support dynamic loading of libraries (maybe should have been obvious), so gonna do some experimenting with how to have gpu support included in the release binaries.
Maybe a separate release binary statically linked with glibc for x86 with gpu support or something similar.

Does it really? My musl-based Void installation has a bunch of .so files in /lib and /usr/lib so I'd be very surprised if they weren't being used. Either way, I'm pretty sure that official NVIDIA drivers don't work with musl anyway and we can just link RSMI statically, so I don't think that will be a problem.

> But hey, awesome work so far :)
GPU support wouldn't have been anywhere near this state of completion without your efforts!

Thanks! I really wanted btop to have GPU support and then thought "Wait, I know C++ and a bit about GPUs. Why not just implement it myself?" so here we are. The beauty of open-source!

aristocratos:
> If there's a GPU detected then the gpu-totals graph is selected automatically, it isn't only when there's something else set in the config already (e.g. when updating from a previous version).

Ah ok, missed that.

> I think it'd be fine to keep it as is, but make a default-on show_gpu option control the per-GPU brief info shown in the CPU box independently of GPU-related graphs. This way people with old configs from an older version not following changelogs will still be able to see atleast that.
> 

That works.

> Does it really? My musl-based Void installation has a bunch of .so files in /usr/lib so I'd be very surprised if they weren't being used. Either way, I'm pretty sure that official NVIDIA drivers don't work with musl anyway and we can just link RSMI statically, so I don't think that will be a problem.

I assume that those libraries are also compiled with musl.
Static linking with libc has a lot of issues, so maybe having a extra "gpu binary" for x86_64 dynamically linked (with a minimum glibc requirement) could be a solution.
This issue should solve itself eventually when the btop versions with gpu support eventually gets picked up by the distributions.

Gonna take a look at how much work it would be to get gpu working for macos and freebsd. I'm assuming the freebsd drivers should have similar if not the same API, so should be able to adapt the linux code for how freebsd handle dynamic loading. And for macos there's only AMD to look at really since there isn't even OS support for Nvidia anymore.

BlueskyFR:
Hey guys. So hyped for this!

Can I help with this? Just tried it on x84-64 with 8 NVIDIA GPUs and cannot get anything to display.

romner-set:
> Hey guys. So hyped for this!
> 
> Can I help with this? Just tried it on x84_64 with 8 NVIDIA GPUs and cannot get anything to display.

If you set btop's loglevel to DEBUG, are there any GPU-related warnings or errors in `~/.config/btop/btop.log`?

Also I'd suggest making an [issue](https://github.com/romner-set/btop-gpu/issues) on my fork to not clutter this PR.

aristocratos:
@romner-set 
Haven't really had time to look over FreeBSD and MacOS support yet, but I'm thinking this is ready to merge as GPU support for Linux if you don't have any objections?

Will be a little while before publishing a new release so should be some time to detect any issues from people compiling from main.

romner-set:
> @romner-set Haven't really had time to look over FreeBSD and MacOS support yet, but I'm thinking this is ready to merge as GPU support for Linux if you don't have any objections?
> 
> Will be a little while before publishing a new release so should be some time to detect any issues from people compiling from main.

I've been quite busy the past several weeks so I didn't have much time to work on it, but before that I added a mostly-working `enable_gpu` option to dynamically shutdown/initialize NVML and ROCm and make GPU support generally less confusing. I'll finish it hopefully sometime this week and put it on a separate branch on my fork, since I'm pretty sure an auto-merge won't be possible with the commits you made.

Other than that I think it is ready to merge if you also think so.

aristocratos:
@romner-set 
Sounds good, let me know when you created the branch and I'll take a look if I can merge together any conflicts from the code I changed/added.

The macro code isolation I added for freebsd and macos is temporary btw, gonna fix it up properly before the merge.

romner-set:
@aristocratos https://github.com/romner-set/btop-gpu/tree/enable-gpu
Sorry that it took so long, I got ill and kind of forgot about it for a while. The merge conflicts don't seem to be too terrible, but I'm not sure how to handle my enable_gpu vs your show_gpu_info, so I'll leave that up to you. Feel free to ask if you need anything else.

aristocratos (approved):

v4u6h4n:
Woo! :-)

GrandAdmiralThrawn:
I have a question about this.

I tried to compile the current version but found that the linker breaks due to AMD ROCm missing when GPU support is enabled. So for AMD GPUs, it needs the closed-source binary driver from AMD?

It would be great if it would also work with the free and open-source radeon drivers that are available for Linux and FreeBSD kernels.

It should be possible, as e.g. radeontop works on top of the free driver as well. Not sure about the NVIDIA side of things with that nouveau driver though, as I'm not using that one, so I do not know it's GPU monitoring capabilities.

Thank you!

errnoh:
@GrandAdmiralThrawn just tested and this works fine with open source drivers. (ROCm doesn't require the closed-source binary drivers either)

GrandAdmiralThrawn:
Hmm, I forgot to install rocm-smi on my box, so I did just that on my RedHat Enterprise Linux 8.9 system, packages come from the EPEL repository. GPU is a GCN 5.1 Vega inside of my AMD Ryzen 7 5700G CPU. But it still doesn't work. This is the errors I'm getting:

`[ 60%] Building CXX object CMakeFiles/btop.dir/src/btop_shared.cpp.o
[ 70%] Building CXX object CMakeFiles/btop.dir/src/btop_theme.cpp.o
[ 80%] Building CXX object CMakeFiles/btop.dir/src/btop_tools.cpp.o
[ 90%] Building CXX object CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o
[100%] Linking CXX executable btop
/tmp/ccBpUgYp.ltrans17.ltrans.o: In function `Shared::init()':
<artificial>:(.text+0x5d0f): undefined reference to `dlopen'
<artificial>:(.text+0x5d2a): undefined reference to `dlopen'
<artificial>:(.text+0x6b85): undefined reference to `dlerror'
/tmp/ccBpUgYp.ltrans24.ltrans.o: In function `Gpu::Rsmi::init()::{lambda(char const*)#1}::operator()(char const*) const [clone .isra.0]':
<artificial>:(.text+0x19): undefined reference to `dlsym'
<artificial>:(.text+0x21): undefined reference to `dlerror'
/tmp/ccBpUgYp.ltrans24.ltrans.o: In function `Gpu::Nvml::init()::{lambda(char const*)#1}::operator()(char const*) const [clone .isra.0]':
<artificial>:(.text+0x270d): undefined reference to `dlsym'
<artificial>:(.text+0x2715): undefined reference to `dlerror'
/tmp/ccBpUgYp.ltrans24.ltrans.o: In function `Gpu::Rsmi::init() [clone .isra.0]':
<artificial>:(.text+0xa123): undefined reference to `dlopen'
<artificial>:(.text+0xa13d): undefined reference to `dlerror'
/tmp/ccBpUgYp.ltrans24.ltrans.o: In function `Gpu::Nvml::shutdown() [clone .isra.0]':
<artificial>:(.text+0x2635): undefined reference to `dlclose'
/tmp/ccBpUgYp.ltrans24.ltrans.o: In function `Gpu::Rsmi::shutdown() [clone .isra.0]':
<artificial>:(.text+0x288c): undefined reference to `dlclose'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/btop.dir/build.make:225: btop] Error 1
make[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/btop.dir/all] Error 2
make: *** [Makefile:136: all] Error 2`

So I went ahead and installed _all_ ROCm libraries (hey, who knows?) and attempted another build. No luck, it still failed with the same errors.

My rocm-smi version is 5.7.1. Kernel version is 4.18.0-513.9.1.el8_9, but that really doesn't say much because of all the backporting RedHat does.

Do you have an idea of what might be going wrong here?

Thanks!

GrandAdmiralThrawn:
Thank you imwints! That resolved the issue, btop now compiles, links, runs and shows GPU information! :)

So that's for load-time linking? I've never had to use that linker flag before.

GrandAdmiralThrawn:
Ah, runtime! Yeah, that's what I meant! Just not what I actually _said_, heh. I did so much compiling and linking today (not because of btop) that my head is a mess now. ;) Interesting background information about glibc though. Well, RedHat is always behind in terms of features...

PR #534, opened 2023-05-20
Add fmtlib for string formatting (WIP)

aristocratos:
Use https://github.com/fmtlib/fmt for better string building readability.
Should also make some of the custom string functions redundant.

Help is appreciated since there is a lot of string concatenations to convert.

Example of how to format short strings:
``` cpp
out = format("{0} Text: {1:<10} + {2:>15.15}", Theme::c("main_fg"), left_aligned, right_aligned_clipped);
```

Example of how to format longer strings with repeated variables:
``` cpp
out_string = format(
	"{clear}{bg_black}{fg_white}" // Formatting string split up per output line
	"{mv1}Terminal size too small:"
	"{mv2} Width = {fg_width}{width:4d} {fg_white}Height = {fg_height}{height:4d}"
	"{mv3}{fg_white}Needed for current config:"
	"{mv4}Width = {minWidth:4d} Height = {minHeight:4d}",
	// Repeated arguments first
	"clear"_a = Term::clear, "bg_black"_a = Global::bg_black, "fg_white"_a = Global::fg_white,
	// Then using cursor placements as "scope"
	"mv1"_a = Mv::to((height / 2) - 2, (width / 2) - 11),
	"mv2"_a = Mv::to((height / 2) - 1, (width / 2) - 10),
		"fg_width"_a = (width < minWidth ? Global::fg_red : Global::fg_green),
		"width"_a = width,
		"fg_height"_a = (height < minHeight ? Global::fg_red : Global::fg_green),
		"height"_a = height,
	"mv3"_a = Mv::to((height / 2) + 1, (width / 2) - 12),
	"mv4"_a = Mv::to((height / 2) + 2, (width / 2) - 10),
		"minWidth"_a = minWidth,
		"minHeight"_a = minHeight
);
```
fmtlib reference: https://hackingcpp.com/cpp/libs/fmt.html

Some examples of formatting options:

String options:

Floating-Point options:

Signed Integer options:

Unsigned Integer options:

PR #537, opened 2023-05-23
Use `fmt` as a git submodule

Syndelis:
## Summary

This PR deletes the copy-and-pasted `fmt` headers and replaces them with a submodule at `lib/fmt` pointing to [`fmt`'s repository](https://github.com/fmtlib/fmt)

## The Problems

I'm suggesting theses changes mainly for security. There may come a time where a contributor wants to bump `fmt`'s versions and in order to do so they'd have to copy and paste the newest headers. However, everyone reviewing the PR would (*hopefully*) need to read all lines changed in the headers to guarantee nothing wrong is in place among the possible thousands of line changes of `fmt`.

These changes could be huge and reviewers may (*understandably!*) skip over or not read them as thoroughly as they would read other pieces of code.

Also, with the way `fmt` is included in this repository right now, it's very hard to say the exact version of the library in use.

## The Solution

With [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), it's possible to include `fmt` as a "link" to another repository, frozen in a specific commit. Bumping the library's version is easy and can be done in 3 steps:
1. `cd` to `lib/fmt`;
2. `git pull` on it;
3. Go back to the parent folder and `git add lib/fmt`;

What Git will actually save is the hash to the latest commit of the repository when you `git add`ed it. This solves both problems:
- Library bumps are trivial to verify because the commits will always be pulled from the official repository;
    - In fact, GitHub doesn't even show lines changed for submodules. If you want/need to check it, you can `git diff` between the changed hashes.
- The last added commit's hash is in the repository, meaning it's easy to know what's the currently used library version;

Furthermore, if changes are needed in the core of `fmt` (maybe to add new formatting options or to change current behavior) that shouldn't go upstream, one may simply fork the library and change this repository's `.gitmodules` to point to the fork rather than the official repository.

To quote [Git's documentation](https://git-scm.com/book/en/v2/Git-Tools-Submodules):
> Perhaps it’s a library that a third party developed

## Related PRs
- #534


aristocratos (approved):
Good solution for valid issues. Great PR, thanks! 👍🏼

PR #539, opened 2023-05-23
Modernize using nullptr.

nobounce:
TLDR: The problem is that NULL is of type int and relies on proper
implicit conversion which can lead to issues when using overloaded
functions

See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2431.pdf

It is also a C++ 'best practise'

### DIFF
--- a/.github/workflows/continuous-build-linux.yml
+++ b/.github/workflows/continuous-build-linux.yml
@@ -13,7 +13,7 @@ on:
       - '!src/freebsd/**'
       - 'include/**'
       - 'Makefile'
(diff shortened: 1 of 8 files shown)


aristocratos (approved):

nobounce:
@aristocratos I've not used the resolve conflicts thingy from github. Is it okay to just rebase the branch and force push it again?

nobounce:
@aristocratos 
Had to rebase again, main branch changed a lot (few new NULL's sneaked in there as well 🤣).

PR #540, opened 2023-05-24
update cpu load average display

cpalv:
load average display contains extra characters as string grows and shrinks in the case of condensed display

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -495,6 +495,7 @@ namespace Cpu {
 	int b_columns, b_column_size;
 	int b_x, b_y, b_width, b_height;
 	int graph_up_height;
+	long unsigned int lavg_str_len = 0;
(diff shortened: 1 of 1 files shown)


cpalv:
https://github.com/aristocratos/btop/assets/13182421/28b57136-fe5a-4a93-8c24-f2773c227989

screen cap showing issue, as load average increases to double digits and then decreases characters from previous string remain

aristocratos:
@cpalv 
Good fix!
But `Mv::to(b_y + b_height - 2, b_x + cx + 1) + Theme::c("main_fg")` should be moved done so it's added to `out` instead including it in `lavg_str_len` since those are of dynamic length.
If the terminal size is decreased, layout changed or the theme changed the size might not be accurate any more, so a change in the length of the load average could be missed.

cpalv:
something like this?

aristocratos (approved):

Issue #545, opened 2023-05-26
[BUG] Network graph autoscaling does not work for large bandwidths.


dorrellmw:
**Describe the bug**

With "auto" turned on and high network bandwidths, the network graph chooses arbitrary values for the graph maximum. The scale is never large enough to show the full bandwidth, and it usually just stays at 10k. The minimum bandwidth to see the issue should be around 409 MiB/s.

**To Reproduce**

Since 409 MiB/s is above 1 Gibps, many people won't be able to reproduce this issue using a physical interface, but loopback interfaces can show it. For example, in Linux:
```
$ ncat -l 10004 >/dev/null &
$ ncat 127.0.0.1 10004 </dev/zero
```
(10004 is an arbitrary port number, but numbers below 1000 usually require root privileges, and the maximum port number is 65535)

**Expected behavior**

I expect the graph to scale so that the full bandwidth fits on the graph.

**Screenshots**

![image](https://github.com/aristocratos/btop/assets/40182582/998e8de6-6fdc-4579-83dc-a7d877e97b9a)

**Info (please complete the following information):**
 - btop++ version: 1.2.13, also reproducible with git (808b8c18240962390532449573e7d472cb5da3d7).
 - Binary: compiled through Portage, also reproducible when compiled manually
 - (If compiled) Compiler and version: gcc (Gentoo 12.2.1_p20230428-r1 p2) 12.2.1 20230428
 - Architecture: x86_64
 - Platform: Linux, but should also affect FreeBSD and OSX
 - (Linux) Kernel: 6.3.3-gentoo
 - (OSX/FreeBSD) Os release version: Gentoo
 - Terminal used: rxvt-unicode with 24-bit-color patch
 - Font used: Droid Sans Mono / Inconsolata

PR #546, opened 2023-05-26
Fix integer overflows in btop_collect.cpp

dorrellmw:
Correct data types in calls to std::accumulate(). The "bandwidth" deques have type "long long", so the initial value of the accumulator (0) must also be "long long" (i.e., "0ll") to prevent integer overflows. Also, since since the bandwidth deques are (signed) "long long", the avg_speed should presumably be a signed "long long" instead of an unsigned "uint64_t". The previous behavior was for large bandwidth values to overflow the accumulator, resulting in a negative total, which then was cast to be a huge "uint64_t" value. As a consequence, the network graph autoscaling was broken for large bandwidths.

Changes are very minor but untested on FreeBSD and macOS.

Fixes aristocratos/btop#545.


aristocratos (approved):
Great, thanks for fix! 👍🏼

PR #547, opened 2023-05-26
Improve load average retrieval on linux and prefer `emplace_back`

nobounce:
Get the load average directly from the stdlib without parsing a file.

Improve readability of `Cpu::collect()` function on linux and use emplace_back where appropriate.

### DIFF
--- a/src/btop_shared.hpp
+++ b/src/btop_shared.hpp
@@ -112,7 +112,7 @@ namespace Cpu {
 		vector<deque<long long>> core_percent;
 		vector<deque<long long>> temp;
 		long long temp_max = 0;
-		array<float, 3> load_avg;
+		array<double, 3> load_avg;
 	};
 
 	//* Collect cpu stats and temperatures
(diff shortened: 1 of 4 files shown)


aristocratos:
@nobounce
Changing to using stdlib loadavg is all good, but I would ask you to remove _your_ formatting.

As mentioned in CONTRIBUTING.md:

> Split up multiple unrelated changes in multiple pull requests.

How we get the current load average and the formatting of the `Cpu::collect()` function aren't really related.
And I'm really not sure why you are changing `ll` to `LL`?

If you want to suggest a change to how the formatting should be done, then open up a new issue and argue your point.
I don't really feel strongly about it, and because of that i also rather not enforce formatting with clang-format or similar.

nobounce:
@aristocratos 
`LL` should be preferred for better readability (no confusion `ll`, `II`, depends on the font), but I'll remove that as well

aristocratos (approved):

PR #548, opened 2023-05-28
Additional Linux package managers to README.md

StitiFatah:
Additional Linux package managers


imwints:
Thanks for contributing! I think the couple of examples are enough for user the guess that the package name is btop everywhere and just change the command from dnf or others to their distros command.

Issue #508, opened 2023-03-19
utilise _FILE_OFFSET_BITS=64


nekopsykose:
**Describe the bug**

building btop with a recent commit of musl will fail, with:  

```                                                  ^
/home/demon/src/aports/community/btop/src/btop-1.2.13/src/linux/btop_collect.cpp:1082:23: error: variable has incomplete type 'struct statvfs64'
                                        struct statvfs64 vfs;
                                                         ^
/home/demon/src/aports/community/btop/src/btop-1.2.13/src/linux/btop_collect.cpp:1082:13: note: forward declaration of 'statvfs64'
                                        struct statvfs64 vfs;
                                               ^
/home/demon/src/aports/community/btop/src/btop-1.2.13/src/linux/btop_collect.cpp:1083:46: error: invalid operands to binary expression ('statvfs64' and 'int')
                                        if (statvfs64(mountpoint.c_str(), &vfs) < 0) {
                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
```

this is because of https://github.com/bminor/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc , but the reason is mostly that the `64` interfaces are legacy; they are from a period where they were created for transitional reasons.  

in modern programming, it suffices to pass `-D_FILE_OFFSET_BITS=64`, and to utilise the non-64 name, i.e. `statvfs`, and the types are 64-bit on linux. see e.g. the glibc docs for it: https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html . in musl, the non-64 names were always 64-bit; the 64 wrappers were added as a compatibility shim that are now not exposed (and will eventually be removed, left in ABI compat only).

**To Reproduce**

build against musl since https://github.com/bminor/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc without passing -D_LARGEFILE64_SOURCE

PR #556, opened 2023-06-15
ensure statvfs64 is declared

Im-0xea:
I was testing btop on risc-v with musl, and I got build errors because statvfs64 is not defined, my first instinct was to just step down to statvfs(not 64) if the LFS defines are not set, #555, because of a bit of a miss understanding of mine of how they work, sorry for that >.<, anyhow this should ensure that under musl libc and glibc statvfs64 is defined.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -22,12 +22,14 @@ tab-size = 4
 #include <cmath>
 #include <unistd.h>
 #include <numeric>
-#include <sys/statvfs.h>
 #include <netdb.h>
 #include <ifaddrs.h>
 #include <net/if.h>
 #include <arpa/inet.h> // for inet_ntop()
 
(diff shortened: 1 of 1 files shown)


nobounce:
This is related to #508 and will be closed in #510.

From the [Musl FAQ](https://wiki.musl-libc.org/faq.html#Q:-Do-I-need-to-define-%3Ccode%3E_LARGEFILE64_SOURCE%3C/code%3E-to-get-64bit-%3Ccode%3Eoff_t%3C/code%3E?):
> -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE have nothing to do with supporting large files properly but with exposing the idiotic legacy interfaces with 64 on the end of their names (like open64, lseek64, etc.). only -D_FILE_OFFSET_BITS=64 should ever be used, anywhere, even on glibc.

Issue #549, opened 2023-05-29
Multiple error messages of type: `println` is not a member of `fmt`


stefanos82:
I've just pulled the latest changes and attempted to compile the code:

```
Building btop++ (v1.2.13) Linux x86_64
Compiling src/btop_draw.cpp
Compiling src/btop.cpp
Compiling src/btop_config.cpp
Compiling src/btop_input.cpp
src/btop.cpp: In function ‘void argumentParser(const int&, char**)’:
src/btop.cpp:116:30: error: ‘println’ is not a member of ‘fmt’
  116 |                         fmt::println(
      |                              ^~~~~~~
src/btop.cpp:132:30: error: ‘println’ is not a member of ‘fmt’
  132 |                         fmt::println("btop version: {}", Global::Version);
      |                              ^~~~~~~
src/btop.cpp:148:38: error: ‘println’ is not a member of ‘fmt’
  148 |                                 fmt::println("ERROR: Preset option needs an argument.");
      |                                      ^~~~~~~
src/btop.cpp:155:38: error: ‘println’ is not a member of ‘fmt’
  155 |                                 fmt::println("ERROR: Preset option only accepts an integer value between 0-9.");
      |                                      ^~~~~~~
src/btop.cpp:164:30: error: ‘println’ is not a member of ‘fmt’
  164 |                         fmt::println(" Unknown argument: {}\n"
      |                              ^~~~~~~
src/btop.cpp: In function ‘void clean_quit(int)’:
src/btop.cpp:269:22: error: ‘println’ is not a member of ‘fmt’
  269 |                 fmt::println(std::cerr, "{}ERROR: {}{}{}", Global::fg_red, Global::fg_white, Global::exit_error_msg, Fx::reset);
      |                      ^~~~~~~
src/btop.cpp: In function ‘int main(int, char**)’:
src/btop.cpp:763:22: error: ‘println’ is not a member of ‘fmt’
  763 |                 fmt::println("WARNING: Could not get path user HOME folder.\n"
      |                      ^~~~~~~
src/btop.cpp:768:30: error: ‘println’ is not a member of ‘fmt’
  768 |                         fmt::println("WARNING: Could not create or access btop config directory. Logging and config saving disabled.\n"
      |                              ^~~~~~~
make: *** [Makefile:276: obj/btop.o] Error 1
make: *** Waiting for unfinished jobs....
20%  -> obj/btop_config.o             (1.5MiB) (05s)
[...]

PR #558, opened 2023-06-21
fmt: make btop compatible with libfmt 8 and 9

NexAdn:
Simply replacing calls to fmt::println with calls to fmt::print allows btop to be compiled with libfmt 8 and 9, allowing compatibility with the system libraries of more operating systems. This allows btop to be built against libfmt 8, 9 and 10 (and possibly more versions, which I couldn't test).

Closes: https://github.com/aristocratos/btop/issues/549


mapi68:
Debian bullseye still use libfmt 7 and it doesn't compile......

aristocratos:
@NexAdn 
The fmt submodule has been reverted to instead directly include the needed fmtlib files. So this shouldn't be an issue any more.

PR #560, opened 2023-06-21
FreeBSD swap info

rrveex:
Inspired from htop.

Moved `kvm_openfiles_wrapper` to `Shared` so it can be used by both `Mem` and `Proc`.

Tested on FreeBSD 13.2


aristocratos (approved):
Looks good to me :)
Thanks for contributing!

PR #576, opened 2023-07-15
Add "show_free" option to hide "free" values in disk and memory

tdolega:
Hi

This pull request adds a new option in config to show/hide "free" values in disk and memory boxes. I believe that some people (including me) may consider this value redundant and taking up unnecessary space, when above it the "used" value is always visible anyway.

Please review this pull request and let me know if any further changes or improvements are needed. Your feedback is greatly appreciated.

image


aristocratos (commented):
Looks good to me!
I'm all for more customizability.

I can however see the possibility of wanting to hide "free" for disks and swap but not for RAM, since "free" can be deduced from "used" as you point out (approximately anyway, since there can also be reserved space on disks).
For RAM however (total - used) doesn't necessarily equal free because of cache and so on.

Would you be willing to split this up in to two separate config flags?
Something like "disk_show_free" and "mem_show_free" or similar, and with a note in the descriptions that "disk_show_free" includes swap.

tdolega:
Yes I can do that. Now when I think about it again, I agree with you, "free" for disks and swap is something different than "free" for system memory. So I can make 2 options: One to hide "free" in disks and swap, and I'm thinking about this second option. Should it really hide only memory "free"? Or will it be ok for you to make it hide both memory "free" and "available", or even hide all 3 values "available", "cached" and "free", and just leave simple "used" like in disks. I personally would like the second and third options, but what do you (or others) think about it?

aristocratos:
@tdolega 
Personally I wouldn't see much point in only seeing the `used` RAM value, considering how Linux handles memory.
But an idea could be to have a separate toggle for all of the four RAM values?

PR #587, opened 2023-07-25
Add Debian/Ubuntu binary install command to README.md

ipepe:
Hey. I noticed that there is no installation command in readme for debian/ubuntu linuxes and added that, as I managed to successfully install it using `sudo apt install -y btop`

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,10 @@ Also needs a UTF8 locale and a font that covers:
 	```sh
 	pkg install btop
 	```
+* **Ubuntu/Debian**
+	```bash
+	sudo apt install -y btop
+	```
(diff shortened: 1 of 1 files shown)


nobounce:
I'd say this is a good idea but the README is already pretty large, so what do you think about collapsing the different distro specific instructions by default with a tag to keep the whole section clean and tidy?

<details>
<summary>Debian</summary>

```bash
sudo apt install btop
```

</details>

imwints:
Thanks for contributing! I think the couple of examples are enough for user the guess that the package name is btop everywhere and just change the command from dnf or others to their distros command.

Issue #583, opened 2023-07-23
[REQUEST] Up-to-date CMake fork and CMake support


nobounce:
The [cmake fork](https://github.com/jan-guenter/btop) by @jan-guenter seems unmaintained at the moment.

With the fork as inspiration I've created a new [CMakeLists.txt](https://github.com/nobounce/btop/blob/cmake/CMakeLists.txt) to build btop with CMake.
I'm committed to maintaining this file as an alternative to the Makefile.

And I want to ask again, after reading the discussion #82, if you want to consider including the CMakeLists.txt file alongside the Makefile in the official repository. I would like to maintain it (in the hypothetical case it gets included and you wouldn't want to). You've made good points against the arguments provided and I agree with you that the Makefile is completely capable of building `btop` and does not need to be replaced.

The point I want to add in favor of CMake is maintainability. I've written the adjustments for LLVM and it was not trivial (86 changes). I've maintained a private CMake file for a couple of months and adding support was a one-liner, which should just point out the convenience of CMake. It was a lot easier and it will be for future tasks, like handling the `fmt` dependency (e.g.: using the system fmtlib and fallback to the submodule, yeah a weird usecase) or GPU support (detecting system capabilities at compile time, excluding flags and checks based on the configuration). The Makefile will grow substantially with every new feature and will become harder to maintain. Make wasn't designed for this.

I would be happy if you can link the new repository in the README.
Honestly I don't think anyone besides me will use it unless it gets officially supported but it's enough for now to bring CMake up-to date.

aristocratos:
@nobounce 
Adding a CMakeLists.txt and using a `build` folder for the files cmake creates would work fine.
I believe I even mentioned that the last time this was discussed, as long as someone is willing to maintain it I'm all for it. 👍🏼

But I noticed the compile time with the generated files from the CMakeLists.txt you linked above was really slow:
1 min 21 sec versus 24 seconds from the original Makefile. So you should probably add some parallelization by default in the CMakeLists.txt file.

Not sure why you needed to make "86 changes" for llvm support? Should only be 2 or 3 compile flags that differ no?

I'm planning on removing the "auto compiler finder" and some other "extra" functionality in the Makefile, and instead just print out a warning if the compiler isn't known to work.

> The Makefile will grow substantially with every new feature and will become harder to maintain. Make simply wasn't designed for this.

Has not been my experience so far.

[...]

PR #589, opened 2023-07-26
Add CMake support for Linux

nobounce:
Some parts for the other OSes are already there but I'm still testing.

Don't know if the documentation is to your liking, feedback appreciated!

Closes: #583

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,9 +51,8 @@ bin
 btop
 .*/
 
-
-#do not ignore .github directory
-!.github
+# Don't ignore .github directory
+!.github/
 
 # Ignore files created by Qt Creator
 *.config
@@ -64,3 +63,19 @@ btop
 *.cxxflags
 *.files
 *.includes
+
+# CMake
+CMakeLists.txt.user
+CMakeCache.txt
+CMakeFiles
+CMakeScripts
+Testing
+Makefile
+cmake_install.cmake
+install_manifest.txt
+compile_commands.json
+CTestTestfile.cmake
+_deps
+
+# CLion
+cmake-build-*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,167 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# CMake configuration for btop
+#
+
+cmake_minimum_required(VERSION 3.12)
+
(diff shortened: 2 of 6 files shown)


aristocratos (changes requested):
@nobounce 
Looks good so far, is it ready to merge?

I would like to ask that you move the instructions for the regular Makefile above the CMake instructions and add `(Community maintained)` or something similar after `### With CMake` if that's alright?

nobounce:
Yes, I've addressed the review

Also see the commit message for some quirks!

Otherwise ready to go

aristocratos:
Unless you plan to update the CMakeLists.txt file for every release, look at the file @jan-guenter made: https://github.com/jan-guenter/btop/blob/main/CMakeLists.txt for how to get the version string from the source.

> FreeBSD is not able to create a static executable for now.

This is already known for both FreeBSD and MacOS, which is why the Makefile only statically links libgcc and libstdc++ for them. Something that's also mentioned in the compilation instructions in the README.md

nobounce:
> > FreeBSD is not able to create a static executable for now.

This was just referring this particular `CMakeLists.txt` on FreeBSD, I've not added the work around :)

Regarding the version:
I'd rather see the build system like Make or CMake define a version and pass it to the source. This allows for things like the git hash to be appended to the version if desired.
Otherwise I can update this manually or even with CI on my branch

nobounce:
Ok, I'm a lot happier now :)
Added some useful options and documented them.

Added the `cmake/Modules` directory which contains files to dynamically detect the libraries required on FreeBSD.

Static linking in FreeBSD works now with Clang, however Clang only works with FreeBSD 14 :/
 (or with GCC's libraries which defeats the point kind of). The install `libc++` version depends on the FreeBSD release and cant be updated with a port

aristocratos (approved):

aristocratos:
@nobounce 
You should be able to merge this yourself since it's approved or am I wrong in that assumption?

If the PR is done that is?

nobounce:
@aristocratos 
for me it says:
>  Changes approved
1 approving review by reviewers with write access.
This branch has no conflicts with the base branch
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

PR #595, opened 2023-08-09
Add power CPU consumption monitoring on windows

YuriiShkrobut:


robke96:
### Seems to work, tried with 2 systems: 
- AMD Ryzen 7 1700 | Fedora 41
- AMD Ryzen 5 4600G | Fedora 41 Live ISO

For some reason, I had to run it as ``sudo``, on main R7 1700 system, I get error: 
```json
ERROR: Exception in runner thread -> Cpu:: -> key not found
```
On an R5 4600G system, btop launches without ``sudo`` but doesn't display power usage; with sudo, it does.

Ryzen7_1700
Ryzen5_4600G

Tom94:
Apologies for resurrecting this old PR -- I'd *love* to be able to monitor my CPU's power draw.

I'd be happy to apply whatever changes are needed and open a new PR to get this merged, but wanted to first ask for why it didn't go in in the first place.

Note: I tested this on my amd threadripper 9980x and it works (with sudo). Definitely not intel-exclusive.

aristocratos:
@Tom94 
Hi, I believe what needs fixing is the logic for how it's placed inside the CPU core panel.

In the post above for example you can see in the first screenshot that it overwrites where core 5 should be showing.

So it needs to be more adaptable, move around and change size depending on window size and number of cores (just like how the load average does).

Could possibly need some update to the `make setcap` command in the Makefile to allow reading the power usage without root privileges also.

You are more than welcome to fix the current PR or open a new one :)

Tom94:
Thank you, I'll get to work then! :)

I noticed that the load average display also overlaps a core when btop can't display them all, see this screenshot:
image

If I'm already looking into how to cleanly display the pkg power draw, I assume you wouldn't mind me fixing the LAV alongside?

I propose always reserving an entire line for PWR (named analogously to GPU PWR) and LAV akin to op's screenshot:
image

I'll play around with both placing it at the top (right below `CPU xxxxx 7%`, same as the GPU box) as well as the very bottom (matching current LAV display if all cores fit) and post both.

Tom94:
I also looked into using `setcap` to let btop read the file. It's unfortunately a bit heavy-handed... one has to grant btop access to read *any* file via CAP_DAC_READ_SEARCH. I couldn't find a way to make it more finegrained.

One could also target *just* the relevant file via
```sh
sudo chmod 400 /sys/class/powercap/intel-rapl*/*/energy_uj
```
on btop install, but that seems even worse from a security standpoint because other processes can then read the file, too. I'll go with CAP_DAC_READ_SEARCH unless you have objections.

aristocratos:
> If I'm already looking into how to cleanly display the pkg power draw, I assume you wouldn't mind me fixing the LAV alongside?

That would be great, hadn't noticed it was also broken...

> I'll go with CAP_DAC_READ_SEARCH unless you have objections.

No, that would be the correct way of handling it in my opinion.

`make setcap` is an optional step, and the alternative would be to run with `sudo`, which also would give access to all files.

Would probably be a good idea to also add a note in the readme somewhere that `make setcap`, `make setuid` or `sudo` is required to show CPU power.

aristocratos:
Implemented in #1227

PR #600, opened 2023-08-23
[macos/freebsd] support gcc13

joske:
I tried compiling on my hackintosh, but it fails because it fails to find gcc-13, and subsequently fails with apple clang of course.

Homebrew now installs gcc@13 so we better support that too.


nobounce:
Can you add another branch for `g++13`, for FreeBSD?

aristocratos (approved):
The g++ "autodetection" is planned to be removed and replaced with just a warning if gcc < 10 or clang < 16.

PR #602, opened 2023-08-25
Fix getting zfs pool name with '.' char in freebsd

jfouquart:
### Take a look at 10.3-64 in the screenshot and at the top right of the window:

before patch
btop-without-patch
and after
btop-with-patch

### DIFF
--- a/src/freebsd/btop_collect.cpp
+++ b/src/freebsd/btop_collect.cpp
@@ -543,14 +543,16 @@ namespace Mem {
 
 	// find all zpools in the system. Do this only at startup.
 	void get_zpools() {
+		std::regex toReplace("\\.");
 		PipeWrapper poolPipe = PipeWrapper("zpool list -H -o name", "r");
+
 		while (not std::feof(poolPipe())) {
 			char poolName[512];
 			size_t len = 512;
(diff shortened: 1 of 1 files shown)


src/freebsd/btop_collect.cpp
```diff
@@ -543,14 +543,16 @@ namespace Mem {
...
 		PipeWrapper poolPipe = PipeWrapper("zpool list -H -o name", "r");
+
 		while (not std::feof(poolPipe())) {
 			char poolName[512];
 			size_t len = 512;
 			if (fgets(poolName, len, poolPipe())) {
 				poolName[strcspn(poolName, "\n")] = 0;
 				Logger::debug("zpool found: " + string(poolName));
-				Mem::zpools.push_back(poolName);
+				Mem::zpools.push_back(std::regex_replace(poolName, toReplace, "%25"));
```
joske, line 555: What is `"%25"`?

joske: Never mind, found:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257316

joske (approved):

Issue #518, opened 2023-04-22
[REQUEST] add a shortcut to maximize a box


c02y:
**Is your feature request related to a problem? Please describe.**

Normally I open all boxes in the btop window, but sometimes, the process box is too small to show all the info such as the long `command` column, so I have to disable other boxes using `1/2/3` keys one by one, after I get the info in process box, I reopen all the `1/2/3` boxes again one by one.

**Describe the solution you'd like**

Could you please add a new shortcut to maximize a box such as `M-1/2/3/4` to maximize `1/2/3/4` box, pressing `M-1/2/3/4`  again it restores the layout?

**Describe alternatives you've considered**

**Additional context**

PR #603, opened 2023-08-25
Add shortcut to maximize a box

mdxio:
Attempt at solving the first part of  #518 which is to maximize a box

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -519,6 +519,12 @@ namespace Config {
 		return true;
 	}
 
+	void maximize_box(const string& box) {
+		current_boxes.clear();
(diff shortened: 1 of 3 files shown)


aristocratos (changes requested):
@mdxio 
I like the idea and effort but I'm not sure about the implementation.
When the GPU code gets merged there will be a dynamic number of extra boxes depending on number of gpus available and maintain a consistent experience (using `shift` as a modifier) will be difficult considering different keyboard layouts and so on.

Maybe a better approach would be to use a single key (sadly `m` is taken) and then the next number entered in a specific time period (say 1 second) would maximize the box.

You should probably also merge #604 in to this same PR since they are related.

Issue #535, opened 2023-05-20
[REQUEST] Continued implementation of fmtlib


aristocratos:
Continued implementation of https://github.com/fmtlib/fmt for better string building readability.
Should also make some of the custom string functions redundant.

Help is appreciated since there is a lot of string concatenations to convert.

Example of how to format short strings:
``` cpp
out = format("{0} Text: {1:<10} + {2:>15.15}", Theme::c("main_fg"), left_aligned, right_aligned_clipped);
```

Example of how to format longer strings with repeated variables:
``` cpp
out_string = format(
	"{clear}{bg_black}{fg_white}" // Formatting string split up per output line
	"{mv1}Terminal size too small:"
	"{mv2} Width = {fg_width}{width:4d} {fg_white}Height = {fg_height}{height:4d}"
	"{mv3}{fg_white}Needed for current config:"
	"{mv4}Width = {minWidth:4d} Height = {minHeight:4d}",
	// Repeated arguments first
	"clear"_a = Term::clear, "bg_black"_a = Global::bg_black, "fg_white"_a = Global::fg_white,
	// Then using cursor placements as "scope"
	"mv1"_a = Mv::to((height / 2) - 2, (width / 2) - 11),
	"mv2"_a = Mv::to((height / 2) - 1, (width / 2) - 10),
		"fg_width"_a = (width < minWidth ? Global::fg_red : Global::fg_green),
		"width"_a = width,
		"fg_height"_a = (height < minHeight ? Global::fg_red : Global::fg_green),
		"height"_a = height,
	"mv3"_a = Mv::to((height / 2) + 1, (width / 2) - 12),
	"mv4"_a = Mv::to((height / 2) + 2, (width / 2) - 10),
		"minWidth"_a = minWidth,
		"minHeight"_a = minHeight
);
```
fmtlib reference: https://hackingcpp.com/cpp/libs/fmt.html

Some examples of formatting options:

String options:
![](https://hackingcpp.com/cpp/libs/fmt_spec_string.svg)

Floating-Point options:
![](https://hackingcpp.com/cpp/libs/fmt_spec_float.svg)

Signed Integer options:
![](https://hackingcpp.com/cpp/libs/fmt_spec_int.svg)

Unsigned Integer options:
![](https://hackingcpp.com/cpp/libs/fmt_spec_unsigned.svg)

nobounce:
You should think about adding a .clang-format file to the project, so the multi-line statements like
```
out_string = format(
[...]

PR #606, opened 2023-08-29
Use fmt for logging and allow to log to system daemon directly

imwints:
This PR improves the logger in a few ways:
1. The logger now lives in it's own file, which avoids unnecessary recompilation
2. The logger uses `fmt`, See #535 
   - The logging functions now have the same syntax as `fmt::format`
   - The logging backend logs with `fmt::print` instead of `std::ofstream`
3. `FMT_HEADER_ONLY` is now defined by Make and passed to the compiler. Other files including `<fmt/*.h>` are now independent of the definition in `btop_tools.hpp`
4. By default, if `<syslog.h>` is present, btop will log it's messages there
5. When compiled with `make ENABLE_JOURNALD=true` btop can directly log to journald. This can log interesting metadata like the source file location for each message
6. Parse `BTOP_LOG_LEVEL` environment variable and set the log level accordingly

On systemd systems, with either `syslog` or `journald` logging, btops logs can be read with `journalctl -t btop`

If this is not a good default, it's an easy fix to put this behind a `ENABLE_...` flag!
I'm curious if this is a nice feature or not. Looking forward for feedback :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,7 @@ target_compile_definitions(btop PRIVATE
   _GLIBCXX_ASSERTIONS _LIBCPP_ENABLE_ASSERTIONS=1
   # Only has an effect with optimizations enabled
   $<$<NOT:$<CONFIG:Debug>>:_FORTIFY_SOURCE=2>
+  FMT_HEADER_ONLY
 )
 
 # Enable GPU support
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,8 @@ include(CMakeDependentOption)
 option(BTOP_STATIC "Link btop statically" OFF)
 option(BTOP_LTO "Enable LTO" ON)
 option(BTOP_USE_MOLD "Use mold to link btop" OFF)
+option(BTOP_SYSLOG "Log with syslog instead of log file" OFF)
+option(BTOP_JOURNALD "Directly log to journald instead of log file" OFF)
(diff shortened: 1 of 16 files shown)


nobounce:
Apples `shell` command within make has some kind of issue with line 168, but I'm not sure why

nobounce:
Fixed the shell issue. You need to escape the `#` in Apples make, which on the other hand doesn't work on Linux... `#` omitted :)

aristocratos:
@nobounce
Nice work!
I like the idea, but would prefer the defaults to keep the code contained (with only simple logfile logging).
But having the option to easily enable `syslog` or `journald` options would be great.
Don't forget to add documentation in the compile sections of the README.md

aristocratos (changes requested):
@nobounce
Nice work!
I like the idea, but would prefer the defaults to keep the code contained (with only simple logfile logging).
But having the option to easily enable syslog or journald options would be great.
Don't forget to add documentation in the compile sections of the README.md

nobounce:
@aristocratos 
I'm on it

nobounce:
@aristocratos 
Rebased, added the same logic to CMakeLists.txt, changed default back to plain log file and added documentation

nobounce:
You can actually omit your additional commit (aristocratos Merge PR `#XY` from ZZ) when you merge PR's unless this is done for a reason

nobounce:
@aristocratos
Rebased, added the same logic to CMakeLists.txt, changed default back to plain log file and added documentation

nobounce:
@aristocratos
In theory this is good to go

PR #607, opened 2023-08-31
openbsd bringup

joske:
Everything seems to work. There are some limitations though:

- disk i/o is per disk, so all mount points on that disk show same I/O
- CPU temp is per core, not thread, so all threads on same core have same temp
- same for CPU usage

I noticed that using `<` and `>` to change the sort doesn't work (in linux it doesn't work either). Is this a know issue?

To compile you need to
```bash
pkg_add g++ git gmake
```

Select g++ 11 when it prompts for the version.

BTW, this is very fast, collect takes ~2 ms. On macOS, this is in the order of 50 ms or so (linux ~20 ms). Also it only uses about 7MB and doesn't leak any memory.


aristocratos:
@joske 
Nice work man!
Haven't had time to go through the code and setup a VM yet, but will get to it soon.

> I noticed that using < and > to change the sort doesn't work (in linux it doesn't work either). Is this a know issue?

That would be the `left` and `right` arrow keys :) The arrow symbols aren't available in some fonts that's why I'm using the `less than` and `greater than` symbols instead.

joske:
> @joske Nice work man! Haven't had time to go through the code and setup a VM yet, but will get to it soon.
> 
> > I noticed that using < and > to change the sort doesn't work (in linux it doesn't work either). Is this a know issue?
> 
> That would be the `left` and `right` arrow keys :) The arrow symbols aren't available in some fonts that's why I'm using the `less than` and `greater than` symbols instead.

I could have sworn that in the past < and > worked.

joske:
Update on OpenBSD ports build: it's not possible to build with gcc11 in a port, so it must be clang, which is too old at the moment. There is a LLVM 16 port, but the libs/include files are not up to date yet (problems with ranges). They tell me it will probably have to wait until OpenBSD 7.4 (end of year). 

But you can build local local builds using gcc 11.

joske:
There is a lot of overlap between OpenBSD/FreeBSD and macOS (which is in essence also a BSD). Some consolidation work would be nice once this PR is merged.

joske:
@aristocratos bump?

joske:
@aristocratos it's been almost 4 months :wink:

aristocratos:
@joske 
I haven't forgotten it, just been really busy.

But have some vacation coming up from christmas and onward, so will have some time to go through current PR's and my own WIP stuff then 🤞

aristocratos (approved):

aristocratos:
@joske 
Nice work! Seems to be working great.

PR #609, opened 2023-09-05
Fix short conversion of 1000-1023 *iB

scorpion-26:
floating_humanizer([1000-1023], true) with base 8 returns "2K", whereas it should return "1.0K" to align with other formats. The conversion is also broken for all other units(e.g. 1023MiB is also broken and returns "2G")

AFAIK, the branch ```out.size() > 3``` could also be hardcoded to ```out = "1.0"; start++;```, since only the range 1000-1023 triggers it. I left it out for now.


aristocratos (approved):
Great fix, thanks!

Issue #566, opened 2023-06-28
[REQUEST] Remove compiler version selection in the Makefile


imwints:
I don't think it's necessary to set the compiler to a `g++-<version>` anymore when running make.
I'm encountering an issue where I have my `/usr/bin/g++` pointing to GCC 13 (which is not yet covered by the Makefile) and I've installed GCC 10 alongside for testing. This configuration selects GCC 10 over 13.

I think we should trust that the distribution has set g++ to the latest and most stable version of the compiler that should be used on that system. If the user wants to, they can pass a specific compiler as an environment variable. If we clearly state this in the installation instructions it should be sufficient.
The worst thing that can happen is, as it does now, that the compilation fails. There is also a version check so if the compiler version is too old this gets printed out.

And I don't think we should take care of those rare situations where, e.g. GCC 9 and GCC 11 are installed and
```
/usr/bin/g++ -> /usr/bin/g++-9
```
This is a setup where it's up to the user or developer to call `make CXX=g++-11`, not the Makefile. It's just too much to account for.

Of course we could with no issue just add a few more checks for every new compiler but this is really just putting more work on maintaining that Makefile. The solution above is a lot simpler.

correabuscar:
I've hit this on Gentoo, as follows:
```
...
Linking and optimizing binary...
x86_64-pc-linux-gnu-g++ -o bin/btop obj/btop.o obj/btop_config.o obj/btop_draw.o obj/btop_input.o obj/btop_menu.o obj/btop_shared.o obj/btop_theme.o obj/btop_tools.o obj/linux/btop_collect.o -Wl,-O1,--sort-common,--as-needed,-z,relro -pthread -D_FORTIFY_SOURCE=2 -D_GLIBCXX_ASSERTIONS -D_FILE_OFFSET_BITS=64 -fexceptions -fstack-clash-protection -fcf-protection -fstack-protector   -Wall -Wextra -pedantic
100% -> bin/btop                      (27MiB) (00s)

Build complete in (01s)
>>> Source compiled.

>>> Test phase [not enabled]: sys-process/btop-1.2.13-r2

>>> Install sys-process/btop-1.2.13-r2 into /var/tmp/portage/sys-process/btop-1.2.13-r2/image
make --no-keep-going --output-sync=target -j18 PREFIX=/usr DESTDIR=/var/tmp/portage/sys-process/btop-1.2.13-r2/image install
ccache: error: Could not find compiler "g++-12" in PATH
[...]

PR #613, opened 2023-09-13
Remove detailed compiler check

imwints:
Currently g++ is forced to be the latest intalled version

Since we give an error if the version of the compiler is not sufficient, we can remove this check/overwrite

It's now possible to specify an older version (on purpose) and new compiler version don't need to be added manually

Closes: #566

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,34 +73,9 @@ ifeq ($(CXX_IS_CLANG),true)
 	ifneq ($(shell test $(CXX_VERSION_MAJOR) -lt $(MIN_CLANG_VERSION); echo $$?),0)
 		CLANG_WORKS := true
 	endif
-endif
-ifeq ($(CLANG_WORKS),false)
-	#? Try to find a newer GCC version
-	ifeq ($(shell command -v g++-13 >/dev/null; echo $$?),0)
-		CXX := g++-13
-	else ifeq ($(shell command -v g++13 >/dev/null; echo $$?),0)
-		CXX := g++13
-	else ifeq ($(shell command -v g++-12 >/dev/null; echo $$?),0)
-		CXX := g++-12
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-build-freebsd.yml
```diff
@@ -38,11 +38,11 @@ jobs:
         with:
           release: 13.2
           usesh: true
-          prepare: |
-            pkg install -y gmake gcc11 coreutils git
-            git config --global --add safe.directory /Users/runner/work/btop/btop
```
aristocratos, line 43: Why is line 43 removed?

From https://github.com/aristocratos/btop/actions/runs/6177943184/job/16770149424?pr=613#step:3:1461

```
  Build complete in (02m:25s)
  fatal: detected dubious ownership in repository at '/Users/runner/work/btop/btop'
  To add an exception for this directory, call:
  
  	git config --global --add safe.directory /Users/runner/work/btop/btop
```

aristocratos (changes requested):
I believe I mentioned I was planning to do this (but forgot), so good initiative 👌

.github/workflows/continuous-build-freebsd.yml
nobounce: Good catch, again this wasn't supposed to be committed. Fixed it

aristocratos (approved):

aristocratos:
@kz6fittycent 
The build for the workflow `test-snap-can-build.yml` fails after this was merged.
I'm guessing the compiler needs to be specified in the build script, I can't however see the actual build command in the workflow, so not sure how to fix it.

aristocratos:
@kz6fittycent 
Added `CXX=g++-11` to the make flags in https://github.com/aristocratos/btop/blob/main/snap/snapcraft.yaml
And it's working again.

Issue #614, opened 2023-09-13
[REQUEST] Switch to new mainted hash map and set


nobounce:
The robin hood hashing containers used here are not actively developed anymore, but the same author has a successor called [unordered_dense](https://github.com/martinus/unordered_dense).

Promises fast iteration speeds and low memory usage. I think it's an idea to switch

PR #617, opened 2023-09-13
Replace robin hood hash set/map with unordered-dense set/map

nobounce:
Closes: #614


aristocratos:
@nobounce 
It might be the case that we could use `std::unordered_map` instead.
I believe the sizes of the containers used in btop probably never reach a size where a custom hash map would show any measurable difference in performance.

nobounce:
@aristocratos
Totally agree and didn't think of that at all.

Should I discard this PR then?

aristocratos:
Probably better to create a new PR if you wanna do the `std::unordered_map` replacement, to not have a bunch of unnecessary commits in the history.

Issue #598, opened 2023-08-20
[REQUEST] Aggregate child process values in tree mode


magneticflux-:
**Is your feature request related to a problem? Please describe.**
I'd like to view the process tree, but also sort by total child CPU usage so that (_non-collapsed_) subtrees with a lot of resource usage appear higher.

**Describe the solution you'd like**
An option to report aggregate statistics about child processes regardless of the nodes being collapsed or not.

**Describe alternatives you've considered**
Introduce another toggle like `per-core` called `aggregate`?

**Additional context**
This seems mostly-implemented already since collapsed nodes already sum their values, I'd just like to see the sums while also seeing the child processes.

nobounce:
I've implemented this in https://github.com/nobounce/btop/tree/aggregate-child-processes, but I find this hardly distinguishable from the normal per-process statistics. Any idea how this could be made visible that this a the combined resource usage so user won't freak out if they are not aware of this?

magneticflux-:
Awesome that works perfectly! It makes a huge difference when there's deeply-nested trees and I want to show the most "active" ones first, like compiling a bunch of packages in NixOS.

<details>
<summary>Details</summary>

![image](https://github.com/aristocratos/btop/assets/9124288/4238d0af-afa7-4eca-8dfe-2978e27a6f28)

Overriding the `btop` package in Nix to use this branch:
```nix
(btop.overrideAttrs (prev: {
  name = "btop/aggregate-child-processes";
  src = pkgs.fetchFromGitHub {
    owner = "nobounce";
    repo = "btop";
    rev = "aggregate-child-processes";
    hash = "sha256-wWpcqnZXBH4nIE/EzWWooFyzM0odX2ewAHq1fIuX5M8=";
  };
}))
```

</details>


Before this, the `systemd/sshd/nixos-rebuild` tree wasn't guaranteed to be at the top because random background services with 2% CPU usage got sorted above `sshd` with 1%, even though the _tree_ was a lot more active.

---

To make it explicitly clear to users, you could split the display of values into "self" and "child" parts, so something like `/sbin/init` spawning a child with 5G memory would look like `1.6M + 5G`. The downside is that it's much wider than the original.

In my opinion, having the aggregate stats option default to off and making the description really clear is a good approach if you're concerned about confusion. I suggest the description be:
> In tree-view, include all child resources with the parent even while expanded.

nobounce:
[...]

PR #618, opened 2023-09-14
Add option to accumulate a child's resources in parent in tree-view

nobounce:
Closes: #598


aristocratos:
@nobounce 
A way to fix the accumulated threads being cut off could be to add a check in `Proc::draw` that if threads > 1000, then divide by 1000 and round with 1 decimal and a `k`, like `1.1k` for 1100, and if in the extreme case of more than 100k threads just show as `99k+`.

nobounce:
Used this for testing
```cpp
#include <chrono>
#include <csignal>
#include <cstdint>
#include <limits>
#include <thread>
#include <vector>

int main() {
	std::vector<std::thread> threads;

	for (uint64_t i = 0; i < 10000; ++i) {
		threads.emplace_back([&i] {
			auto num = i;
			while (true) {
				std::this_thread::sleep_for(std::chrono::milliseconds(250));
				if (num == std::numeric_limits<uint64_t>::max()) { num = 0; }
				++num;
			}
		});
	}

	sigset_t sigset;
	sigemptyset(&sigset);
	sigaddset(&sigset, SIGINT);
	pthread_sigmask(SIG_BLOCK, &sigset, nullptr);

	std::thread signal_handler { [&sigset] {
		int signum = 0;
		while (signum != SIGINT) {
			sigwait(&sigset, &signum);
		}
	}};

	signal_handler.join();

	for (auto& thread : threads) {
		thread.join();
	}
}
```

aristocratos (approved):
Looks good!

PR #627, opened 2023-09-22
Try get terminal size of "/dev/tty" if stdout fails

imwints:
LLDB does not set the terminal sizes for stdout so try to get them for "/dev/tty". This helps to get back traces on arm64 where gdb is not supported


aristocratos (changes requested):
Great idea!

But I think this could be reworked to be a bit more efficient.
For example if the deafult `STDOUT_FILENO` fails, setting a bool that switches over to using `/dev/tty` instead.

nobounce:
@aristocratos


> Great idea!
> 
> But I think this could be reworked to be a bit more efficient. For example if the deafult `STDOUT_FILENO` fails, setting a bool that switches over to using `/dev/tty` instead.

I can't quite follow what you mean by that. Currently it is checked if `STDOUT_FILENO` can be used to get valid terminal dimensions, if it's not the case it tries to fall back to `/dev/tty` ...

Where would the bool come into play?

aristocratos:
> Where would the bool come into play?

`Term::refresh` is called at a minimum once every update cycle, since [SIGWINCH](https://www.gnu.org/software/libc/manual/html_node/Miscellaneous-Signals.html#index-SIGWINCH) isn't always reliable.
So if we are running on a system where `STDOUT_FILENO` fails to get current size, we will continue to make that same syscall every time we run `Term::refresh` even though it's gonna fail every time.

Syscalls cost a lot more than branching.

aristocratos (approved):

PR #628, opened 2023-09-22
Remove robin-hood-hashing dependency

nobounce:
As per https://github.com/aristocratos/btop/pull/617 using a specialized hash map/set is not worth the effort


aristocratos:
This one is gonna have to wait a little bit because of the amount of touched code.

But in the meantime, would you be willing to do some performance testing between `std::unordered_map` and the robin hood hash maps?

It would be interesting to see if we actually are giving up any performance with the switch.

aristocratos:
Btw @nobounce, I hope you know that I really appreciate the interest you've taken in btop and your efforts to improve the code!
I sadly don't have as much free time as I used to.
But you and the others that graciously choose to contribute improvements and new functionality to btop is a great help! ❤

nobounce:
> It would be interesting to see if we actually are giving up any performance with the switch.

I had a look at it. I only looked at the memory footprint of btop itself. `unordered_dense` is around 400kB better than the current version with `robin-hood-hashing`. The standard implementation (of libstdc++) uses around 2M more RAM than the current implementation. In both this and the previous PR I deleted calls to `compact()` without a replacement. The proper function to replace it is `rehash()` in both the standard lib and `unordered_dense`. This saves some memory as well

PR #630, opened 2023-09-24
Fix signal list on non-linux/weird linux platforms

lvxnull:
The previous list of signals was correct for linux only, with some exceptions. This one handles non-linux platforms, as well as linux on some cpu architectures that have signals assigned differently.


aristocratos (approved):
Thanks for contributing!

PR #631, opened 2023-09-27
Update workflows

imwints:
Skipped CI in the last push because the previous run succeeded and I just removed code behind comments.

And yeah next time I'm gonna do that more often ;)


aristocratos:
Missing reason for why the workflows needed updating?
Why are the workflows renamed?
And why are the macos workflow replaced with a new file when the rest is just modified?

nobounce:
> Missing reason for why the workflows needed updating?
There where some extraneous steps removed. Also keeping them as similar and simple as possible helps maintainability (hopefully)

> Why are the workflows renamed?
Renaming of workflows was pushed accidentally

> And why are the macos workflow replaced with a new file when the rest is just modified?
Renamed the files so there is less tab completion to do when opening :)
I think diff can't detect enough similarities in the two macos files

Issue #392, opened 2022-07-16
[BUG] Btop showing blank screen on Ubuntu 20.04 LTS


martinszeltins:
**Describe the bug**

When I launch `btop` from my Ubuntu 20.04 it simply shows a blank black screen and nothing happens. I am using zsh with oh-my-zsh but it also does the same on bash. Please see the attached screenshots and screencast video. Is there a way I can fix this issue on my machine?

**Screenshots**
![Peek 2022-07-16 11-01](https://user-images.githubusercontent.com/34019878/179346162-27650a82-367b-4a43-bf22-2cfa1c8e8342.gif)

**Info (please complete the following information):**
 - btop++ version: `1.2.8`
 - Binary: Installed using: I tried both from binary and from snap - same results
 - Architecture: `x86_64`
 - Platform: `Ubuntu 20.04.4 LTS`
 - (Linux) Kernel: `5.15.0-41-generic`
 - Terminal used: `zsh` and `bash` (both have the same result)

contents of `~/.config/btop/btop.log`

````
$ cat ~/.config/btop/btop.log
   1   │ 
   2   │ 2022/07/16 (10:57:09) | ===> btop++ v.1.2.8
   3   │ 2022/07/16 (10:57:09) | DEBUG: Starting in DEBUG mode!
   4   │ 2022/07/16 (10:57:09) | INFO: Logger set to DEBUG
   5   │ 2022/07/16 (10:57:09) | DEBUG: Using locale en_US.UTF-8
   6   │ 2022/07/16 (10:57:09) | INFO: Running on /dev/pts/3
````

martinszeltins:
I just restarted my computer and it is working correctly now. I don't know what the problem was.

ErrorNoInternet:
Does using a different font change anything?
Or try adding `force_tty = True` to the btop.conf file (~/.config/btop/btop.conf)

KamilKleina:
Facing the same issue, it was working fine but after update today, it stopped working. Tried to install older version like 1.2.7 but still the same. Using ArchLinux.

jamesawebb:
Gah - encountered for the first time today.  A reboot fixed the issue - but it's disconcerting.  Adding this detail as another data point.

crestfallnatwork:
Had the same issue. Ran strace, found out that statvfs was blocking due to some automount magic I had setup a long time ago. 
I solved my issue by removing the said automount magic. But statvfs blocks for network devices as well.
[...]

PR #633, opened 2023-09-27
[fix] Made disks statvfs logic asynchronous.

### LINKED ISSUE #632: [BUG] BTOP freezes when a mount point is faulty
**Describe the bug**
BTOP freezes with a blank screen when there is a automatic mount point that is faulty or takes some time to mount. 

**To Reproduce**

Add a faulty mount point like this to fstab:
`/dev/idonotexist /home/user/mnt ext4 rw,noauto,x-systemd.automount,x-systemd.device-timeout=10s`
Start BTOP.  
   
**Expected behavior**
  
BTOP should run.

**Info:**
 - btop++ version: `1.2.13`

**Additional context**
#392 
[This comment](https://github.com/aristocratos/btop/issues/392#issuecomment-1719056607)

crestfallnatwork:
Fixes #632 

Read the following comment to get the full context of the problem: 
* [Comment](https://github.com/aristocratos/btop/issues/392#issuecomment-1719056607)  

This pull request mostly follows the solution described in the following comment:
* [Comment](https://github.com/aristocratos/btop/issues/392#issuecomment-1722193694)

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -29,6 +29,7 @@ tab-size = 4
 #include <net/if.h>
 #include <arpa/inet.h> // for inet_ntop()
 #include <filesystem>
+#include <future>
 
 #if !(defined(STATIC_BUILD) && defined(__GLIBC__))
 	#include <pwd.h>
@@ -48,12 +49,17 @@ using std::numeric_limits;
 using std::round;
 using std::streamsize;
 using std::vector;
+using std::future;
+using std::async;
+using std::pair;
+
 
 namespace fs = std::filesystem;
 namespace rng = std::ranges;
 
 using namespace Tools;
 using namespace std::literals; // for operator""s
+using namespace std::chrono_literals;
 //? --------------------------------------------------- FUNCTIONS -----------------------------------------------------
 
 namespace Cpu {
@@ -939,6 +945,7 @@ namespace Mem {
 				auto only_physical = Config::getB("only_physical");
 				auto zfs_hide_datasets = Config::getB("zfs_hide_datasets");
 				auto& disks = mem.disks;
+				static unordered_flat_map<string, future<pair<disk_info, int>>> disks_stats_promises;
 				ifstream diskread;
 
 				vector<string> filter;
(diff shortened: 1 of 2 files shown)


src/linux/btop_collect.cpp
```diff
@@ -1078,31 +1085,41 @@ namespace Mem {
...
+						auto promise_res = promises_it->second.get();
+						if(promise_res.second != -1){
+							ignore_list.push_back(mountpoint);
+							Logger::warning("Failed to get disk/partition stats for mount \""+ mountpoint + "\" with statvfs error code: " + to_string(promise_res.second) + ". Ignoring...");
 							it = disks.erase(it);
-						else
-							it++;
+							continue;
+						}
+						disk = promise_res.first;
```
aristocratos, line 1108: `disk = promise_res.first;` will write over any IO stats collected later in this same section.
Since we are passing a copy of the `disk_info` struct in to the lambda.

Needs to be changed to update `disk.total`, `disk.free`, etc. instead.

aristocratos (changes requested):
See requested change.
Otherwise nice work!

src/linux/btop_collect.cpp
crestfallnatwork: My bad. Didn't notice that we used the same struct to get IO stats as well. 
But even if we just update the fields of the disk_info struct, wouldn't this still cause problems as we try to access discs that may fail statvfs later?
Should I add something like a ready flag to indicate if the disk in question has passed statvfs? 
But this would cause disk IO stats to take at least 2 cycles to update.

aristocratos: > But even if we just update the fields of the disk_info struct, wouldn't this still cause problems as we try to access discs that may fail statvfs later?
Should I add something like a ready flag to indicate if the disk in question has passed statvfs?

The `access(disk.stat.c_str(), R_OK) != 0)` check at the start of the IO collection should be enough i think, since we're "just" reading files from `/sys/`, they either exists or don't, so shouldn't be any issue.

But maybe there should be a higher level ignore list, for disks that keeps getting added when detecting disks but continuously fails `statvfs`, so they get ignored for good after X amount of tries.

nobounce:
Please don't fix issues in a second commit.
You can squash both commits by running something like
```
git reset --soft HEAD^
git commit --all --amend --no-edit
git push -f
```

src/linux/btop_collect.cpp
crestfallnatwork: Point noted. The `is_ready` check is almost certainly useless.
The current ignore list is most probably sufficient if something fails `statvfs`, it is most likely to keep failing `statvfs`.

aristocratos (approved):

aristocratos:
Nice work @crestfallnatwork, merging in to main, hopefully this also solves #392 and possibly other issues where the cause of issue is unknown.

PR #636, opened 2023-09-29
Minor string initialization improvement

nobounce:
See on [quick-bench.com](https://quick-bench.com/q/FcdhectF09oBMWIYhvwzGKqHtB4)


aristocratos (approved):
Good idea! There are probably plenty more instances where lambda initialization could be used for some performance gains.

Issue #551, opened 2023-06-07
[REQUEST] ability to set update_ms as CLI flag


code-n-go:
As the title states, adding an ability to set the update_ms upon startup using a command-line flag.  I have it launch automatically in a devcontainer, and have to manually adjust that every time.  I think there would be value under other circumstances to be able to set this upon launch also.

nobounce:
Can you try [this branch](https://github.com/nobounce/btop/tree/redraw-frequency)?

code-n-go:
@nobounce ... I thought I commented on this; however, realized I did not when I checked up the status of this.  This worked great.  Any update on when we may be able to get this merged and implemented?  Highly appreciate you taking the time to do this!

PR #640, opened 2023-10-03
Refresh rate

imwints:
Closes: #551

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -112,7 +112,7 @@ namespace Global {
 }
 
 //* A simple argument parser
-void argumentParser(const int& argc, char **argv) {
+void argumentParser(const int argc, char **argv) {
 	for(int i = 1; i < argc; i++) {
 		const string argument = argv[i];
 		if (is_in(argument, "-h", "--help")) {
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -16,6 +16,7 @@ indent = tab
 tab-size = 4
 */
 
+#include <algorithm>
 #include <csignal>
 #include <clocale>
 #include <pthread.h>
@@ -51,6 +52,7 @@ tab-size = 4
 #include "btop_theme.hpp"
 #include "btop_draw.hpp"
 #include "btop_menu.hpp"
+#include "fmt/core.h"
 
 using std::atomic;
 using std::cout;
@@ -109,6 +111,7 @@ namespace Global {
 	bool arg_tty{};         // defaults to false
 	bool arg_low_color{};   // defaults to false
 	int arg_preset = -1;
+	int arg_update = 0;
 }
 
 //* A simple argument parser
@@ -117,7 +120,7 @@ void argumentParser(const int argc, char **argv) {
 		const string argument = argv[i];
 		if (is_in(argument, "-h", "--help")) {
 			fmt::println(
-					"usage: btop [-h] [-v] [-/+t] [-p <id>] [--utf-force] [--debug]\n\n"
+					"usage: btop [-h] [-v] [-/+t] [-p <id>] [-u <ms>] [--utf-force] [--debug]\n\n"
 					"optional arguments:\n"
 					"  -h, --help            show this help message and exit\n"
 					"  -v, --version         show version info and exit\n"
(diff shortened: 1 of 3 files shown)


src/btop.cpp
```diff
@@ -108,22 +110,24 @@ namespace Global {
...
-					"usage: btop [-h] [-v] [-/+t] [-p <id>] [--utf-force] [--debug]\n\n"
+					"usage: btop [-h] [-v] [-/+t] [-p <id>] [-r <ms>] [--utf-force] [--debug]\n\n"
 					"optional arguments:\n"
 					"  -h, --help            show this help message and exit\n"
 					"  -v, --version         show version info and exit\n"
 					"  -lc, --low-color      disable truecolor, converts 24-bit colors to 256-color\n"
 					"  -t, --tty_on          force (ON) tty mode, max 16 colors and tty friendly graph symbols\n"
 					"  +t, --tty_off         force (OFF) tty mode\n"
 					"  -p, --preset <id>     start with preset, integer value between 0-9\n"
+					"  -r, --refresh <ms>    set the refresh rate in milliseconds\n"
```
aristocratos, line 130: Would you mind changing `-r` to `-u` and `refresh rate` to `program update rate`, to keep it consistent between how it's named in the config?

aristocratos (changes requested):

src/btop.cpp
```diff
@@ -109,22 +111,24 @@ namespace Global {
...
 
 //* A simple argument parser
-void argumentParser(const int& argc, char **argv) {
+void argumentParser(const int argc, char **argv) {
 	for(int i = 1; i < argc; i++) {
 		const string argument = argv[i];
 		if (is_in(argument, "-h", "--help")) {
 			fmt::println(
-					"usage: btop [-h] [-v] [-/+t] [-p <id>] [--utf-force] [--debug]\n\n"
+					"usage: btop [-h] [-v] [-/+t] [-p <id>] [-r <ms>] [--utf-force] [--debug]\n\n"
```
aristocratos, line 123: `-u <ms>`

src/btop.cpp
```diff
@@ -109,22 +111,24 @@ namespace Global {
...
 					"optional arguments:\n"
 					"  -h, --help            show this help message and exit\n"
 					"  -v, --version         show version info and exit\n"
 					"  -lc, --low-color      disable truecolor, converts 24-bit colors to 256-color\n"
 					"  -t, --tty_on          force (ON) tty mode, max 16 colors and tty friendly graph symbols\n"
 					"  +t, --tty_off         force (OFF) tty mode\n"
 					"  -p, --preset <id>     start with preset, integer value between 0-9\n"
+					"  -u, --refresh <ms>    set the program update rate in milliseconds\n"
```
aristocratos, line 131: Should probably be `-u, --update <ms>` to be more consistent.

src/btop.cpp
```diff
@@ -159,6 +163,19 @@ void argumentParser(const int& argc, char **argv) {
...
+			if (++i >= argc) {
+				fmt::println("ERROR: Refresh option needs an argument");
+				exit(1);
+			}
+			const std::string value = argv[i];
+			if (isint(value)) {
+				Global::arg_refresh = std::clamp(std::stoi(value), 100, 99999);
+			} else {
+				fmt::println("ERROR: Invalid refresh rate");
+				exit(1);
```
aristocratos, line 176: refresh -> update

aristocratos (changes requested):

src/btop.cpp
```diff
@@ -159,6 +163,19 @@ void argumentParser(const int& argc, char **argv) {
...
 			}
 		}
+		else if (is_in(argument, "-u", "--refresh")) {
+			if (++i >= argc) {
+				fmt::println("ERROR: Refresh option needs an argument");
+				exit(1);
+			}
+			const std::string value = argv[i];
+			if (isint(value)) {
+				Global::arg_refresh = std::clamp(std::stoi(value), 100, 99999);
```
aristocratos, line 173: The current max value from config is `86400000`

See https://github.com/aristocratos/btop/blob/f6cab3d3bdc06bec91c5b3681fbb2f00418e17ef/src/btop_config.cpp#L492

imwints: Should we give this value a name?

aristocratos: Well it's equal to 1 day in ms, so could be written as `1000 * 60 * 60 * 24` to make it more clear.

imwints: I think the most readable approach would be to make an named constant like `constexpr int ONE_DAY_MILLIS = 1000 * 60 * 60 * 24;` in Config

aristocratos: Sure.

aristocratos (approved):

Issue #642, opened 2023-10-06
[BUG] Proc scrollbar doesn't get cleared when scrolling


DecklynKern:
**Describe the bug**

Previous scroll thumb position in Proc is not always cleared, causing the scrollbar to fill up with block characters. This seemingly only happens when the terminal is thin enough horizontally.

**To Reproduce**

Set the terminal to be just wider than btop's minimum for that font, have enough processes open to cause the scrollbar to appear, then scroll the Proc section down. You can keep progressively making the terminal wider to see that this eventually stops happening.

**Expected behavior**

There would only be one block character in the scrollbar at any given time.

**Screenshots**

Konsole:
![image](https://github.com/aristocratos/btop/assets/71661150/808e9446-b5bd-4fe7-bfef-fe2a1a520f62)

Alacritty:
![image](https://github.com/aristocratos/btop/assets/71661150/3380e6cc-87aa-4a2d-9236-baf86a0f65e8)

**Info**
 - btop++ version: 1.2.13
 - Binary: static binary from release
 - Architecture: x86_64
 - Platform: Linux
 - (Linux) Kernel: 6.5.5-arch1-1
 - Terminal used: Reproduced in both Konsole and Alacritty
 - Font used: Source Code Pro

DecklynKern:
I attempted to fix this myself. First PR so sorry if I did something wrong in the process.
#643

PR #643, opened 2023-10-06
Fix scrollbar not clearing sometimes.

DecklynKern:
Refresh the entire scrollbar on redraw.
For #642.


aristocratos (approved):
Great catch and great fix!
Thanks!

PR #645, opened 2023-10-14
Use unordered_dense hash map

imwints:
Closes: #614

Note: I added the `<utility>` header in some places.
When testing this PR against the standard implementation of unordered map some units failed to compile since they relied on transitive includes in `<robin_hood.h>` and now `<unordered_dense.h>` header, which is not optimal.


imwints:
@aristocratos 

Is this still relevant after the switch to std::unordered?

aristocratos:
@imwints 
> Is this still relevant after the switch to std::unordered?

No, since there no apparent performance impact since the change (even with the added `safeVal()` calls, I don't see much point in using a external map library.

PR #648, opened 2023-10-18
Increase fortification level

imwints:
Newer versions of GCC and Clang support _FORTIFY_SOURCE=3 which increases code safety by providing better size checks

For reference see
https://developers.redhat.com/articles/2023/02/06/how-improve-application-security-using-fortifysource3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ option(BTOP_LTO "Enable LTO" ON)
 option(BTOP_USE_MOLD "Use mold to link btop" OFF)
 option(BTOP_PEDANTIC "Enable a bunch of additional warnings" OFF)
 option(BTOP_WERROR "Compile with warnings as errors" OFF)
+option(BTOP_FORTIFY "Detect buffer overflows with _FORTIFY_SOURCE=3" ON)
 option(BTOP_GPU "Enable GPU support" ON)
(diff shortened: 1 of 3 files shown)


NexAdn:
How about just providing a default value and letting users/packagers override the default as desired? Then we can have a sane default for the standard build and allow for setting a custom fortification level e.g. for testing increased fortification levels or just enabling them for hardened builds.

imwints:
Can you make a suggestion how this would look like?

NexAdn:
For the Makefile, just put `-D_FORTIFY_SOURCE=3` (or whatever level you choose) into some overridable variable (similar to `OPTFLAGS`). For CMake, there is the `$<IF:` [generator expression](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressions), so an `option()` to enable/disable default fortification should be possible.

imwints:
I've put the flag behind a build flag. It is enabled by default and can be disabled, described in the readme.

Might be also worth considering that this is usually a type of flag distributions set or unset, like Alpine and Gentoo do with their `fortify.h` header and we shouldn't worry about it and maybe remove the flag on our side.

@NexAdn is this what you where thinking about?

@aristocratos what do you think?

NexAdn:
> @NexAdn is this what you where thinking about?

Looks good to me. Thanks!

aristocratos (approved):

PR #659, opened 2023-11-06
Add alternative key codes for Delete, Insert, Home, End

ivanp7:
Mentioned keys produce other codes on some terminals (like st). Therefore, they are not detected in btop when opened in such terminals.


aristocratos (approved):
Thanks for contributing!

PR #666, opened 2023-11-20
Enable macos clang

muneebmahmed:
# Notes

The `Makefile` tests for `clang` >= 16, however, Apple clang is versioned based on Xcode rather than LLVM.
Apple clang 15.0.0 is [based on LLVM 16.0.0](https://en.wikipedia.org/wiki/Xcode#Toolchain_versions), so should be compatible

# Testing
I tested `make` on macOS Ventura on Intel and Apple Silicon, and was able to run `bin/btop` on both

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,10 +59,14 @@ override CXX_VERSION_MAJOR := $(shell echo $(CXX_VERSION) | cut -d '.' -f 1)
 
 CLANG_WORKS = false
 GCC_WORKS = false
+MIN_CLANG_VERSION = 16
 
 #? Supported is Clang 16.0.0 and later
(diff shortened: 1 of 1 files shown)


aristocratos (changes requested):
This would then also need to check _which_ Clang is being used, since "regular" Clang 15 would fail to compile.

muneebmahmed:
Updated, and tested again. I used just a grep for `Apple`, as this is the first line of `clang --version`:
```
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
```
and an example non-Apple clang:
```
Homebrew clang version 17.0.6
```

aristocratos (approved):

Issue #663, opened 2023-11-15
[REQUEST] Switch to vmactions/freebsd-vm@v1


imwints:
The old v0 based on macos and virtualbox has had a pretty high boot failure rate in the last months.
The author has addressed this since, and switched to a ubuntu based workflow with qemu. All comments on this thread where positive that it works and is even faster. I'll get to it tomorrow or so.

PR #674, opened 2023-11-30
Provide FreeBSD static release binaries

imwints:
Bumps vmaction@freesdb-vm to version 1 which runs on Linux and doesn't hang all the time. Also uses clang for full static compilation

Closes: #663


aristocratos:
GCC11 and Make should be used to make the only dependent variable between the platforms the changed code.
If the continuous build for FreeBSD fails, it should be because a platform specific issue, not a specific compiler issue or missing configuration in the CMakeLists.txt file.

An idea however might be to create more builds using Clang for all platforms, since we are now also supporting Clang.

aristocratos (changes requested):
Comment above.

imwints:
Should I also bump gcc to the latest version?

PR #675, opened 2023-12-02
Add CMake workflow for all platforms, fix RPATH on FreeBSD, remove unused compile flag from OSX and format

imwints:
Separate CI files so workflows are not being re-run on changes in `Makefile` and vice versa.

### DIFF
--- a/.github/workflows/cmake-freebsd.yml
+++ b/.github/workflows/cmake-freebsd.yml
@@ -1,0 +1,40 @@
+name: FreeBSD CMake
+
+on:
+  push:
+    branches: main
+    tags-ignore: '*.*'
+    paths:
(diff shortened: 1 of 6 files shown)


imwints:
With this macOS would be also supported with CMake. Since the build instructions got reallllllllyy lengthy with make, cmake and three platforms etc., it might be an idea to add a `docs/Build.md` (or something similar) file and just link from the readme. There are probably not so many people out there who read that anyways

aristocratos (approved):
Great!

aristocratos:
>  it might be an idea to add a docs/Build.md (or something similar) file and just link from the readme.

With most being collapsed in `details` tags I don't think it's that bad.
But I'll keep it in mind when merging in Open BSD support, since that will add some more compile instructions.

Issue #676, opened 2023-12-04
[BUG] CPU box displays a "Q" as the name of the CPU


kekePower:
**Read the README.md and search for similar issues before posting a bug report!**

Any bug that can be solved by just reading the [prerequisites](https://github.com/aristocratos/btop#prerequisites) section of the README will likely be ignored.

**Describe the bug**

I have an older laptop and in the CPU cores/threads box, it only displays a Q as the CPU name.
The CPU in the laptop is ```Intel(R) Core(TM) i7 CPU       Q 840  @ 1.87GHz.```

```
$ cat /proc/cpuinfo
processor	: 7
vendor_id	: GenuineIntel
cpu family	: 6
model		: 30
model name	: Intel(R) Core(TM) i7 CPU       Q 840  @ 1.87GHz
stepping	: 5
microcode	: 0xa
cpu MHz		: 1867.000
cache size	: 8192 KB
physical id	: 0
siblings	: 8
core id		: 3
cpu cores	: 4
apicid		: 7
initial apicid	: 7
fpu		: yes
fpu_exception	: yes
cpuid level	: 11
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm sse4_1 sse4_2 popcnt lahf_lm ssbd ibrs ibpb stibp tpr_shadow flexpriority ept vpid dtherm ida vnmi flush_l1d
vmx flags	: vnmi preemption_timer invvpid ept_x_only flexpriority tsc_offset vtpr mtf vapic ept vpid
bugs		: cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit mmio_unknown
bogomips	: 3723.85
clflush size	: 64
cache_alignment	: 64
address sizes	: 36 bits physical, 48 bits virtual
power management:
```

**To Reproduce**

Install and use btop on a laptop with the same CPU.

**Expected behavior**

Would be nice to see the full name of the CPU and not just a Q.

**Screenshots**

[...]

PR #677, opened 2023-12-05
Fix printed model name for older Intel CPU

imwints:
Closes: #676


kekePower:
Thanks. Applied this patch on the laptop and it now displays the correct CPU name.

aristocratos (approved):

Issue #678, opened 2023-12-05
[BUG] Apple M CPU Generation not displayed


masiboss:
**Describe the bug**

The CPU name is displayed as Apple M, while on a Apple M2
i belive it is read here
https://github.com/aristocratos/btop/blob/0f8498fa4e68e3af36a50b1d9f97b931b41b0824/src/osx/btop_collect.cpp#L207-L212
however
`sysctl -a | grep machdep.cpu.brand_string`
returns
`machdep.cpu.brand_string: Apple M2`

**To Reproduce**

Install on a Apple Silicon Mac

**Expected behavior**

Displays Apple M2

**Screenshots**

![image](https://github.com/aristocratos/btop/assets/32394683/43d01aab-627b-4d91-a232-214f0683324a)

**Info (please complete the following information):**
 - btop++ version: 1.2.13
 - Architecture: arm64
 - Platform: OSX
 - (OSX/FreeBSD) Os release version: 14.1.1
 - Terminal used: iTerm2 3.4.22

masiboss:
i don't know c++ but i had to make a minor change to get it to compile

``` cpp
#include <iostream>
#include <string>

#include <sys/sysctl.h>

int main() {
	char buffer[1024];
	std::size_t size = sizeof buffer;
	sysctlbyname("machdep.cpu.brand_string", &buffer, &size, nullptr, 0);
	std::cout << std::string(buffer) << '\n';
}
```

`Apple M2`

masiboss:
ok i just launched btop to have another look and its actually an issue with the text length and different scaling
<img width="1002" alt="image" src="https://github.com/aristocratos/btop/assets/32394683/08b142d3-2bf7-4a58-88c7-007b2d541161">
<img width="1710" alt="image" src="https://github.com/aristocratos/btop/assets/32394683/bfd2f18a-264d-4be2-acff-88972c8b60c7">

aristocratos:
@masiboss 
The issue isn't in `Cpu::get_cpuName()`.
In most cases the current cpu frequency is detected and showed next to the cpu name.
When that happens the cpu name needs to be trimmed for both to fit.

To fix this the following line need to be edited:
https://github.com/aristocratos/btop/blob/0d357468b50a13b48868b4969b4f19000c90d5d6/src/btop_draw.cpp#L2066

And needs to check if cpu frequency is available, if not, the `-14` needs to be changed to a more appropriate size.

aristocratos:
[...]

PR #679, opened 2023-12-07
Fix Apple Silicon CPUs misprinted

masiboss:
resolves: #678

if a Ultra version chip is used the last character still gets cut off when the CPU frequency box is in 1 column mode


masiboss:
> @masiboss What's the complete line you get from `sysctlbyname("machdep.cpu.brand_string", &buffer, &size, nullptr, 0)`?
Apple M2
> 
> To get rid of `Apple` shouldn't it be enough to add it to the list further down in `get_cpuName()`:
> 
> ```c++
> for (const auto& replace : {"Processor", "CPU", "(R)", "(TM)", "Intel", "AMD", "Core"})
> ```
yeah thats simpler, missed that

aristocratos (approved):
Looks good, thanks for contributing!

PR #684, opened 2023-12-12
GH Action for a test build of the snap and issue template changes`

kz6fittycent:
- Added a GH action to perform a sanity check for the snap builds.
  - Added a small insert to delineate between the snap and other binaries.
  - Added `opengl` interface for GPU support - possibly.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -29,6 +29,7 @@ Any bug that can be solved by just reading the [prerequisites](https://github.co
 
 **Info (please complete the following information):**
  - btop++ version: `btop -v`
+   - If using snap: `snap info btop`
  - Binary: [self compiled or static binary from release]
  - (If compiled) Compiler and version:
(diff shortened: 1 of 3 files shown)


aristocratos:
> Added opengl interface for GPU support - possibly.

Doubt that would be enough.
GPU support relies on loading of dynamic libraries present from the display drivers.

Is that even possible through snap?

kz6fittycent:
> > Added opengl interface for GPU support - possibly.
> 
> Doubt that would be enough. GPU support relies on loading of dynamic libraries present from the display drivers.
> 
> Is that even possible through snap?

Not likely. That plug is known for being janky, but I've included it just in case. It's in the `edge` channel, so testing is still underway. I'm just keeping us synced, since it's been a while since I've done so.

aristocratos:
You might want to add some ignore tags to the action so it only runs when there are changes to shared or linux specific code.
See https://github.com/aristocratos/btop/blob/main/.github/workflows/continuous-build-linux.yml for example of ignores.

aristocratos (approved):

PR #685, opened 2023-12-14
Added a way to install btop on deepin

XXTX-TOP:
Added a way to install btop on deepin

You can install btop on deepin using
''' sudo apt install btop'''


imwints:
Thanks for contributing! I think the couple of examples are enough for user the guess that the package name is `btop` everywhere and just change the command from `dnf` or others to their distros command.

Issue #594, opened 2023-08-04
[REQUEST] Display Power Usage Stats


Sol33t303:
**Describe the solution you'd like**
I'd love for btop to also be able to show a power usage graph, possibly one that can show volts, amps and watts used reported from the battery in one graph. I think this would work increadibly well in conjunction with the rest of the stats being provided, being able to view power usage as CPU, disk and network utilisation goes up and down would be very helpful. If UI realestate is at a premium even just a small "W: <wattage used>" besides the battery indicator would be helpful. I also wander if it would be possible to list energy use by each process like in Android? Although I think that might be a bit far fetched (I know powertop estimates this, maybe it might be possible to look there for inspiration?).

**Describe alternatives you've considered**
None, I am not aware of any other tools that graph power usage on linux off the top of my head, and certainly none that provide other system stats at the same time.

**Additional context**
I think it would be best to leave the feature turned off by default, as it will primarily only be used by laptop users (although I do like that btop actually detects the battery in my bluetooth controller on my desktop), I think battery information is an important thing to include in any general system monitors as it is an integral part of the system.

YuriiShkrobut:
Created PR with this feature, only Intel supported for now (have no AMD CPU to test).
https://github.com/aristocratos/btop/pull/595

vsey:
Hi @YuriiShkrobut,

I've looked at your changes in PR #595 where you use `/sys/class/powercap/intel-rapl:0/energy_uj` to calculate CPU power. I'd like to suggest an improvement that might broaden its applicability.

My suggestion is to add a power consumption indicator for the system battery to the battery indicator on laptops.
This would have the advantage of being an indicator of the energy consumption of the whole system, another advantage is that it also works for AMD systems, the disadvantage would be that it only works for laptops. 
Therefore, my suggestion is to add a power consumption to the battery section for laptops with the following metrics

- Battery current: `/sys/class/power_supply/BAT0/current_now`.
- Battery voltage: `/sys/class/power_supply/BAT0/voltage_now`.
- Battery current: `battery current * battery voltage`.

A small caveat to this method is that it is only effective when the laptop is running on battery power. When the laptop is plugged in, these readings reflect the power going to the battery, not the actual power being used by the system. 

I think adding this metrics to the battery indicator could be useful for laptops users. I look forward to hearing your thoughts on this!

Best,
vsey

PR #689, opened 2023-12-15
Add battery power draw for linux and freebsd

vsey:
I added an indicator for the charge/discharge power of the battery next to the predicted battery duration. As was described by @Sol33t303 in issue #594.  Because I don't own a FreeBSD laptop, I'm not sure if it works as expected on FreeBSD.

btop-power


aristocratos (approved):
Nice, thanks for contributing!

PR #690, opened 2023-12-15
conditional compile on Big Sur and up

joske:
Conditional compile the IOKit stuff only on macOS Big Sur and above.


aeiouaeiouaeiouaeiouaeiouaeiou:
Can we use `101504` instead of `__MAC_10_15_4`? The definition of this string is missing in older versions of the macOS SDK.

joske:
Yes, that makes sense!

aristocratos (approved):
Nice work!

aristocratos:
Should this be merged or is there more changes coming and this should be kept open?

joske:
I think this can go in as is. I don't have anything else for the moment.

Issue #698, opened 2023-12-25
[BUG] GPU monitoring not working on nobara


vsey:
**Describe the bug**
I compiled btop with GPU support, but the gpu stats aren't visible, because  btop cant find libnvidia-ml.so 

[A clear and concise description of what the bug is.]
The reason for this that the libnvidia-ml.so library is called libnnvidia-ml.so.1 or libnvidia-ml.so.535.104.05 on my system

**To Reproduce**
Compile btop with gpu support on nobara

**Info (please complete the following information):**
 - btop++ version: 1.3.0
 - Binary: self compiled
 - (If compiled) Compiler and version: Red Hat 13.2.1-1
 - Architecture: x86_64
 - Platform: Linux Nobara
 - (Linux) Kernel:  6.5.6-200
 - Terminal used: kitty
 - GPU: gtx 1070
 - gpu driver: 535.104.05

**Additional context**
I recompile it with the only difference being the change of the library name from libnvidia-ml.so to libnvidia-ml.so.1 in
`src/linux/btop_collect.cpp` and it starts working. I made a pull request with the change, but I don't know if this works in general, so I marked it as a draft for now, my idea was otherwise to replace ` libnvidia-ml.so` with a glob something like this ` libnvidia-ml.so.*` to catch all driver versions

Contents of `~/.config/btop/btop.log`:
2023/12/25 (19:05:11) | ===> btop++ v.1.3.0
2023/12/25 (19:05:11) | DEBUG: Starting in DEBUG mode!
2023/12/25 (19:05:11) | INFO: Logger set to DEBUG
2023/12/25 (19:05:11) | DEBUG: Using locale en_IE.utf8
2023/12/25 (19:05:11) | INFO: Running on /dev/pts/1
2023/12/25 (19:05:11) | INFO: Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: libnvidia-ml.so: cannot open shared object file: No such file or directory
2023/12/25 (19:05:11) | DEBUG: Failed to load librocm_smi64.so, AMD GPUs will not be detected: librocm_smi64.so: cannot open shared object file: No such file or directory
2023/12/25 (19:05:11) | DEBUG: Shared::init() : Initialized.
2023/12/25 (19:05:11) | DEBUG: Loading theme file: /home/peter/.config/btop/themes/catppuccin_mocha.theme

aristocratos:
@vsey 
[...]

PR #699, opened 2023-12-25
Fix GPU monotoring for nvidia on nobara

vsey:
Fixed GPU monitoring on nobara descried in #698

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -946,7 +946,7 @@ namespace Gpu {
 			if (initialized) return false;
 
 			//? Dynamic loading & linking
-			nvml_dl_handle = dlopen("libnvidia-ml.so", RTLD_LAZY);
+			nvml_dl_handle = dlopen("libnvidia-ml.so.1", RTLD_LAZY);
 			if (!nvml_dl_handle) {
 				Logger::info(std::string("Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: ") + dlerror());
 				return false;


aristocratos:
Not sure what the thinking is here?

Should the next person whose `libnvidia-ml.so` has an alternative name create a PR which replaces this one?

This would also break it for everyone which has the regularly named `libnvidia-ml.so`.

I will add some possible alternative names to look for during init...

zenofile:
Please also add `librocm_smi64.so.5` since Fedora (and Nobara, Debian, ...) don't package unversioned shared libraries for non-development packages (https://docs.fedoraproject.org/en-US/packaging-guidelines/#_devel_packages).

I was carrying this workaround around for Fedora and Debian, but keeping something like this updated is probably annoying.

<details>
<summary>Patch</summary>

```patch
diff --git a/src/linux/btop_collect.cpp b/src/linux/btop_collect.cpp
index 33a0be8..95b9f24 100644
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -946,16 +946,26 @@ namespace Gpu {
 			if (initialized) return false;
 
 			//? Dynamic loading & linking
-			nvml_dl_handle = dlopen("libnvidia-ml.so", RTLD_LAZY);
+			const array sonames = {
+				"libnvidia-ml.so",
+				"libnvidia-ml.so.1",
+			};
+
+			for (const auto& lib : sonames) {
+				nvml_dl_handle = dlopen(lib, RTLD_LAZY);
+				if (nvml_dl_handle != nullptr) {
+					break;
+				}
+			}
 			if (!nvml_dl_handle) {
-				Logger::info(std::string("Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: ") + dlerror());
+				Logger::info("Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: "s + dlerror());
 				return false;
 			}
 
 			auto load_nvml_sym = [&](const char sym_name[]) {
 				auto sym = dlsym(nvml_dl_handle, sym_name);
 				auto err = dlerror();
-				if (err != NULL) {
+				if (err != nullptr) {
 					Logger::error(string("NVML: Couldn't find function ") + sym_name + ": " + err);
 					return (void*)nullptr;
 				} else return sym;
@@ -1211,19 +1221,29 @@ namespace Gpu {
 
 			//? Dynamic loading & linking
 		#if !defined(RSMI_STATIC)
-			rsmi_dl_handle = dlopen("/opt/rocm/lib/librocm_smi64.so", RTLD_LAZY); // first try /lib and /usr/lib, then /opt/rocm/lib if that fails
-			if (dlerror() != NULL) {
-				rsmi_dl_handle = dlopen("librocm_smi64.so", RTLD_LAZY);
+			const array sonames = {
+				"/opt/rocm/lib/librocm_smi64.so",
+				"librocm_smi64.so",
+				"librocm_smi64.so.5", // fedora
+				"librocm_smi64.so.1.0", // debian
+			};
+
+			for (const auto& lib : sonames) {
+				rsmi_dl_handle = dlopen(lib, RTLD_LAZY);
+				if (rsmi_dl_handle != nullptr) {
+					break;
+				}
 			}
+
 			if (!rsmi_dl_handle) {
-				Logger::debug(std::string("Failed to load librocm_smi64.so, AMD GPUs will not be detected: ") + dlerror());
+				Logger::info("Failed to load librocm_smi64.so, AMD GPUs will not be detected: "s + dlerror());
 				return false;
 			}
 
 			auto load_rsmi_sym = [&](const char sym_name[]) {
 				auto sym = dlsym(rsmi_dl_handle, sym_name);
 				auto err = dlerror();
-				if (err != NULL) {
+				if (err != nullptr) {
 					Logger::error(string("ROCm SMI: Couldn't find function ") + sym_name + ": " + err);
 					return (void*)nullptr;
 				} else return sym;
```

</details>

aristocratos:
Added check for alternative library names (from @zenofile's suggestion) in commit ad14554f32cc9c19f3cdc8e28ea6012a974a533a

PR #700, opened 2023-12-26
Fix the GPU Monetoring for nvidia

vsey:
I removed the old implementation of loading `libnvidia-ml.so`, because otherwise a false negative is returned, when `libnvidia-ml.so` is not found before looking for alternative names.

PS: I'm not sure but in my opinion changing the name to `libnvidia-ml.so.1` would also fix the issue, because on all distros i had a look add (Debian, Ubuntu, Arch) both `libnvidia-ml.so.1` and `libnvidia-ml.so` are present and only on fedora only `libnvidia-ml.so.1` is present

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -948,13 +948,6 @@ namespace Gpu {
 		bool init() {
 			if (initialized) return false;
 
-			//? Dynamic loading & linking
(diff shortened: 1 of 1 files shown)


imwints:
@vsey
This issue was addressed in ad14554. You should try to compile a binary from git.

Issue #362, opened 2022-06-13
[QUESTION] CMake build support


xvitaly:
We want to contribute full CMake building and installation support. Will you merge such a PR when it's ready?

Issue #330, opened 2022-05-07
[REQUEST] Cross compilation support


jaskij:
**Is your feature request related to a problem? Please describe.**
I want to cross-compile btop for AArch64, using Yocto project. A quick look through your `Makefile` tells me the experience will not be pleasent - overriding `CXX` is a giant red flag in my book.

**Describe the solution you'd like**
Best would be either fixing the `Makefile`, or adding CMake support. I know CMake much better than make, so I'm prepared to contribute a `CMakeLists.txt` if you help me understand what's going on in your Makefile. Changing `:=` to `?=` [here](https://github.com/aristocratos/btop/blob/841941afbac20e9379375b300d54b0039d715cec/Makefile#L57-L63) would be a first step, but I'm not sure it would be sufficient.

**Describe alternatives you've considered**
I could keep a private, patched fork. But that's a stupid solution.

**Additional context**
The two approaches I proposed - fixing Makefile or adding CMake support - are only two possibilities, I'm sure there are other options. I want to help with this, and this issue is mostly to open discussion.

jaskij:
I guess my lack of GNU Make experience shows - I can probably set `CXX` as a make argument.

aristocratos:
@jaskij 
The makefile respects the CXX flag, `CXX := ` only applies if make is called with no `CXX=` flag.
You need to use `override` before the assignment in the makefile to ignore the make flags.
To set it when calling make just use `make CXX=<your cross compiler>`.
The CXX flag must be explicitly set when calling make to not automatically get auto set to best fitting GCC candidate.
This is because GCC is currently the only guaranteed compiler to work with the C++20 functions used in btop++.

I set it up like this to not have to deal with bug reports from people having CXX set to `clang` in the environment and that didn't bother reading the compile instructions.

There is also a cmake fork as mentioned in the README.md: 
> For a cmake based build alternative see the [fork](https://github.com/jan-guenter/btop/tree/main) by @jan-guenter

It looks to be a version behind, but the CmakeLists.txt should still work if copied over.

jaskij:
Okay, I must apologize - after reading the README and actually applying myself, it seems to build properly. I won't be able to test it on hardware until Monday.

That said, Yocto was throwing a `host-user-contaminated` QA at me (any specific reason for `cp -p`?):

[...]

PR #211, opened 2021-12-13
Add manpage

mariogrip:
This adds a manpage for btop

This is still *WIP* as it's still missing to be included in makefiles and is still missing some things like configuration file and keyboard navigation/bindings. but i put it up here for early review.

PR #715, opened 2024-01-14
WIP: Native Debian packaging

ottok:
> **WORK-IN-PROGRESS**: Do not merge

Hi!

I wanted to contribute to btop the native Debian packaging so that the project CI can build .deb packages on every commit and for example publish in a PPA, or just offer .deb package directly in the release artifacts. Additionallyo my goal is to make the absolute state-of-the-art build with all optimization flags, hardening flags, and have the build use the latest version of [fmtlib](https://tracker.debian.org/pkg/fmtlib) in Debian unstable instead of stale embedded copy from `include/fmt`. Being state-of-the-art includes that the packaging must pass 100% all tests in [Salsa-CI](https://salsa.debian.org/salsa-ci-team/pipeline/) and Lintian. 
The code base is small, so being perfect should be a reasonable amount of work.

I've already done most of the work, and I can split parts of this PR into separate submissions (e.g. the man page could be an independent PR), but there are a couple of things I am stuck on:

- **Link with system fmtlib**: Instead of the bundled `include/fmt` I'd like the build to use files from `/usr/include/fmt`. Can you advice how I should extend the Makefile to use `/usr/include/fmt` if it exist?

- **Verbose build**: I am trying to make the build as verbose as possible so that every single compiler/linker invocation prints out in the build log. However, no matter what I experiment with VERBOSE=true and QUIET=false the resulting log isn't verbose enough for [blhc](https://manpages.debian.org/unstable/blhc/blhc.1p.en.html). Attached build log: [btop_1.x.y_amd64.build.txt](https://github.com/aristocratos/btop/files/13930622/btop_1.x.y_amd64.build.txt). Am I getting the most verbose mode now or not? Could the Makefile perhaps be extended with some inline comments explaining the purpose of QUIET and VERBOSE? Also I don't understand why there is also a CMakeLists.txt for CMake builds. When is CMake better, why was it created? The inline comment simply states _CMake configuration for btop_, nothing else.

- **Hardening flags**: Related to above I am trying to debug what hardening flags are passed, but there are so many CFLAGS and LDFLAGS being overridden in the Makefile that I lost track. Is the Makefile intentionally written so that it will override most of the preset environment flags?
 Current state of [hardening-check](https://manpages.debian.org/unstable/devscripts/hardening-check.1.en.html) shows:
```
hardening-check --verbose bin/btop 
bin/btop:
 Position Independent Executable: yes
 Stack protected: yes
 Fortify Source functions: yes (some protected functions found)
	unprotected: memset
	unprotected: memcpy
	unprotected: wmemcpy
	unprotected: vsnprintf
	unprotected: memmove
	unprotected: gethostname
	unprotected: wmemset
	unprotected: read
	protected: memcpy
	protected: wmemcpy
	protected: fprintf
	protected: vsnprintf
 Read-only relocations: yes
 Immediate binding: yes
 Stack clash protection: yes
 Control flow integrity: no, not found!
```

- **Test suite**: Does btop have any tests that could run automatically after the build, or be triggered manually with say `make test` (none such target exists currently)? I can easily but btop in a build array that builds and runs it on 10 architectures, but I would need to have some test command to verify after the build that it actually runs.


imwints:
Completely agree with the comment above 👍🏽 

Regarding the man pages: there's already an [open pull request](#211). It is stale for a while now, maybe we can merge the efforts. 
Although I would think about the necessity of a man page altogether. btop is not a library nor does it export any functions to the outside. There is little to no need for documentation like that. 99% of the users will simply call `btop` without any flags. The few options which are available are easily listed with `btop -h` and are more or less only needed for debugging or scripts. I guess the man page would be a nice-to-have but also overkill for a project of this size. 

#### Optimization

> Additionally my goal is to make the absolute state-of-the-art build with all optimization flags ...

Which optimization flags are you talking about? btop is built with `-O2` (or even `-O3` with CMake release) and LTO by default.

#### Hardening

> ... hardening flags ...

Which flags are you talking about? btop is built with `_FORTIFY_SOURCE _GLIBCXX_ASSERTIONS _LIBCPP_ENABLE_ASSERTIONS` and `-fstack-clash-protection -fcf-protection -fstack-protector`
by default. This is more than most other application provide. Compilers default to better hardening options in newer versions anyway.

#### FMT 

> ... and have the build use the latest version of [fmtlib](https://tracker.debian.org/pkg/fmtlib) in Debian unstable instead of stale embedded copy from include/fmt.

> Link with system fmtlib: Instead of the bundled include/fmt I'd like the build to use files from /usr/include/fmt. Can you advice how I should extend the Makefile to use /usr/include/fmt if it exist?
 
This was [discussed before](https://github.com/aristocratos/btop/issues/549#issuecomment-1600359568).
There is no need to use any system library since btop doesn't link against `libfmt`. It's a header-only build dependency. 

### Debian

#### Pipeline

> Being state-of-the-art includes that the packaging must pass 100% all tests in [Salsa-CI](https://salsa.debian.org/salsa-ci-team/pipeline/) and Lintian. 

I don't see a reason why btop would need to conform to any external pipeline or CI.
Again, this will tie this project to a distribution which may prevent moving forward with newer C++ features etc.. Especially with a something as outdated as Debian.

#### Verbose build

> I am trying to make the build as verbose as possible so that every single compiler/linker invocation prints out in the build log. However, no matter what I experiment with VERBOSE=true and QUIET=false the resulting log isn't verbose enough for [blhc](https://manpages.debian.org/unstable/blhc/blhc.1p.en.html).

This is all described in the [README](https://github.com/aristocratos/btop#compilation-linux).

Yes, `make VERBOSE=true` gets you what you want. If you want some cleaner output you could also use CMake: `cmake -B build -G Ninja && ninja -C build -v` (`-v` for verbose output)

This will get you something like this:

```
ninja: entering directory 'build'
[1/10] /usr/lib/llvm/18/bin/clang++ -DGPU_SUPPORT -D_FILE_OFFSET_BITS=64 -D_GLIBCXX_ASSERTIONS -D_LIBCPP_ENABLE_ASSERTIONS=1 -isystem ~/src/btop.git/main/include -g -std=c++20 -flto=thin -fcolor-diagnostics -Wall -Wextra -Wpedantic -ftree-vectorize -fstack-clash-protection -fstack-protector -fcf-protection -MD -MT CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o -MF CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o.d -o CMakeFiles/btop.dir/src/linux/btop_collect.cpp.o -c ~/src/btop.git/main/src/linux/btop_collect.cpp
[2/10] /usr/lib/llvm/18/bin/clang++ -DGPU_SUPPORT -D_FILE_OFFSET_BITS=64 -D_GLIBCXX_ASSERTIONS -D_LIBCPP_ENABLE_ASSERTIONS=1 -isystem ~/src/btop.git/main/include -g -std=c++20 -flto=thin -fcolor-diagnostics -Wall -Wextra -Wpedantic -ftree-vectorize -fstack-clash-protection -fstack-protector -fcf-protection -MD -MT CMakeFiles/btop.dir/src/btop_tools.cpp.o -MF CMakeFiles/btop.dir/src/btop_tools.cpp.o.d -o CMakeFiles/btop.dir/src/btop_tools.cpp.o -c ~/src/btop.git/main/src/btop_tools.cpp
[3/10] /usr/lib/llvm/18/bin/clang++ -DGPU_SUPPORT -D_FILE_OFFSET_BITS=64 -D_GLIBCXX_ASSERTIONS -D_LIBCPP_ENABLE_ASSERTIONS=1 -isystem ~/src/btop.git/main/include -g -std=c++20 -flto=thin -fcolor-diagnostics -Wall -Wextra -Wpedantic -ftree-vectorize -fstack-clash-protection -fstack-protector -fcf-protection -MD -MT CMakeFiles/btop.dir/src/btop_theme.cpp.o -MF CMakeFiles/btop.dir/src/btop_theme.cpp.o.d -o CMakeFiles/btop.dir/src/btop_theme.cpp.o -c ~/src/btop.git/main/src/btop_theme.cpp
...
``` 
If Debian wants a hardened package they can adjust their build flags. This is described in the [README](https://github.com/aristocratos/btop#compilation-linux)

There is no reason for btop to provide more by default. All those come with a performance penalty which is in some cases undesirable and since not all Distributions want that behavior we should keep this minimal and let packagers adjust as needed.

#### CMake

> Also I don't understand why there is also a CMakeLists.txt for CMake builds. When is CMake better, why was it created? The inline comment simply states CMake configuration for btop, nothing else.

There are #28 #324 #330 #362. All requests for CMake as a build system. It is not uncommon for project to provide multiple ways to build the package, mostly a combination of [meson](https://mesonbuild.com/) and [CMake](https://cmake.org/). 

The Makefile currently does a lot what a Makefile generator otherwise would do. This was historically autotools and now modern generators like CMake or meson do this as well. They are meant for configuring a build file and provide faster rebuilds with [ninja](https://ninja-build.org/).

The top level comment in the `CMakeLists.txt` is just for orientation. It is an alternative way to build btop, mostly because it's more flexible in what is does (generating makefile) and also easy to integrate in packaging.

#### Hardening flags (v2)

> Related to above I am trying to debug what hardening flags are passed, but there are so many CFLAGS and LDFLAGS being overridden in the Makefile that I lost track. Is the Makefile intentionally written so that it will override most of the preset environment flags?

As mentioned above.

No, it's not intended to override the environment. Again, see the [README](https://github.com/aristocratos/btop#compilation-linux) for details. 
`make info` will spit out all compilation flags.

#### Testing

> Test suite: Does btop have any tests that could run automatically after the build, or be triggered manually with say make test (none such target exists currently)? I can easily but btop in a build array that builds and runs it on 10 architectures, but I would need to have some test command to verify after the build that it actually runs.

No, btop currently does not provide tests.


------------------

I don't know why you added `lowdown` in every CI file. the `cmake-*` builds do not produce artifacts, they are only meant to very the build step.

And there is already a [debian package](https://packages.debian.org/bookworm/btop).

ottok:
Thanks for taking the time to write a long answer addressing all the points. Yes, I have read the README, the Makefile, the CMakeLists.txt and skimmed through most of the source code. The reason I still had questions about the how the build is intended to work and what is the role of CMake is that the documentation only describes partially how the system works, and lacks explanations on *why* things are as they are, thus making it hard to reason about how changes should be designed and prompting questions about the background and purpose of things.

I added `lowdown` to the CMake dependencies to get CI job `Linux CMake / cmake_build_on_linux` to pass. I also added it to the musl jobs but they won't pass as the musl Docker images are 2+ years old and run Alpine v3.14 which is too old to include `lowdown`, which is Alpine starting from v3.15.

imwints:
Okay I took that a little too offensive. It's just an alternative to make to build btop. I personally find CMake better tool for the job since it is designed for configuring a Makefile, where in the raw Makefile we jump through a lot of hoops currently and adding new functionality is mostly more involved (IMO).

> I added lowdown to the CMake dependencies to get CI job Linux CMake / cmake_build_on_linux to pass.

I don't quite understand? The build passed before? Since the CMake jobs are only to verify that the actual build works and _not_ to generate artifacts with e.g. man pages included this won't be necessary.

Regarding the musl images this needs some work probably but I don't know of any alternative currently. But `apk` allows you to pull from a newer repository, so you could try to install it with `apk add --no-cache --<mirror-flag>=<alpine-v3.15-repo-url> lowdown`. But this might blow up CI times if this needs to update a lot of dependencies.

PR #722, opened 2024-01-16
Add config reload support

### LINKED ISSUE #683: [REQUEST] Reload config from disk using signal

I want to automate switching between light and dark mode on all my applications. btop is ****ing awesome, so I constantly have it open on all my PCs.

Currently I don't see any *good* way of automatically switching between light and dark themes in btop without doing it manually through the UI.

It would be nice if I could send a signal (`SIGHUP` or `SIGUSR1`) to btop to make it reload the configuration from disk. That way I could write a `darkman` script that toggles the btop theme in the config file and runs `pkill -USR1 btop` and presto!

Wrapper script that restarts btop any time the config file changes. Not ideal, since all state is lost.

**Additional context**

I'm hopelessly addicted to home automation and the lack of this feature is causing me considerable anguish.

MartinPit:
Btop is now able to respond to `SIGUSR2` signal and apply any changes made to config file, which were made while it has been running. Also added a key bind to do so manually. 

This changes addresses my needs but also resolves #683.

- By supplying `SIGUSR2` signal the config is reloaded from disk
- Pressing `ctrl+r` in app also does the same
- Help menu extended with `ctrl+r` entry


imwints:
`SIGUSR1` is already used to interrupt the key polling (? something like that).

imwints:
That is not commented out, this is just a comment what the signal is used for.

We have a couple of places in the code which will send `SIGUSR1` to the process. This will now always lead to reloading of the config.

MartinPit:
Oh, my bad then. Would it be fine if instead it would be switched to SIGUSR2?

imwints:
I guess so. Have you thought about making this a key bind instead?

MartinPit:
Edited with suggested changes.

aristocratos (approved):

aristocratos:
Thanks for contributing!

PR #704, opened 2024-01-07
Add unresolved externs for rpi 32 build

ntc490:
Got btop building on my RPi 3B+ running Raspian Lite 32.

PR #725, opened 2024-01-18
cmake: link CMAKE_DL_LIBS

alalazo:
Without the explicit command below, I get a build failure when building from sources on Ubuntu 20.04 and adding GPU support.

The build was done with Spack, see https://github.com/spack/spack/pull/42139

If it is of any use, I attach a log of the failed build (before this change): [spack-build-out.txt](https://github.com/aristocratos/btop/files/13975792/spack-build-out.txt)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,9 @@ if(LINUX AND BTOP_GPU)
 
     target_link_libraries(btop ROCm)
   endif()
+  if(NOT BTOP_STATIC)
+    target_link_libraries(btop ${CMAKE_DL_LIBS})
+  endif()
 endif()
 
 if(BTOP_USE_MOLD)


imwints:
This is because `glibc` on Ubuntu 20.04 is too old. This is the same issue as [here](https://github.com/aristocratos/btop/pull/704). 

There are newer Ubuntu LTS's available with proper support for CMake, glibc and gcc. (20.04 doesn't have a new enough CMake anyway, and only gcc-10 while 11 or newer is recommended).

I'd like to keep this in line with the Makefile which doesn't support that either.

alalazo:
> There are newer Ubuntu LTS's available with proper support for CMake, glibc and gcc. (20.04 doesn't have a new enough CMake anyway, and only gcc-10 while 11 or newer is recommended).

I built the tool with a package manager that doesn't require root privileges. We can currently deal with `cmake` and `compiler` but we don't yet build our own `glibc`. So `glibc` remains our only issue for a pristine build.

Since I'll use that patch to build `btop`, I thought it would be good to upstream it. I think one point would be: would it make building with newer `glibc` worse? If not this seems an improvement :slightly_smiling_face:

alalazo:
> I'd like to keep this in line with the Makefile which doesn't support that either.

Does this mean that a similar change in the Makefile is welcome, or is it a policy to build only with the system provided toolchain on each OS?

imwints:
No, there is no such policy, but it's just unusual to run into a system with brand new compiler and outdated libc.

You don't even need to patch the build files. Just configure cmake with proper LDFLAGS like so `LDFLAGS=-ldl cmake -B ...` 
IMO this is not worth patching since it's so uncommon and LDFLAGS is even simpler.

haampie:
Hi @imwints,

Could you please merge this? You're explicitly using `dlopen`, so link to `libdl.so`.

Multiple people have run into this, and it should be the build system that gets it right, especially if it's just 3 lines.

antoine-morvan:
> it's just unusual to run into a system with brand new compiler and outdated libc.

What's unusual for you might be usual for someone else ;)

I agree with haampie, if there is an explicit dependency, then the build system should handle it explicitly (actually doesn't matter 3 lines or 3K ^^').

Issue #582, opened 2023-07-23
[REQUEST] Support vim keybindings for faster moving: C-d, C-u, C-f, C-b, {, }, [[, ]]


rootex-:
**Is your feature request related to a problem? Please describe.**
I found that commands like G and gg work as expected and it's a pleasure to be able to move at the top or at the bottom of the processes list. 
But I'm always frustrated when I'm unable to move faster between processes using standard vim keybindings: 
* jumping half page up / down 
* jumping page up / down
* jumping to the next process in the process tree

**Describe the solution you'd like**
I offer to mimic the general vim keybindings for moving faster:
* jumping half page up / down - Ctrl-u, Ctrl-d 
* jumping page up/ down - Ctrl-f, Ctrl-b
* jumping to the next process in the process tree - could be achieved using { and } or [[ and ]]
* jumping to the next node of the same level - could be achieved using { and } or [[ and ]]

**Describe alternatives you've considered**
None.

astyagun:
IMHO `j` and `k` to move down and up in the procs list would be nice too. I was surprised just now when I've tried using these and was suggested to kill a process 🙂

A4-Tacks:
Can I also add single line screen scrolling using `Ctrl+E/Y`? This will give a very smooth feeling

itpropro:
Was there a way added in the meantime to use vim keybindings? Every other terminal tool I use has it as default or has at least support for rebinds. It’s quite confusing in a tui to not move up and down with j and k.

deckstose:
@itpropro 4th row in the options menu

PR #726, opened 2024-01-19
Add ctrl+u/d/b/f vim keys (#582)

isak102:
Ctrl+u/d = fast scroll up/down
Ctrl+b/f = page up/down

These keys are very useful imo. They don't conflict with anything so maybe they should be enabled even without enabling vim mode? 

And maybe the jump size should be configurable when using ctrl+u/d, right now I hardcoded it to 10.


src/btop_draw.cpp
```diff
@@ -1517,6 +1517,21 @@ namespace Proc {
 			start = max(0, numpids - select_max);
 			if (selected > 0) selected = select_max;
 		}
+		else if (vim_keys and cmd_key == "ctrl+u") {
+			if (start > 0 and selected <= 10) {
+				start = max(0, start - 10);
+			}
+			else selected = max(0, selected - 10);
```
imwints, line 1524: Why doesn't 

```cpp
max(0, selected - 10);
```

work directly?

PR #734, opened 2024-01-23
Include metadata in binary

imwints:
This will include some metadata in the binary. The main motivation was to make bug reports a little easier for the author.

`-v` will now print the version and the git commit when the binary was build in a git tree.

```bash
# From source tarball
./btop -v
btop version: 1.3.0

# From git tree
./btop -v
btop version: 1.3.0+31ee115
```

`--version` is now more verbose than `-v`, it will also include information about the compiler, the build command and the enabled features, e.g. GPU support. This is then easily included by the author without without any digging around when they copy the output of `btop --version` as requested in the issue template.

```bash
# Makefile
btop version: 1.3.0+31ee115
Compiled with: x86_64-pc-linux-gnu-g++-14 (14.0.0)
Configured with: make STATIC= FORTIFY_SOURCE=true GPU_SUPPORT=true RSMI_STATIC=

# CMake
./btop --version
btop version: 1.3.0+31ee115
Compiled with: clang++ (17.0.6)
Configured with: cmake -G Ninja -DBTOP_STATIC=OFF -DBTOP_USE_MOLD=OFF -DBTOP_FORTIFY=ON -DBTOP_GPU=ON
```

Empty means default, I didn't want to default initialize the options just for prettier output as this would clutter the Makefile.


ottok:
Currently btop has a fully reproducible build (https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/btop.html). Be careful when embedding this type of things that it will not change on every build, and that it also supports build out-of-git-tree.

imwints:
> Be careful when embedding this type of things that it will not change on every build

Change what on every build? There is no randomness in this PR. It is all deterministic.

> and that it also supports build out-of-git-tree.
>> ```bash
>> # From source tarball
>> ./btop -v
>> btop version: 1.3.0
>>
>> # From git tree
>> ./btop -v
>> btop version: 1.3.0+31ee115
>> ```

aristocratos (approved):
Nice! 👌

PR #747, opened 2024-02-02
fix: don't mangle memory for zombie processes

joske:
On macOS, for a zombie process, `proc_pidpath` returns 0, and nothing is written in fullname, so it's uninitialized garbage. We should not try to do any string mangling on it.

There is no program name, and `ps -ef` returns `<defunct>` for the zombie, so I'm doing the same for btop.


joske:
You can easily test this with the following little program that creates a zombie process:
```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

// This program creates a zombie process, the parent never calls wait()
int main() {
  int pid = fork();
  if (pid == 0) {
    exit(0);
  } else {
    printf("Child process: %d\n", pid);
    for (;;) {
      sleep(30);
    }
  }
  return 0;
}
```

aristocratos (approved):

PR #761, opened 2024-02-12
fix: Can't detect librocm 6.0.x

imwints:
librocm reports it's version as 7.0.0.0 in the 6.0.x release series.


aristocratos:
@imwints 
Well, that's not confusing at all...

But are you sure the minor version won't change until version 7?

Wouldn't it make sense that version 7 would be 8.0.0 internally?

imwints:
Why is that not confusing? The 5.x releases all had the major version 5, I'd expect a 6.x release to have the major version 6...

If you build rocm from git, the major version is 2 for some reason, which confuses me even more.

Also I probably didn't mean internal but rather what the library reports.

aristocratos:
> Why is that not confusing?

That was irony...

But I was asking why you are checking `version.minor == 0`, we are only concerned with a major release change, since AMD *hopefully* won't do any breaking changes until next major release.

And by the current logic that would be 8.0.0.0 (or 9.0.0.0 if they are skipping one again).

imwints:
:) went over my head

> And by the current logic that would be 8.0.0.0 (or 9.0.0.0 if they are skipping one again).

I just thought if they change their scheme again and release a 7.x with the version 7.1.0.0 or similar this is save for now and would only accept the real 6.x for now. I just wanted to limit the allowed version range so this just wouldn't allow any new release. But I'm happy to remove that

aristocratos:
Hm, there is also a `rsmi_version_str_get()` function that returns a string.

Descriptions:

> rsmi_version_get(): Get the build version information for the currently running build of RSMI

> rsmi_version_str_get(): Get the driver version string for the current system.

Might be an idea to test `rsmi_version_str_get()` instead and check if it's *correct* then if so check if first char is 5 or 6.

@imwints Do you have a AMD card to test with?

The definition is at: https://rocm.docs.amd.com/projects/rocm_smi_lib/en/docs-6.0.0/.doxygen/docBin/html/group__VersQuer.html#ga6d4f034afcef2e50b4e8392c5dfd7a57

imwints:
This will just give you the output of `uname -r`

alerque:
c.f. [Upstream issue](https://github.com/ROCm/rocm_smi_lib/issues/157) filed bi @imwints.

aristocratos:
> This will just give you the output of `uname -r`

Huh, `rsmi_version_str_get()` gives you the output of `uname -r`?

That makes even less sense to me...

imwints:
[Here is the implementation](https://github.com/ROCm/rocm_smi_lib/blob/5d2cd0c2715ae45b8f9cfe1e777c6c2cd52fb601/src/rocm_smi.cc#L3693-L3739)

aristocratos:
Will merge and create a new release later today, since I'm guessing it's unlikely anything will come from the issue in the ROCm repository in any reasonable time.

aristocratos (dismissed):

aristocratos (approved):

PR #763, opened 2024-02-12
cmake: Remove version

imwints:
This version variable is currently not used but could you still bump the version on a new release? Or we could just remove it for now.


aristocratos:
Probably better to remove it if it's not used.

Issue #767, opened 2024-02-13
Don't use all CPU cores by default during build


lcheylus:
During build/compilation with `make`, all CPU cores are used by default.

See definition of `THREADS` variable in `Makefile` and `MAKEFLAGS := --jobs=$(THREADS)` (line 154).

**Expected behavior**

Compilation with `make` must respect `--jobs` flag and **doesn't use ALL CPU cores by default**.

**Info (please complete the following information):**
 - btop++ version: 1.3.2
 - Binary: self compiled
 - Compiler and version: OpenBSD clang version 16.0.6
 - Architecture: amd64 (x86_64)
 - Platform: OpenBSD

aristocratos:
> Compilation with make must respect --jobs flag [..]

It does...

> [..] and doesn't use ALL CPU cores by default.

Why?

markg85:
You're doing a poor job explaining why you want this @lcheylus!

I do think i get your idea but that is just me assuming.
Build flags, like core counts too, are **sometimes** defined system wide. Like for example if i'm using archlinux to make a package for arch then it uses the flags set by the distribution or changed by the user. So in that specific case i can see why setting the core count in the makefile could be annoying.

However.

When you compile you **definitely should** be using all cores!

Please do explain what you mean as i'm very curious about your reasoning.

lcheylus:
Invalid => closed

markg85:
You give me a thumbs down, don't explain anything at all and just close it. Arrogance.

PR #768, opened 2024-02-13
build: don't use all CPU cores by default

lcheylus:
Fix aristocratos/btop#767


lcheylus:
Bug #767 => PR not merged/closed

Issue #769, opened 2024-02-13
[Build] Add a "really" quiet mode in Makefile


lcheylus:
During build via GNU `make` (using `Makefile`), I'm using "Quiet" mode via `QUIET=true` (especially to integrate btop build in automatic packaging toolchain).

The "Quiet" build is not REALLY quiet:

- initial header with version / OS / arch
- display of build steps with colors
- display of build steps with % and time used
- total time to complete build

**To Reproduce**

Compilation of `btop` with `QUIET=true VERBOSE=true make`

**Expected behavior**

I would like to have a *REALLY quiet mode** for build/compilation: 
- no initial header with version / OS / arch
- no display of build steps with colors
- no display of build steps with % and time used
- no display of total time to complete build

PR #772, opened 2024-02-14
build: add a really quiet mode for compilation

lcheylus:
With QUIET=true, build with make is really quiet:
- no initial header with version / OS / arch
- no display of build steps with colors
- no display of build steps with % and time used
- no display of total time to complete build
- with VERBOSE=true, only logs from compiler

Fix aristocratos/btop#769

---
Build with GNU make and tests OK on OpenBSD current/amd64 and Linux Debian/testing with:
- clang++ v16
- g++ v11

Only logs from compiler displayed during build with `QUIET=true VERBOSE=true make`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,7 @@ endif
 
 
 info-quiet: | info rocm_smi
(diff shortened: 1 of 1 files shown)


aristocratos:
> add a really quiet mode for compilation

This doesn't add a really quiet mode, it changes the current quiet mode.

Issue #387, opened 2022-07-08
[REQUEST] Regular expression to filter process


ghz-max:
**Describe the solution you'd like**
A way to filter process using regex.

**Additional context**
I'm used to have the feature with glances and find it very handy at times
![image](https://user-images.githubusercontent.com/48886505/178011916-6b673ec9-81c9-4ed5-b74a-a29299c80011.png)

Erotemic:
I would also like this. I had written up an issue before I thought to check if one existed. But I think my text has more detail: 

### Problem

htop lets me filter items in the process list with F4, so even though I like btop more, I often have to fall back to htop when I need to inspect subsets of processes. 

### Desired Solution

Add the ability to filter the process list by name, perhaps using a regular expression. Btop could implement some hotkey that allows the user to input a regular expression. Using the return key accepts the imputed text and closes the dialog. Other UIs could work. Regardless once btop has been given some "filter text", it uses that information to hide items that don't match it from the process list. That could be via regex, or plain text (preferably regex or maybe glob pattern matching).

aristocratos:
@Erotemic 
From README.MD:
![image](https://user-images.githubusercontent.com/59659483/197408508-b01d3a72-c573-4bbb-8610-7abcb784be71.png)

From the help menu:
![image](https://user-images.githubusercontent.com/59659483/197408599-06040ab1-d16b-4c52-8173-d238a3995ea6.png)

From the main display of btop:
![image](https://user-images.githubusercontent.com/59659483/197408663-8dc47990-6f0f-4895-95d3-2ca3b0b56790.png)

Erotemic:
Well, now I feel silly. Thank you :)

kanyck:
+1
At least simple OR statement "|" would be very helpful.

imwints:
For searching by field type, see https://github.com/aristocratos/btop/issues/980.

PR #806, opened 2024-03-20
Add regex filtering

imwints:
Filters starting with '!' will try to match processes pid, name, cmd and user with extended regex as defined by the C++ standard. A single '!' will not filter.

Partially implements https://github.com/aristocratos/btop/issues/387, but is missing the option to filter by type, meaning by user or command only.

### DIFF
--- a/src/btop_shared.cpp
+++ b/src/btop_shared.cpp
@@ -17,6 +17,8 @@ tab-size = 4
 */
 
 #include <ranges>
+#include <regex>
+#include <string>
 
 #include "btop_config.hpp"
 #include "btop_shared.hpp"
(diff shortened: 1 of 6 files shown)


mrlee14 (approved):
It wrorks!

imwints:
@aristocratos 

If it's OK I would like to implement the search by type in another PR.

aristocratos:
@imwints Looks good! But should do some performance testing on how the filtering behaves before and after.

Should likely be faster, but you never know.

aristocratos:
> But since the regex matching is only performed with the leading '!' [...]

Ah, right :)
Well the extra branch would slow it down, but nothing noticeable.

Nice work!

aristocratos (approved):

imwints:
@aristocratos 

[Here](https://quick-bench.com/q/EFvwvi63nU-rvJRwmwlADN9IduE) is just a simple comparison without any context...

aristocratos:
@imwints 
Ouch!
Doesn't seem to be any faster even if the regex is static...
But on the other hand, this isn't anything that will be noticeable in human time scales.

yogasw:
how use this feature, I try !chrome but the result does not show all processes other than chrome

PR #819, opened 2024-04-12
fix pwr_usage not being defined correctly during rsmi collection

kalkafox:
This commit fixes an issue where pwr_usage under the rsmi collection stack is not properly defined, resulting in the AMD GPU power usage status reporting as 0W in btop.


aristocratos (approved):

PR #838, opened 2024-05-02
Add missing linker flag for static links on FreeBSD

imwints:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278551


aristocratos:
@imwints 
I'm fairly certain I remember there was issues compiling with `-static` on freebsd.
Have you tried compiling with both gcc and clang and also running the produced binary?

imwints:
@aristocratos 
yes, that is why I filed the upstream bug.

tried both and they work now, freebsd 14.0 clang 18 gcc 13.2

aristocratos:
Alright, missed the link :)

Great 👍

aristocratos (approved):

Issue #375, opened 2022-06-26
[REQUEST] Display all BTRFS subvolumes as one partition (i.e. Don't treat BTRFS subvolumes as individual partitions)


KSXGitHub:
**Is your feature request related to a problem? Please describe.**

Btop has a section that monitor mounted partitions. Below is the Btop UI displaying my filesystem:

![Screenshot of Disks in Btop](https://user-images.githubusercontent.com/11488886/175800950-4b628d47-d7b6-422b-80ee-9af30b7ceb8c.png)

In the above screenshot, `root`, `home`, `cache`, and `log` are all subvolumes of the same BTRFS partition. They all have the exact same statistics (same "Used", same "Free", same "Total", and same "IO%"). Therefore, it is redundant to display all of them in different rows.

**Describe the solution you'd like**

Show only a single row for all subvolumes. The name of the row could either be the name of the partition, or a list of subvolume names (e.g. `root,home,cache,log`), or just the name of the most important subvolume (e.g. `root`). 

**Describe alternatives you've considered**

**Additional context**

My `fstab`:

```
# /etc/fstab: static file system information.
#
# Use 'blkid' to print the universally unique identifier for a device; this may
# be used with UUID= as a more robust way to name devices that works even if
# disks are added and removed. See fstab(5).
#
# <file system>                                            <mount point>         <type>   <options>                                                                                          <dump> <pass>
UUID=F832-EB0A                                             /boot                 vfat     defaults,noatime                                                                                       0      2
UUID=a036be94-037c-4bc5-a79c-2748298843dd                  /                     btrfs    subvol=/@,defaults,noatime,space_cache,autodefrag,x-gvfs-icon=archlinux,x-gvfs-symbolic-icon=archlinux 0      1
UUID=a036be94-037c-4bc5-a79c-2748298843dd                  /home                 btrfs    subvol=/@home,defaults,noatime,space_cache,autodefrag                                                  0      2
UUID=a036be94-037c-4bc5-a79c-2748298843dd                  /var/cache            btrfs    subvol=/@cache,defaults,noatime,space_cache,autodefrag                                                 0      2
UUID=a036be94-037c-4bc5-a79c-2748298843dd                  /var/log              btrfs    subvol=/@log,defaults,noatime,space_cache,autodefrag                                                   0      2

[...]

PR #869, opened 2024-06-02
Display btrfs subvolume as one partition

koitococo:
issue #375 

Skip some BTRFS subvolumes if other subvolume  in same partition is added, to avoid messing up the Disks view

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -1837,6 +1837,21 @@ namespace Mem {
 						size_t zfs_dataset_name_start = 0;
 						if (fstype == "zfs" && (zfs_dataset_name_start = dev.find('/')) != std::string::npos && zfs_hide_datasets) continue;
 
+						//? skip BtrFS subvolumes
+						if (fstype == "btrfs") {
+							bool other_subvol_found = false;
+							string devname = fs::canonical(dev, ec).filename();
+							for (auto& [mp, info] : disks) {
+								if (info.dev.filename() == devname and mp != mountpoint) {
+									other_subvol_found = true;
+									break;
+								}
+							}
+							if (other_subvol_found) {
+								continue;
+							}
+						}
+
 						if ((not use_fstab and not only_physical)
 						or (use_fstab and v_contains(fstab, mountpoint))
 						or (not use_fstab and only_physical and v_contains(fstypes, fstype))) {


Krutonium:
I would love to see this merged. Also, LGTM even though I have no authority hee.

Krutonium (approved):
I think this looks good. I don't see any bugs.

imwints:
Should this be a toggle-able option?

koitococo:
> Should this be a toggle-able option in the config? 

Yes, I added a option just now.

> And what name is displayed then? Is it the first volume's name or am I missing something?

The first volume's name is displayed, which follows the same order as /etc/mtab

imwints:
Does that make sense tho? It should display the disk name like `/dev/sda1`. I'd be confused if I see my home volume but not my root volume

bionade24:
>  It should display the disk name like `/dev/sda1`. I'd be confused if I see my home volume but not my root volume

`sda1` is can change on a (re)boot and is not static. In that case, you'd be even more confused. So with your idea it would have to be `PARTLABEL` or `PARTUUID` if none provided. Additionally, btop currently reports disk usage and I/O per partititon. That's correct for the 1st metric, but wrong for the second. I tried to approach both those things with the Btrfs subvol problem together 2 years ago or so by using util-linux APIs and the complexity compared to the prev. solution went overboard quickly.

imwints:
> I think 99% of the users are very much aware of it when they have 2 subvolumes for /home and / on one Btrfs.

I don't think this is true.

I always check the size of the drive and otherwise I'm screwed and have to check `df`. ^^

Also, a concern is that if you have multiple subvolumes in a weird order, like `/var, /home, /, /usr`, it's not __nice__ to have it displayed as `/var` where you'd expect it to be `/`. Also sometimes you don't create the subvolumes yourself, e.g. with any modern graphical installer or `systemd-homed`.

aristocratos:
I would agree with @imwints here.
If it's a container that is displayed, then the name should be the device/partition name, not one of the subvolumes.

> sda1 is can change on a (re)boot and is not static.

Why would it change? Unless the disk has been moved to another port on the motherboard, this would stay the same, no?

bionade24:
> Unless the disk has been moved to another port on the motherboard, this would stay the same, no?

No, udev doesn't make any guarantees on that and the device naming order can change on each boot. I have experienced this myself hundreds of times with `nvme0` and `nvme1` changing their device order myself. Additionally, if you plug another drive into your system, it gets assigned some random identifier and not the next one that hasn't been used before.  E.g. I have a Root disk `nvme0n1` and a data Disk `sda` and now add another data disk. The new disk may very well be assigned `sda`  on 99% of boots and the old disk now `sdb`. 

There is a reasons why Linux distros default to use filesystem UUIDs in `/etc/fstab` instead of sticking with `nvmeXn1` or `sdX`.  If it should show some label that gives usable information, it has the be either the filesystem label or the partition label.

bionade24:
> Also, a concern is that if you have multiple subvolumes in a weird order, like /var, /home, /, /usr, it's not intuitive to have it displayed as /var where you'd expect it to be /.

AFAIK you always have to mount your root filesystem first, so this specific scenario can never ever happen.

src/linux/btop_collect.cpp
```diff
@@ -1865,6 +1866,7 @@ namespace Mem {
 				auto use_fstab = Config::getB("use_fstab");
 				auto only_physical = Config::getB("only_physical");
 				auto zfs_hide_datasets = Config::getB("zfs_hide_datasets");
+				auto only_first_btrfs_subvol = Config::getB("only_first_btrfs_subvol");
```
imwints, line 1869:
```suggestion
				auto btrfs_group_subvolumes = Config::getB("btrfs_group_subvolumes");
```

maybe? This is more inline with the previous option for `zfs`.

src/linux/btop_collect.cpp
```diff
@@ -1943,6 +1946,12 @@ namespace Mem {
 						size_t zfs_dataset_name_start = 0;
 						if (fstype == "zfs" && (zfs_dataset_name_start = dev.find('/')) != std::string::npos && zfs_hide_datasets) continue;
 
+						//? skip BtrFS subvolumes
+						if (only_first_btrfs_subvol and fstype == "btrfs") {
+							string devname = fs::canonical(dev, ec).filename();
+							if (!found_btrfs_subvolumes.insert(devname).second) continue;
```
imwints, line 1952:
```suggestion
							if (!found_btrfs_subvolumes.insert(devname).second) {
							    continue;
							}
```

imwints (commented):
Please squash all commits into one:

```shell
git reset --soft <sha of first your first commit>
git commit --amend
```

bionade24:
> No, udev doesn't make any guarantees on that and the device naming order can change on each boot

I have observed this the last couple of months and now can say with absolute certainty that the device files (`nvmeXnY`) do change their assignment regularly, on average at least every 4th reboot. So displaying device file names would be very unreliable.

Krutonium:
> > No, udev doesn't make any guarantees on that and the device naming order can change on each boot
> 
> I have observed this the last couple of months and now can say with absolute certainty that the device files (`nvmeXnY`) do change their assignment regularly, on average at least every 4th reboot. So displaying device file names would be very unreliable.

Indeed. For anyone interested, it largely comes down to the order in which the devices show up to the system, which can change from any number of things, but is usually just run to run variance if you have two nearly identical solid state devices. If you had HDD's, chances are they show up in order of fastest to slowest to spin up.

zeerayne (approved):

PR #873, opened 2024-06-08
Incorrect work with CMake version less than 3.25

AnatoliiShablov:
cmake_minimum_required version is setup to 3.24. But relying on documentation LINUX variable was introduced in 3.25
[https://cmake.org/cmake/help/latest/variable/LINUX.html#linux](https://cmake.org/cmake/help/latest/variable/LINUX.html#linux)

On my system I get such error
```
CMake Error at CMakeLists.txt:73 (message):
    Linux is not supported
```


imwints:
I think it's better to bump the required CMake version. Debian stable is on 3.25 and we use the `LINUX` variable in a more places.

imwints:
Closed-by: https://github.com/aristocratos/btop/pull/1092

Issue #644, opened 2023-10-14
Add Intel GPUs?


dm17:
Is there a way to display the output from `intel_gpu_top` in btop already? If not, that is kind of what I'm looking for... And perhaps other users can use the same method to poll radeontop and similar.
Thanks; btop is great!

42Willow:
https://github.com/aristocratos/btop/pull/529#issuecomment-1632328432

Support for GPU's was added in this PR, but unfortunately for now there isn't any intel GPU support (99%, unless it just isn't working for me)

notorand-it:
As of 1.3.2, no Intel GPU (mine is Iris Xe Graphics) usage is shown.

Sarajevo1914:
thinkpad T14 Gen 2 i5 1135g7 is not shown (btop 1.3.2)

dontknowhy:
Good news:There is some other way to show Team Blue's GPU usage:
![Screenshot_20240524_184554](https://github.com/aristocratos/btop/assets/85813508/5f0bc2d1-248f-4ad9-992c-4b2f400b2a3d)

Bad news:I think it uses DRM to get the GPU's usage.
[Mission Center](https://gitlab.com/mission-center-devs/mission-center)

As [this pull message](https://github.com/aristocratos/btop/pull/529#issuecomment-1632328432) said,it's hard to use the same way as [igt-gpu-tools](https://gitlab.freedesktop.org/drm/igt-gpu-tools).

Antiarchitect:
btop 1.3.2 IrisXe HP Pro Book - not shown

dm17:
In terms of AMD and Intel GPUs, I've found `nvtop` to be significantly nicer than `radeontop` and `intel_gpu_top`. Would be cool to have them all working in `btop` though!

notorand-it:
The point here, IMHO, is not to the the best ever GPU usage tracking, but rather to add (for now) an idea on how much the GPU(s) are loaded. Nicer, fancier, more accurate can come later.

aristocratos:
Added in commit 24fc97c

dontknowhy:
> Added in commit 24fc97c

@aristocratos 
It works for me.
But why I need to give it my root permission to see my GPU usage?
Is there any way to use it without root?

aristocratos:
@dontknowhy 
Because the user you are launching btop with doesn't have permission to read SYSFS.

You can either check documentation for your distribution how to fix that.
[...]

PR #903, opened 2024-07-31
Intel GPU support

bjia56:
This is a possible implementation for #644. It uses stripped down code from `igt-gpu-tools` to handle requests to the Linux perf framework and access metrics from the Intel PMU in the same way that `intel_gpu_top` works. Tested on an Intel Gemini Lake processor's iGPU.
image

Much of the line count comes from sources pulled (and modified) from `igt-gpu-tools`, including a database of Intel chipsets and their corresponding pretty-print names. Extra dependencies that would have been pulled in by `igt-gpu-tools` were removed to keep `btop` standalone.


ingenarel:
oh my god i was searching how to have a gpu graph. currently using igpu-tools. would love to see this merged!

bjia56:
If folks are interested in testing out this change, I'm building downloadable btop zips (with the Intel support) here: https://github.com/bjia56/btop-builder

My use case is to run btop as part of a Docker or LXC container as part of the Scrypted home video platform. Using the Intel PMU requires privileged/root permissions, so the above builds also contain a workaround for pulling GPU metrics from an http server running on the host.

aristocratos:
@bjia56 
Nice work!
Let me now when you think it's ready to go from a draft to a "sharp" PR and I'll start going trough it.

I'm assuming there aren't any conflicting licensing from the code you pulled from `igt-gpu-tools`?

bjia56:
I can clean things up a bit and fix the Makefile (have only been using the CMake builds myself).

Is the http-based metrics endpoint (as a fallback data source) of interest for upstreaming, or would you like to keep only the Intel PMU collection method? The tradeoff is the http endpoint is non-standard, while the PMU requires sudo (or at least CAP_PERFMON permissions). For something like Scrypted, which can be internet-facing, we'd like to minimize the permissions required to run containers.

Regarding licensing, to my understanding `igt-gpu-tools` is MIT-licensed: https://gitlab.freedesktop.org/drm/igt-gpu-tools
The part that I'm unsure of is the DRM headers, as those appear to be pulled from the Linux kernel repo. I am not a lawyer, but since these sources retain their licensing headers, perhaps that's ok?

aristocratos:
@bjia56 
> Is the http-based metrics endpoint (as a fallback data source) of interest for upstreaming [...]

I rather avoid pulling in an entire http library for a single use case. As long as we make sure the CAP_PERFMON permission/sudo requirement is well documented. Could maybe add a `make setcap` to the Makefile that runs `setcap` on the binary with the correct arguments to make it even easier. Could also be a good idea to log a WARNING when starting up if permissions is missing.

> Regarding licensing, to my understanding igt-gpu-tools is MIT-licensed [...]
The part that I'm unsure of is the DRM headers [...]

Looks like there is a MIT notice on the DRM headers also, so if that is the case from the origin it shouldn't be any issue.

It's only a problem if it was GPL, since GPL don't mix with Apache.

aristocratos (approved):

aristocratos:
Nice work @bjia56 :)

Merging in to main.

bjia56:
Thanks for finishing up this PR!

RudolphSedlin:
Are we certain that this works for Alchemist GPUs using the i915 KMD? I would assume so given the KMD is the same, and yet my A770 is unrecognized. I am unsure on the means to investigate this.

bjia56:
Does the A770 show up in `intel_gpu_top`?

RudolphSedlin:
Yeah, its shown as just "Intel Dg2 (Gen 12)". Hardware device is also as expected, just /dev/dri/card1. This is rather bizarre, given the codebase should be the same, though it should be wroth noting I cleanbuilt igt-gpu-tools, perhaps btop is out-of-date somehow?

bjia56:
The reference igt code that was pulled into btop was commit 0f02dc176959e6296866b1bafd3982e277a5e44b from https://gitlab.freedesktop.org/drm/igt-gpu-tools.git. btop uses the PMU to access metrics, so it requires sudo permissions same as igt. If the device itself is unknown by the built-in Intel chip database, then it should show as a Generic device, if I recall correctly.

andras-tim:
I compiled **btop** recently with `GPU_SUPPORT=true`, and then with `sudo` I saw my Intel GPU.

RudolphSedlin:
How odd, then the device is not generic. Tried to compile with GPU_SUPPORT=true, no dice. May try with Xe, maybe there is something there. @andras-tim, do you get metrics on your Alchemist card?

luyu-wu:
Is there anyway to do this without root? I believe nvtop (and missioncenter) gets basic usage stats without root permissions!
I'll try and find some time to look into it if it's possible (and applicable here).

lichrot:
> Is there anyway to do this without root? I believe nvtop (and missioncenter) gets basic usage stats without root permissions! I'll try and find some time to look into it if it's possible (and applicable here).

For those who are still looking for a solution, there's one mentioned in the [Intel kernel driver issue](https://gitlab.freedesktop.org/drm/i915/kernel/-/issues/5018)

To allow `btop` (or similar programs that require access to Intel iGPU sysfs) to run without elevating privileges: `setcap cap_perfmon=+ep /path/to/btop/executable`

PR #950, opened 2024-10-11
Fixed use of the wrong "too" in an error message.

amorous-apple:
I saw that the error message was using the wrong "too" when opening the GPU menu on my small display and decided to make a quick fix :)


imwints:
Closed-by: #1166

PR #958, opened 2024-11-03
intel_name_lookup_shim.c (get_intel_device_name): Fix SEGFAULT

artyom-poptsov:
btop would always fail with "Segmentation fault" when used on machines where the GPU does not have a codename (e.g. on embedded Intel graphics on Intel(R) Atom(TM) CPU D2500.)  The reason for this behavior is that when a GPU does not have codename (it is NULL) the call to "strcpy" segfaults as the procedure effectively tries to access a NULL pointer.

* src/linux/intel_gpu_top/intel_name_lookup_shim.c (get_intel_device_name): Add a check if "info->codename" is null; if it is, set the device name to "(unknown)" to prevent the SEGFAULT error.


aristocratos (approved):
Good fix.
Thanks for contributing!

PR #959, opened 2024-11-03
src/btop_menu.cpp: Fix the description of loglevel option

artyom-poptsov:
* src/btop_menu.cpp (categories): Fix the description of loglevel option in the menu: replace missing "error.log" with actual "~/.config/btop/btop.log" (as in "btop_config.cpp".)

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -360,7 +360,7 @@ namespace Menu {
 				"Show discharge power when discharging.",
 				"Show charging power when charging."},
 			{"log_level",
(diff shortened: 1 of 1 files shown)


artyom-poptsov:
Hmm, it seems strange that "macOS CMake" CI action has failed despite that I haven't changed any CMake files.  Could it be that this failure is unrelated to the changes in the pull request?

imwints:
Can you update the commit to reflect the changes from https://github.com/aristocratos/btop/commit/5f9147d55b4ef872aff81385f2123356dbfcde07?

PR #960, opened 2024-11-03
Support Intel GPUs before Gen-6 (patch from upstream)

w8jcik:
From https://gitlab.freedesktop.org/drm/igt-gpu-tools/-/commit/c727cc3fad56f16f1a9f35023ae5dd7111976fa3.

With this change `intel_gpu_top` works with Intel GMA chips instead of terminating with _Failed to initialize PMU_.

When it comes to `btop`:
- Without this change my Gen-4 Intel GPU (GMA 4500MHD) is not visible in `btop`.
- With the change both of my GPUs (Gen-4 and Gen-9 on two different PCs) are visible in `btop`.


aristocratos (approved):

Issue #808, opened 2024-03-22
[REQUEST] Option to disable GPU monitoring


varma02:
**Is your feature request related to a problem? Please describe.**
The new GPU monitoring is great, but on my laptop I don't always want it active because it wakes the dedicated GPU from sleep.

**Describe the solution you'd like**
A toggle in the config to disable the GPU monitoring or some fix to not wake the GPU would be great.

**Describe alternatives you've considered**
Building without GPU_SUPPORT enabled is an option, but I don't want to do that every time an update comes out.

**Additional context**
None

Laci0503:
Same issue here. Maybe a switch to only monitor the integrated gpu would be nice.

Issue #738, opened 2024-01-27
[REQUEST] Avoid waking up Nvidia discrete GPUs


UltimateBoomer:
**Is your feature request related to a problem? Please describe.**
On laptops with Nvidia discrete GPUs that has been set up to enter d3cold power state when idle, running btop will cause the discrete GPU to wake up, and also make btop show a blank screen a few seconds.

**Describe the solution you'd like**
Btop should check if the GPU is in an idle power state before polling its stats. I think this should be possible by reading `/sys/class/drm/card*/device/power_state`, which will provide the power state of the GPU without waking it up. If the GPU is not in d0, then the GPU monitoring bar should show that the GPU is sleeping.

**Describe alternatives you've considered**
Btop has the functionality to hide GPU stats with the "show GPU info" option, but this doesn't actually disable GPU monitoring and thus leads to the same issue. The only way to work around this issue I found is to compile btop without GPU monitoring.

**Additional context**
I tested this on a laptop with Intel i7-11800H and Nvidia RTX 3060.

LnLcFlx:
I'd like this feature very much!

Also, if "show gpu info" is off, the gpu goes back to sleep after btop is startet.
So apparently it only accesses the gpu once at the beginning.
And after the gpu is back to sleep, closing btop has a similar delay to starting btop starts when it wakes up the gpu.

rigel314:
I have the same problem with the AMD discrete GPU on my laptop.  My current workaround is to have a custom wrapper around the (amd gpu specific) librocm_smi64.so library which lies to btop and tells it there's only one GPU.  This wakes up the dGPU briefly, but then it goes back to sleep shortly, while btop continues to run.

cugola:
Same issue and same request here. This missing "feature" made me going back to bpytop (sigh).

PR #984, opened 2024-11-29
Add ability to select which GPU vendors to enable

thedustinmiller:
I added a simple menu entry that specifies which GPU vendors should be enabled. This is primarily intended to address systems with multiple GPUs, as currently btop initializes all GPUs, even if they were otherwise suspended. I think this would address the following issues.

#738 
#808


src/linux/btop_collect.cpp
```diff
@@ -295,9 +295,13 @@ namespace Shared {
 
 		//? Init for namespace Gpu
 	#ifdef GPU_SUPPORT
-		Gpu::Nvml::init();
-		Gpu::Rsmi::init();
-		Gpu::Intel::init();
+	        auto shown_gpus = Config::getS("shown_gpus");
```
imwints: Indentation

src/linux/btop_collect.cpp
```diff
@@ -295,9 +295,13 @@ namespace Shared {
...
+	        if (s_contains(shown_gpus, "nvidia"))
```
imwints: Add brackets around if blocks.

src/linux/btop_collect.cpp
```diff
@@ -295,9 +295,13 @@ namespace Shared {
...
+	        	Gpu::Nvml::init();
+		if (s_contains(shown_gpus, "amd"))
+			Gpu::Rsmi::init();
+		if (s_contains(shown_gpus, "intel"))
+			Gpu::Intel::init();
```
imwints: Add brackets around if blocks for readability.

imwints (changes requested):
Please squash the commits into a single one.

src/btop_menu.cpp
```diff
@@ -533,6 +533,11 @@ namespace Menu {
 				"Horizontally mirror the GPU graph.",
 				"",
 				"True or False."},
+			{"shown_gpus",
+				"Manually set which gpu vendors to show.",
+				"",
+				"Available values are \"nvidia amd intel\".",
```
imwints, line 539:
```suggestion
				"Available values are \"nvidia\", \"amd\", and \"intel\".",
```

ratcashdev:
@thedustinmiller plz update as suggested. I'd love to see this merged

src/btop_menu.cpp
imwints: As suggested in another issue it would be even better to highlight the options using font features (__bold__, _italic_)

imwints:
Sorry, I just know came to testing this. The menu description leaks past the border of the menu. We need to put a `\n` in the listing of the GPU vendors. Also what do you think about renaming the option the `Gpu vendors`?

ratcashdev:
What's the showstopper here that blocks merge?

src/btop_menu.cpp
```diff
@@ -559,6 +559,11 @@ namespace Menu {
 				"Horizontally mirror the GPU graph.",
 				"",
 				"True or False."},
+			{"shown_gpus",
+				"Manually set which gpu vendors to show.",
+				"",
+				"Available values are \"nvidia\", \"amd\", and \"intel\".",
```
aristocratos: This line will not fit in the options menu, please move `\"nvidia\", \"amd\", and \"intel\".` to a new line below.

Please also add a blank line below that and at the bottom add `A restart is needed to see changes.`

aristocratos (approved):

PR #992, opened 2024-12-08
Add Kanagawa theme

tocariimaa:
Port of the Neovim Kanagawa theme.

### DIFF
--- a/themes/kanagawa.theme
+++ b/themes/kanagawa.theme
@@ -1,0 +1,76 @@
+theme[main_bg]="#16161d"
+theme[main_fg]="#dcd7ba"
+theme[title]="#c8c093"
+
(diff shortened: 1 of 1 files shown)


themes/kanagawa.theme
```diff
@@ -0,0 +1,76 @@
+theme[main_bg]="#16161d"
```
aliaksandr-trush, line 1: main_bg is 1f1f28
example: https://github.com/rebelot/kanagawa.nvim/blob/18ae5771b22d8f913ef541c827a8180b4cb12c8a/extras/kitty/kanagawa.conf#L9

aliaksandr-trush (changes requested):

PR #1030, opened 2025-02-01
Drop macos 12 build, add v14 and v15

yarrick:
macOS 12 is no longer available on github runners.

Use a shared setup instead of a copy per version.

### DIFF
--- a/.github/workflows/continuous-build-macos.yml
+++ b/.github/workflows/continuous-build-macos.yml
@@ -30,8 +30,14 @@ on:
       - '.github/workflows/continuous-build-macos.yml'
 
 jobs:
-  build-macos12:
-    runs-on: macos-12
+  build-macos:
+    strategy:
+      matrix:
+        os:
+          - {runner: 'macos-13', version: 'Ventura'}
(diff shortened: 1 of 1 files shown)


aristocratos (approved):

jauderho:
As commented on 4d465f9e294f8833a4e1ccf5caaca3e53cf4928e , I would like to see the macOS builds stick around albeit with newer versions. 

I have this working on my fork and can submit a PR if there is interest. See https://github.com/aristocratos/btop/commit/666b0840ee4f9dedde3f7072dbdf6f233dbe899c

imwints:
@jauderho there is no comment...?

I don't understand, there are builds for 13, 14 and 15.

imwints:
> But it does appear to still build for x86 even for newer macOS versions.

You are basically correct. It's just not labeled properly.

We produce a `*x86_64*` file, however the [github action runner](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) for MacOS 14 and later is actually an arm machine.

Issue #996, opened 2024-12-14
[BUG] Segfault (stack overflow) when pressing down arrow after filtering processes


Jacajack:
**Describe the bug**

Btop suffers a segmentation fault when I search for a running process and then press the down arrow to select it. This doesn't happen every time - it seems to be quite random as I still haven't figured out the exact circumstances when that happens.

**To Reproduce**

1. Open `btop`
2. Press `/` to search for process
3. Type something - e.g. `pipewire`
4. Press down arrow.
5. It sometimes crashes

**Expected behavior**

No segfault.

**My theory**

I glanced at the source code. You're setting the `proc_filtering` bool attribute in config to false and then recursively calling `process()`:

https://github.com/aristocratos/btop/blob/5535f50e2023f032309333ae6bade3a9824d2f56/src/btop_input.cpp#L284-L293

Now, if this property happens to be locked, this new value won't be placed in `bools`, but rather in `boolsTmp`:

https://github.com/aristocratos/btop/blob/5535f50e2023f032309333ae6bade3a9824d2f56/src/btop_config.hpp#L97-L101

Unless the `proc_filtering` is unlocked, this thread is going to recurse into infinity, because the value for `filtering` is read form `bools` and not `boolsTmp` in the config class. 

https://github.com/aristocratos/btop/blob/5535f50e2023f032309333ae6bade3a9824d2f56/src/btop_config.hpp#L81-L82

**Screenshots**

![Screenshot_20241214_221903](https://github.com/user-attachments/assets/e6e6c633-519d-44b4-92e0-78a3e21e642a)

**Info (please complete the following information):**
 - btop++ version: `1.4.0`, `1.4.0+5535f50`
 - Binary: from Arch official repo, self-compiled
 - Architecture: x86_64
 - Platform: Linux
 - Kernel: `6.12.4-arch1-1`
 - Terminal used: `alacritty`
 - Font used: `Iosevka Term Extended`

**Additional context**

Contents of `~/.config/btop/btop.log`:
```
2024/12/14 (22:24:24) | ===> btop++ v.1.4.0
2024/12/14 (22:24:24) | DEBUG: Running in DEBUG mode!
2024/12/14 (22:24:24) | INFO: Logger set to DEBUG
2024/12/14 (22:24:24) | DEBUG: Setting LC_ALL=en_US.UTF-8
2024/12/14 (22:24:24) | INFO: Running on /dev/pts/3
[...]

PR #1031, opened 2025-02-07
Lock/unlock config to avoid infinite recursion when exiting process filtering with down arrow

Jacajack:
This is a potential fix for #996. This doesn't seem like a great solution, but seems to be enough to fix the issue.

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -288,6 +288,8 @@ namespace Input {
 						Config::set("proc_filtering", false);
 						old_filter.clear();
 						if(key == "down"){
+							Config::unlock();
+							Config::lock();
 							process("down");
 							return;
 						}


Louis-Ladd:
Bump

I have run into this issue as well and I do think this is a good solution. 

Perhaps we could consolidate this into a `Config::flushLocks()` so this is easily callable around different systems in case locking is an issue again.

Jacajack:
Yup, I'm still here - would like to get this merged too :)

> Perhaps we could consolidate this into a Config::flushLocks() so this is easily callable around different systems in case locking is an issue again.

That's a nice idea, but I'm worried that `flushLocks()` leaving the config in a locked state might be slightly counter intuitive from interface standpoint.  If I do `if (locked) { unlock(); lock(); }` to prevent that, this creates another potential TOCTOU issue. Maybe it's better if it remains a local "hack"? What do you think?

Louis-Ladd:
That is fair, I guess giving ourselves the footgun of unlocking and locking anywhere wouldn't be smart.

Def on you with it staying a local hack.
Hopefully this gets reviewed as I'm sure @aristocratos has better insight into the overall flow of the locking system.

imwints:
> That is fair, I guess giving ourselves the footgun of unlocking and locking anywhere wouldn't be smart.

I agree. We need to figure out why this race for reading `bools` happens and how we could solve this globally. Maybe making `bools` atomic works directly, it's not like where are seeking for maximum performance here.

But it requires to understand the `Config` code in a greater context.

aristocratos:
My intent is to rewrite the config system completely lock-free later this summer.

So if this band-aids the current issue, I would say that's good enough. No need to delve deeper in the flawed configuration system, since it will replaced soon.

Jacajack:
> We need to figure out why this race for reading bools happens and how we could solve this globally.

The root cause is generally `process()` being called recursively in that special case when `"down"` is pressed when filtering processes. This conflicts with the "caching" behavior of the config class.

> Maybe making bools atomic works directly, it's not like where are seeking for maximum performance here.

I don't know the exact plans for the config system rewrite, but if I can add my two cents here, I'd advise avoiding overly complex synchronization for that purpose. Having separate atomic variables creates opportunity for another pitfall when you need to change more than one value atomically. There's no way to do it without a global lock, I'm afraid.

I imagine the perfect solution would be separating responsibilities such that there's only one user interaction thread updating the config class, and multiple worker/renderer threads only reading it. Then, at the beginning of the render loop: lock a mutex, create a local copy of the config, unlock the mutex and do the job using the copied config. 

> So if this band-aids the current issue, I would say that's good enough.

I rebased onto `main` locally and it still seems to work. If you're happy with this workaround, I suppose we can continue with this PR as a temporary remedy.

aristocratos:
> I don't know the exact plans for the config system rewrite, but if I can add my two cents here, I'd advise avoiding overly complex synchronization for that purpose. Having separate atomic variables creates opportunity for another pitfall when you need to change more than one value atomically. There's no way to do it without a global lock, I'm afraid.
> 
> I imagine the perfect solution would be separating responsibilities such that there's only one user interaction thread updating the config class, and multiple worker/renderer threads only reading it. Then, at the beginning of the render loop: lock a mutex, create a local copy of the config, unlock the mutex and do the job using the copied config.

It will be even simpler than that since it will go hand in hand with changes I'll be doing to the threading model.

Only the "main" thread will change the configuration, the "runner" threads will receive a copy of the configuration at start and the few configuration values that can be changed because of new data in the collection stage will be synced by the main thread when data is returned.

No need for any mutexes. The current "runner" thread will be broken up in a more functional fashion, with futures that return values, instead of changing data accessible by the main thread.
And no data will be passed between running threads except for specific thread control logic (like "stop") in the form of atomic bools.

imwints:
Please remove the merge commit, after that LGTM.

Jacajack:
Okay, rebased on top of `main`

imwints (approved):

PR #1058, opened 2025-03-06
Replace brackets surrounding process sorting options with arrows

taha-yassine:
This PR replaces the square brackets surrounding the process sorting options with arrows so it's more in-line with the rest of the tui (`select` is surrounded by arrows too for example). Also, arrows are imo less confusing than brackets.

Before
image

After
image

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1410,8 +1410,8 @@ namespace Net {
 
 			//? Interface selector and buttons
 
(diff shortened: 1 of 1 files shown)


imwints (changes requested):
Then please also change this in the `net` box.

imwints:
I'll merge this when the CI failures are resolved

imwints (approved):

PR #1074, opened 2025-03-22
include/fmt/core.h: fix building with clang-20

dpronin:
During compilation with clang-20 we receive an error

```bash
include/fmt/format-inl.h:1380:28: error: call to consteval function 'fmt::basic_format_string<char, unsigned int &>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
 1380 |       out = format_to(out, FMT_STRING("{:08x}"), value);
```

This is due to incompliance of **consteval**-ness of all the call stack up to the target

I think that **constexpr** would do here


imwints:
Closed-by: https://github.com/aristocratos/btop/pull/1091

PR #1075, opened 2025-03-22
use correct definitions in debug mode with libc++ and libstdc++

dpronin:
- llvm suggests we should use _LIBCPP_HARDENING_MODE rather than _LIBCPP_ENABLE_ASSERTIONS and they are about to remove _LIBCPP_ENABLE_ASSERTIONS from llvm's sources at all, so we need to prepare
- libstdc++ uses _GLIBCXX_ASSERTIONS, libc++ by llvm uses _LIBCPP_HARDENING_MODE, so let's use them separately

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,10 +143,29 @@ if(HAS_FCF_PROTECTION)
   target_compile_options(btop PRIVATE -fcf-protection)
 endif()
 
+check_cxx_source_compiles(
+  "
+  #include <algorithm>
+
+  #if !defined(__clang__) || !defined(_LIBCPP_VERSION)
+  # error \"This is not clang with libcxx by llvm\"
+  #endif
+
+  int main() {
+    return 0;
+  }
+  "
+  IS_CLANG_LIBCPP
+  FAIL_REGEX "This is not clang with libcxx by llvm"
+)
+
 target_compile_definitions(btop PRIVATE
   FMT_HEADER_ONLY
   _FILE_OFFSET_BITS=64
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -135,10 +135,29 @@ if(HAS_FCF_PROTECTION)
   target_compile_options(btop PRIVATE -fcf-protection)
 endif()
 
+check_cxx_source_compiles(
+  "
+  #include <algorithm>
```
imwints, line 140: Can you get away with using a smaller header like `<__config>`?

imwints: It's around 8 times faster. I don't see a reason to use some random header when we know exactly where the macro is defined.

dpronin: Mmm, indeed. '8 times faster' finding, which is known only from your words, just words, by the way, is absolutely essential when talking about configuring being done only once. Of course, it is. Also, 8 times could be different, you know. 1us is 8 times faster than 8us, or 1ms 8 times faster than 8ms. Do you really find it serious to speak about? Or we're talking about 8 minutes instead of 1 minute?
Also, we do not know where the macro is defined. We know that every header of STL implementation includes something that inposes _LIBCPP_VERSION, this is the fact and is said in the LLVM documentation. Tomorrow there could be no __config file anymore because it is not 'conventional' and has been substituted by another and what it would lead to? I may imagine how `__config` could be substituted or _LIBCPP_VERSION is moved to another place in the future but I cannot imagine how is `algorithm` would be replaced

PR #1079, opened 2025-03-29
Allows MidnightBSD to build btop using the existing FreeBSD support.

laffer1:
Tested with GCC 14 / gmake on MidnightBSD 3.2.2 amd64

(Upstreaming Patch in mports https://github.com/MidnightBSD/mports/commit/cc90a487249395fb6ecdf0ca2b863d82fbfc42fb)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ $(error $(shell printf "\033[1;91mERROR: \033[97m$(CXX) can't statically link gl
 		endif
 	endif
 
-	ifeq ($(PLATFORM_LC),$(filter $(PLATFORM_LC),freebsd linux))
+	ifeq ($(PLATFORM_LC),$(filter $(PLATFORM_LC),freebsd linux midnightbsd))
 		override ADDFLAGS += -DSTATIC_BUILD -static
 	else
 		ifeq ($(CXX_IS_CLANG),false)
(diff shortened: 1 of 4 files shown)


imwints:
Does CMake support MidnightBSD?

laffer1:
It's not upstreamed yet, but it does work with the MidnightBSD cmake package on MidnightBSD.

Makefile
```diff
@@ -146,6 +146,19 @@ else ifeq ($(PLATFORM_LC),macos)
 	THREADS	:= $(shell sysctl -n hw.ncpu || echo 1)
 	override ADDFLAGS += -framework IOKit -framework CoreFoundation -Wno-format-truncation
 	SU_GROUP := wheel
+else ifeq ($(PLATFORM_LC),midnightbsd)
```
imwints, line 149:
```suggestion
else ifeq ($(PLATFORM_LC),$(filter $(PLATFORM_LC),freebsd midnightbsd))
```

and remove on of the blocks?

imwints (changes requested):
If you can, please add the logic for CMake as well. Otherwise I can do it for you.

Makefile
laffer1: It should be done.

imwints (changes requested):
Please do not add merge commits. Either rebase or I will debase before merging I to main. Also please do both changes on a single commit.

imwints:
Again, please squash all commits into one.

imwints (approved):

PR #1097, opened 2025-04-14
CI: Bump OpenBSD release to 7.6

imwints:


imwints:
Seems like the package index is gone for `7.4` 😮

aristocratos (approved):

Issue #1068, opened 2025-03-16
[REQUEST] Network speed: MiB/s (or MB/s) always with Mbps (never Mibps!)


georgev93:
**Is your feature request related to a problem? Please describe.**
When validating my ISP's dubious claims of network speed, I like comparing the measured speed against what I'm paying for. The advertised speed is in Mbps. I like seeing my disk sizes in MiB, so I don't want to switch to base 10 everywhere, but I'd need base 10 for network speed to compare with my ISP's advertised speed.

**Describe the solution you'd like**
Download/Upload speed in the net view should always be in Mbps, regardless of the base10 setting selection in the menu. Or, if that's too disruptive of a change, a base10 config choice could be added to the net menu (with options of `true`, `false`, and `default` to optionally use the base10 setting in the main menu page).

**Describe alternatives you've considered**
N/A

**Additional context**
I enjoy C++, so I'd be happy to send a PR for this! I've poked through the code, and I've got an idea on how to implement this. Before I started down that path, I wanted to get a sense of whether y'all would be supportive of such a change.

imwints:
Please go ahead and open a PR! I'd like to see this as a second option in the `net` settings, with a comment in the `Base 10 sizes` option saying that this additional settings for the network speeds exists.

PR #1101, opened 2025-04-19
Adding a menu option to show bitrates in base 10 separate from the setting to show bytes/bits in base 10

georgev93:
Closes: #1068

### DIFF
--- a/src/btop_config.hpp
+++ b/src/btop_config.hpp
@@ -53,6 +53,7 @@ namespace Config {
 #ifdef GPU_SUPPORT
 	const vector<string> show_gpu_values = { "Auto", "On", "Off" };
 #endif
+    const vector<string> base_10_bitrate_values = { "Auto", "True", "False" };
 	extern vector<string> current_boxes;
 	extern vector<string> preset_list;
 	extern vector<string> available_batteries;
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -197,6 +197,8 @@ namespace Config {
 
 		{"net_iface", 			"#* Starts with the Network Interface specified here."},
 
+	    {"base_10_bitrate",     "#* \"True\" shows bitrates in base 10 (Kbps, Mbps). \"False\" shows bitrates in binary sizes (Kibps, Mibps, etc.). \"Auto\" uses base_10_sizes."},
+
 		{"show_battery", 		"#* Show battery stats in top right if battery is present."},
 
 		{"selected_battery",	"#* Which battery to use if multiple are present. \"Auto\" for auto detection."},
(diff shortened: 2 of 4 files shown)


imwints (approved):

imwints (changes requested):
Please add a reference to `base_10_sizes` that there is this additional option under `net/base_10_bitrate` so users are not confused why their change doesn't affect this.

This should be made a tristate option to default to the `base_10_sizes` option as you suggested in the issue, however this can be done in a follow up later on if we have the code for that.

georgev93:
Tristate… that's a great idea!

Apologies, I had completely forgotten the solutions we had discussed and failed to re-read what I typed up thoroughly enough. I'd be happy to take your feedback to get this PR just right before we merge it; I'll take another swipe.

georgev93:
Alright, here's a tristate solution. I started to add some error handling code to handle bad values in the config for `base_10_sizes`, but the program actually handled things pretty gracefully on its own, so I left out the additional error handling code. This implementation treats `Auto` as the default, so if a bad value is plugged into `base_10_sizes`, it will go with `Auto`.

Originally we had discussed the three options being `True`, `False`, and `Default`, but `Auto` seemed a bit more consistent with some surrounding options. I'm not married to `Auto`, so we can call it whatever.

imwints:
I was under the impression that the `net/base_10_bitrate` would be the tristate option and with set to `auto` would use whatever is provided in `base_10_sizes`.

imwints:
`Auto` is great btw!

georgev93:
Yep, once again that's a great idea which is also what I said I would do in the original issue... thanks for your patience!

imwints:
To wrap this up, the default should be `Auto` so upgrading won't affect current configs. `Auto` should be the first tristate option to reflect that.

Also, please squash all commits into one and edit the description so it describes what was changed in the end now.

```shell
git reset --soft <sha of your first commit>
git commit --amend
```

After that LGTM.

georgev93:
The default option is now `Auto`, the branch is squashed, rebased, and tested after the rebase

src/btop_config.hpp
```diff
@@ -53,6 +53,7 @@ namespace Config {
 #ifdef GPU_SUPPORT
 	const vector<string> show_gpu_values = { "Auto", "On", "Off" };
 #endif
+    const vector<string> base_10_bitrate_values = { "True", "False", "Auto" };
```
imwints, line 56: Please make this the first option as well.

```suggestion
    const vector<string> base_10_bitrate_values = { "Auto", "True", "False" };
```

imwints (approved):

imwints:
Thanks!

PR #1106, opened 2025-04-19
Deprecate GCC 10

imwints:


aristocratos:
If we're dropping GCC10 support the conditional use of semaphore.h could also be removed.

But does this compile on GCC11 with the c++ version flag removed?

imwints:
> But does this compile on GCC11 with the c++ version flag removed?

https://github.com/aristocratos/btop/actions/runs/14549583067/job/40819161102

Yes. _I think_ this was a workaround for GCC 10 from whoever did the CMake integration. We set the C++ language version in CMake with the `CMAKE_CXX_STANDARD` variable, which tests the compiler for the required standard and adds the flag. In the workflow log you can see the flag set by CMake is `-std=c2a` and the flag I deleted is a bit later on the command line.

aristocratos (approved):

Issue #1105, opened 2025-04-19
Move global state produced by the argument parser into a CLI config struct


imwints:
Currently a couple of the options of the argument parser produce state which is stored in the `Global` namespace. As we all know, global state is evil for multithreaded programs. The options are only used in the main function and thus are well suited to be returned as a struct.

PR #1110, opened 2025-04-22
Move the config parser in it's own module and return a config struct

imwints:
Closes: #1105

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -20,6 +20,7 @@ tab-size = 4
 #include <csignal>
 #include <clocale>
 #include <filesystem>
+#include <iterator>
 #include <optional>
 #include <pthread.h>
 #ifdef __FreeBSD__
@@ -36,6 +37,7 @@ tab-size = 4
 #include <regex>
 #include <chrono>
 #include <utility>
+#include <variant>
 #include <semaphore>
 
 #ifdef __APPLE__
@@ -44,6 +46,7 @@ tab-size = 4
 	#include <limits.h>
 #endif
 
+#include "btop_cli.hpp"
 #include "btop_shared.hpp"
 #include "btop_tools.hpp"
 #include "btop_config.hpp"
@@ -51,7 +54,6 @@ tab-size = 4
 #include "btop_theme.hpp"
 #include "btop_draw.hpp"
 #include "btop_menu.hpp"
-#include "config.h"
 #include "fmt/core.h"
 #include "fmt/ostream.h"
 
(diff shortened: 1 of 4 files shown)


src/btop.cpp
```diff
@@ -1055,12 +941,12 @@ int main(int argc, char **argv) {
 	}
 
 	if (Term::current_tty != "unknown") Logger::info("Running on " + Term::current_tty);
-	if (not Global::arg_tty and Config::getB("force_tty")) {
```
imwints, line 1058: I'm not sure about what exactly this is supposed to do here.

src/btop_cli.cpp
```diff
@@ -0,0 +1,177 @@
...
+				"  {2}-c, --config {1}<file>  Path to a config file\n"
+				"  {2}-d, --debug{1}          Start in debug mode with additional logs and metrics\n"
+				"  {2}    --force-utf{1}      Override automatic UTF locale detection\n"
+				"  {2}-l, --low-color{1}      Disable true color, 256 colors only\n"
+				"  {2}-p, --preset {1}<id>    Start with a preset (0-9)\n"
+				"  {2}-t, --tty{1}            Force tty mode with ANSI graph symbols and 16 colors only\n"
+				"  {2}    --no-tty{1}         Force disable tty mode\n"
+				"  {2}-u, --update {1}<ms>    Set an inital update rate in milliseconds\n"
+				"  {2}-h, --help{1}           Show this help message and exit\n"
+				"  {2}-V, --version{1}        Show a version message and exit (more with --version)\n",
```
imwints, line 169: Note that I change a few of the options to be more Unix like (no `+`, inverse of `--option` is `--no-option`).

I don't expect this to be a breaking change since usage in scripts doesn't work anyway and if the option is typed incorrectly the user is delegated to the help section.

Makefile
```diff
@@ -2,7 +2,8 @@
...
 
-override BTOP_VERSION := $(shell head -n100 src/btop.cpp 2>/dev/null | grep "Version =" | cut -f2 -d"\"" || echo " unknown")
+override BTOP_VERSION := $(shell grep '^\s*VERSION ".*"' CMakeLists.txt | cut -d '"' -f 2)
```
aristocratos, line 5: I rather have it pull the version from btop.cpp since that's the one I'm most likely to forget to update...

aristocratos (approved):
Otherwise, looks good.

src/btop.cpp
aristocratos: Print out the tty it's running on at start?

imwints: Why it's checking the config option. I'd expect the command line option to override what's in the config file.

Makefile
aristocratos: Well, `const string Version = "1.4.0";` in the `Global` namespace is still in btop.cpp and that's the one that is displayed in the menu.

src/btop.cpp
imwints: The point is, I'm not a 100% sure that the code I replaced this line with is identical in meaning of what was there before. I'll have to check that.

Makefile
aristocratos: If the version is to be defined in the CMakeLists.txt there need to be a easy to notice comment in btop.cpp and maybe btop.hpp to make clear where the version should be updated at least.

aristocratos: Otherwise you could just include `btop_shared.hpp` in `btop_cli.hpp` an get access to `Global::Version`.

imwints: Ok so I opted for the `VERSION` macro everywhere now. I added a comment where it lives now and where it comes from. I hope this is a nicer solution than the `extern` global variable, which I'd like to get rid of for discoverability.

aristocratos: Why?
It's not a "global" variable, it's a just regular variable living in the namespace `Global`. No different from any other variable in `btop_shared.hpp` which is included in most of the .cpp files.

Using a macro instead which pulls from an external file and requires an additional include in multiple files, just for one variable doesn't seem like the best solution?

aristocratos: And for the cmake to get the correct version, shouldn't this in CMakeLists.txt work:
```bash
execute_process(
	COMMAND head -n100 src/btop.cpp 2>/dev/null
	COMMAND grep "Version ="
    COMMAND cut -f2 -d"\""
	OUTPUT_VARIABLE VERSION
	OUTPUT_STRIP_TRAILING_WHITESPACE)
```

imwints: > Using a macro instead which pulls from an external file and requires an additional include in multiple files, just for one variable doesn't seem like the best solution?

I wrongly accused `VERSION` of being a macro but it`s actually just a `std::string_view`! 🥳 

The reason for touching this part was I _thought_ it would be nice to get rid of some `extern` definitions, since I can't easily `gd` (GotoDefinition) on them. If it is holding back the rest, I will revert it.

imwints: > And for cmake to get the correct version, shouldn't this in CMakeLists.txt work:
> 
> ```shell
> execute_process(
> 	COMMAND head -n100 src/btop.cpp 2>/dev/null
> 	COMMAND grep "Version ="
>     COMMAND cut -f2 -d"\""
> 	OUTPUT_VARIABLE VERSION
> 	OUTPUT_STRIP_TRAILING_WHITESPACE)
> ```

It would, but we don't need the version info in CMake if we got it in the source code already.

aristocratos: Seems to work in vscode, maybe an issue with the language server you use in vim? :P

Yeah, I would rather that you restored `Global::Version`, I just don't see a good reason for the definition to be moved outside of the actual code.

Just including `btop_shared.hpp` in `btop_cli.cpp` seems like the more sensible option to get the version string to the cmd config parser.

aristocratos (approved):

PR #1121, opened 2025-04-30
Skip composefs and read-only filesystems

bad-co-de:


src/linux/btop_collect.cpp
```diff
@@ -1937,6 +1938,9 @@ namespace Mem {
 						//? Skip composefs (read-only overlayfs alike)
 						if (fstype == "overlay" && dev == "composefs") continue;
 
+						//? Skip read-only filesystems
+						if (skip_ro_disks && flags.starts_with("ro,")) continue;
```
imwints, line 1942: Question: Is it possible that the `ro` flag is the only option? Then the `,` needs to be dropped here.

src/linux/btop_collect.cpp
```diff
@@ -1934,6 +1935,12 @@ namespace Mem {
 								continue;
 						}
 
+						//? Skip composefs (read-only overlayfs alike)
+						if (fstype == "overlay" && dev == "composefs") continue;
```
imwints, line 1939: Surround with curly braces.

src/linux/btop_collect.cpp
```diff
@@ -1919,10 +1919,11 @@ namespace Mem {
 				if (diskread.good()) {
 					vector<string> found;
 					found.reserve(last_found.size());
-					string dev, mountpoint, fstype;
+					const bool skip_ro_disks = Config::getB("skip_ro_disks");
+					string dev, mountpoint, fstype, flags;
```
imwints, line 1923: Declare each variable on a single line.

src/linux/btop_collect.cpp
```diff
@@ -1934,6 +1935,12 @@ namespace Mem {
 								continue;
 						}
 
+						//? Skip composefs (read-only overlayfs alike)
```
imwints, line 1938: You can drop the `?` in the comments everywhere IMO.

src/linux/btop_collect.cpp
```diff
@@ -1934,6 +1935,12 @@ namespace Mem {
...
+						if (fstype == "overlay" && dev == "composefs") continue;
+
+						//? Skip read-only filesystems
+						if (skip_ro_disks && flags.starts_with("ro,")) continue;
```
imwints, line 1942: Curly braces.

imwints (commented):
Can you explain why it's beneficial to not show `composefs` filesystems?

src/linux/btop_collect.cpp
bad-co-de: fixed

deckstose:
Can you explain the rational behind the change? Why wouldn't I want to see my `composefs` filesystem? (I'm not using atomic linux distributions)

PR #1123, opened 2025-04-30
Intel GPU tweaks

bad-co-de:


src/btop_tools.hpp
```diff
@@ -463,3 +463,14 @@ namespace Tools {
 	};
 
 }
+
+namespace capi {
+	struct free_deleter {
```
imwints, line 468: We (generally) use CamelCase for structs and namespaces.

imwints (commented):
I see the value in these changes however I will not merge and maintain patches to third party code. This puts an additional burden on us. Please send the patches upstream to https://gitlab.freedesktop.org/drm/igt-gpu-tools or if they've been merged already we can update the sources.

PR #1124, opened 2025-04-30
Color package CPU temperature correctly

bad-co-de:


imwints:
What does "correctly" mean here and what was wrong before? If you fix something, create an issue first so it is documented that there is something wrong an others can find it (and know it is solved then).

imwints:
What does "correctly" mean here and what was wrong before? Can you create an issue for this so we have a reference on the issue tracker and know what what wasn't working before.

bad-co-de:
Current:
Screenshot_20260303_194956

Changed:

Screenshot_20260303_194905

aristocratos:
Fixed in b36357b

PR #1126, opened 2025-05-01
Fix MacOS tree-mode + aggregate memory/thread scaling issue

xaskii:
Fixes issue where using "Proc tree" and "Proc aggregate" on MacOS without running btop as root lets the memory and thread counts scale infinitely.

# Before:
image

# After:
non-root:
image
root:
image

### DIFF
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -1236,6 +1236,11 @@ namespace Proc {
 						cpu_t = pti.pti_total_user + pti.pti_total_system;
 
 						if (new_proc.cpu_t == 0) new_proc.cpu_t = cpu_t;
+					} else {
+						// Reset memory value if process info cannot be accessed (bad permissions or zombie processes)
+						new_proc.threads = 0;
+						new_proc.mem = 0;
+						cpu_t = new_proc.cpu_t;
 					}
 
 					//? Process cpu usage since last update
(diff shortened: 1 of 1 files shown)


src/osx/btop_collect.cpp
```diff
@@ -1229,13 +1229,18 @@ namespace Proc {
...
-					struct proc_taskinfo pti;
+					struct proc_taskinfo pti{};
 					if (sizeof(pti) == proc_pidinfo(new_proc.pid, PROC_PIDTASKINFO, 0, &pti, sizeof(pti))) {
 						new_proc.threads = pti.pti_threadnum;
 						new_proc.mem = pti.pti_resident_size;
 						cpu_t = pti.pti_total_user + pti.pti_total_system;
 
 						if (new_proc.cpu_t == 0) new_proc.cpu_t = cpu_t;
+					} else {
+						//? Reset memory value if process info cannot be accessed (bad permissions or zombie processes)
```
imwints, line 1240: IMO you can drop the `?` in the comment

imwints:
Minor nit: if you can edit the last commit message to just `initialize pti struct` or something (as this currently does not really match what is changed)

src/osx/btop_collect.cpp
xaskii: Fixed

xaskii:
> Minor nit: if you can edit the last commit message to just `initialize pti struct` or something (as this currently does not really match what is changed)

Yeah np, just pushed this

imwints (approved):

xaskii:
Thanks for maintaining this project!

PR #1134, opened 2025-05-02
Added a setting which disables the help and menu

webdevred:
This patch introduces a setting that allows users to disable the in-app help and menu, promoting direct editing of the configuration file instead. The motivation behind this feature is to streamline the interface for advanced users who prefer minimalism and already have familiarity with btop's options. By removing the menu and help overlay, it reduces visual clutter. This is particularly useful in environments where clean, distraction-free monitoring is preferred, such as tiling window managers.

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -50,6 +50,7 @@ namespace Config {
 	bool write_new;
 
 	const vector<array<string, 2>> descriptions = {
+		{"disable_help", "#* Disable menu and help. For the true minimalist!"},
 		{"color_theme", 		"#* Name of a btop++/bpytop/bashtop formatted \".theme\" file, \"Default\" and \"TTY\" for builtin themes.\n"
 								"#* Themes should be placed in \"../share/btop/themes\" relative to binary or \"$HOME/.config/btop/themes\""},
 
@@ -263,6 +264,7 @@ namespace Config {
 	std::unordered_map<std::string_view, string> stringsTmp;
 
 	std::unordered_map<std::string_view, bool> bools = {
+		{"disable_help", false},
 		{"theme_background", true},
 		{"truecolor", true},
 		{"rounded_corners", true},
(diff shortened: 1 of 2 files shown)


webdevred:
> Huh? I'm not sure if this is a serious PR or not.
> 
> What exactly gets "streamlined" in the interface by disabling two keyboard shortcuts?
> 
> > By removing the menu and help overlay, it reduces visual clutter and reinforces a more scriptable, config-driven workflow.
> 
> The simpler solution here is probably that you just don't open the menu, if you don't want to, you know...



You're absolutely right that I could simply avoid opening the menu or help, and that's what I’ve done so far. But as a CLI-focused user who values a minimal interface, I prefer being able to disable features where there is too much text shown or colored logos.

This patch isn’t about improving performance or functionality for everyone, it’s just a small quality-of-life setting for those of us who prefer to eliminate non-essential UI elements entirely.
I understand if this feels like niche functionality, but I think there’s value in supporting a more stripped-down mode for advanced CLI minimalist users. I would be happy to revise the patch in case if it doesn’t fit the project vision.

webdevred:
@aristocratos Putting the current PR aside for a moment, I’ve also been considering submitting a patch that disables the automatic addition of comments in the configuration file.

As a CLI user who tracks dotfiles in Git, I prefer to keep my config files as minimal and readable as possible, avoiding auto-generated comments helps keep things concise when editing or reviewing in a text editor.

Would you be open to a setting that disables comment generation for users with similar preferences?

webdevred:
> Please open a new issue for that

I have now opened a separate issue for that feature request https://github.com/aristocratos/btop/issues/1136
Thanks for your time.

aristocratos:
> eliminate non-essential UI elements entirely.

Again you're not eliminating any UI elements by just disabling the keyboard shortcut.

This PR is completely pointless in my opinion.
It's disabling shortcuts that you already don't use.

Issue #573, opened 2023-07-06
[Development] Add .clang-format file


nobounce:
This project should use a code formatting tool like [clang-format](https://clang.llvm.org/docs/ClangFormat.html). Clang-format allows to format the code according to guidelines put in a `.clang-format`-file.

Why consistency?
Pros
* Easy readability of the code base
* Makes it easier for new contributors to look around in the source code
* A lot of people consider this a best practice, for example [Google](https://google.github.io/styleguide/cppguide.html#Goals)
* Avoids further discussion about how an `if`-statement should look like, if it is already determined

Cons
* I seriously don't know, hit me up

`clang-format` helps to enforce consistent code style and readable code. Any developer can just write code in the style they want and with the click of a button or a single command the code is properly adjusted
[You read more code than you write](https://bayrhammer-klaus.medium.com/you-spend-much-more-time-reading-code-than-writing-code-bc953376fe19), and if we can read it faster, we can spend more time writing new code.

Even if you don't mind or care, this probably makes it easier for others to dig into the source code.
And if someone opens a pull request with code not according to the project's style, it just a matter of "Please run clang-format over that" and it's done. Simple as that.
And in the end, it's just a file. It's has no impact on the users but helps developers which use that tool.

nobounce:
This might also be interesting for a consistent style guide:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
https://github.com/cpp-best-practices/cppbestpractices/blob/master/03-Style.md#style

nobounce:
```
---
BasedOnStyle: LLVM
# Language
Language: Cpp
AccessModifierOffset: -4
AlignAfterOpenBracket: BlockIndent
AllowShortCaseLabelsOnASingleLine: true
AllowShortIfStatementsOnASingleLine: AllIfsAndElse
AllowShortLoopsOnASingleLine: true
AlwaysBreakBeforeMultilineStrings: true
BinPackArguments: false
BraceWrapping:
  BeforeCatch: true
  BeforeElse: true
BreakBeforeBraces: Custom
ColumnLimit: 120
IndentCaseLabels: true
IndentPPDirectives: BeforeHash
IndentWidth: 4
NamespaceIndentation: All
PointerAlignment: Left
QualifierAlignment: Left
ReferenceAlignment: Left
Standard: c++20
TabWidth: 4
UseTab: Always
```

This could be a sample file which follows the general style of btop's code and adjust smaller parts which are not consistent.

[...]

PR #1146, opened 2025-05-06
Add .clang-format file for consistent code formatting

deckstose:
This does _not_ aim to be a config to do the least formatting compared to the current code base, it tries to be readable wherever possible.

This required some manual reordering of headers since preprocessor statements in between header declaration don't play nice with header sorting.

Closes: https://github.com/aristocratos/btop/issues/573

### DIFF
--- a/src/btop_shared.hpp
+++ b/src/btop_shared.hpp
@@ -4,7 +4,7 @@
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
 
-	   http://www.apache.org/licenses/LICENSE-2.0
+		   http://www.apache.org/licenses/LICENSE-2.0
 
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
@@ -30,16 +30,12 @@ tab-size = 4
 #include <unordered_map>
 #include <vector>
 
-#include <unistd.h>
-
-// From `man 3 getifaddrs`: <net/if.h> must be included before <ifaddrs.h>
-// clang-format off
 #include <net/if.h>
 #include <ifaddrs.h>
-// clang-format on
+#include <unistd.h>
 
 #if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
-# include <kvm.h>
+#	include <kvm.h>
 #endif
 
 using std::array;
@@ -51,7 +47,7 @@ using std::vector;
 
 using namespace std::literals; // for operator""s
 
-void term_resize(bool force=false);
+void term_resize(bool force = false);
 void banner_gen();
 
 extern void clean_quit(int sig);
@@ -68,7 +64,7 @@ namespace Global {
 	extern string clock;
 	extern uid_t real_uid, set_uid;
 	extern atomic<bool> init_conf;
-}
+} // namespace Global
 
 namespace Runner {
 	extern atomic<bool> active;
@@ -81,12 +77,12 @@ namespace Runner {
 
 	void run(const string& box = "", bool no_update = false, bool force_redraw = false);
 	void stop();
-}
+} // namespace Runner
 
 namespace Tools {
 	//* Platform specific function for system_uptime (seconds since last restart)
 	double system_uptime();
-}
+} // namespace Tools
 
 namespace Shared {
 	//* Initialize platform specific needed variables and check for errors
@@ -96,13 +92,11 @@ namespace Shared {
 
 #if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
 	struct KvmDeleter {
-		void operator()(kvm_t* handle) {
-			kvm_close(handle);
-		}
+		void operator()(kvm_t* handle) { kvm_close(handle); }
 	};
 	using KvmPtr = std::unique_ptr<kvm_t, KvmDeleter>;
 #endif
-}
+} // namespace Shared
 
 #if defined(GPU_SUPPORT)
 
@@ -120,28 +114,28 @@ namespace Gpu {
 
 	extern std::unordered_map<string, deque<long long>> shared_gpu_percent; // averages, power/vram total
 
-	const array mem_names { "used"s, "free"s };
+	const array mem_names {"used"s, "free"s};
 
 	//* Container for process information // TODO
 	/*struct proc_info {
-    unsigned int pid;
-    unsigned long long mem;
+unsigned int pid;
+unsigned long long mem;
 	};*/
 
 	//* Container for supported Gpu::*::collect() functions
 	struct gpu_info_supported {
-		bool gpu_utilization = true,
-		   	 mem_utilization = true,
-				 gpu_clock = true,
-				 mem_clock = true,
-				 pwr_usage = true,
-				 pwr_state = true,
-				 temp_info = true,
-				 mem_total = true,
-				 mem_used = true,
-				 pcie_txrx = true,
-				 encoder_utilization = true,
-				 decoder_utilization = true;
+		bool gpu_utilization = true;
+		bool mem_utilization = true;
+		bool gpu_clock = true;
+		bool mem_clock = true;
+		bool pwr_usage = true;
+		bool pwr_state = true;
+		bool temp_info = true;
+		bool mem_total = true;
+		bool mem_used = true;
+		bool pcie_txrx = true;
+		bool encoder_utilization = true;
+		bool decoder_utilization = true;
 	};
 
 	//* Per-device container for GPU info
@@ -163,7 +157,7 @@ namespace Gpu {
 		long long mem_total = 0;
 		long long mem_used = 0;
 		deque<long long> mem_utilization_percent = {0}; // TODO: properly handle GPUs that can't report some stats
-		long long mem_clock_speed = 0; // MHz
+		long long mem_clock_speed = 0;					// MHz
 
 		long long pcie_tx = 0; // KB/s
 		long long pcie_rx = 0;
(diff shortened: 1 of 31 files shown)


aristocratos:
I don't see a good reason to enforce formatting.

Formatting aren't usually what makes code hard to read, more so coding *style* and weird naming conventions.

For example the suggested column limit of 120 characters are broken in most if not all files in this repository (which is fine in my opinion, the vast majority of people are on 1080p an up).

imwints:
> For example the suggested column limit of 120 characters are broken in most if not all files in this repository (which is fine in my opinion, the vast majority of people are on 1080p an up).

This can be increased. I don't care about the specific style, just that it is consistent. Having some parameters align on the next line of a function with local variables makes by brain go 💥 

https://github.com/aristocratos/btop/blob/9bd618fd0eb92ae17202d3887f3a851494a8cd39/src/btop_shared.cpp#L181-L186

Take this for example, where does the parameter list end and where do the function local variables begin? It _is_ harder to read.


Naming conventions could be enforced as well (like CamelCase struct names and lower_case variable/function names).

> I don't see a good reason to enforce formatting.

I know we had this discussion before and you said you don't have strong feelings about it. What would be the harm to try this out? And what would be a good reason to __not__ use the same formatted code for new contributions?

aristocratos:
If you follow the git blame back for `_tree_gen` to the original you'll see that it was one line originally that later was changed, likely because of the committers own clang-format.

> What would be the harm to try this out?

Well, I guess waste of energy when all the github actions has to rerun every time someone has to correct their formatting to pass our .clang-format.

What I mean to say is that formatting aren't really that important.
And in the few cases where they are, it's not that hard to just point it out to the PR author.

imwints:
> Well, I guess waste of energy when all the github actions has to rerun every time someone has to correct their formatting to pass our .clang-format.

If we check in the .clang-format file this will hint to most contributors that there is some style guide (we can also add a notice) to follow and will make it easier on the developer side to already format the file before contributing back. If we trust the [stack overflow survey](https://survey.stackoverflow.co/2024/technology#1-integrated-development-environment) then most people will be using VSCode which has clang-format integration in it's Cpp plugin. It will integrate well within the workflow of most people.

> What I mean to say is that formatting aren't really that important.

It is. IMO even on more then variable naming. I don't know your code as well as you do, I (voluntarily 😃) need to look at it and the newly contributed code and understand what is happening. Any deviation of where I expect a certain symbol to be makes it harder. This will also make spotting mistakes easier, since the code format is not something to worry about (it's the same, and predictable, everywhere).

> And in the few cases where they are, it's not that hard to just point it out to the PR author.

This is tedious. Why not let the machine handle it? That's what they're good at.

aristocratos:
Sure, if you can make sure this doesn't happen:

image

I formatted btop_draw.cpp with this .clang-format.

In essence someone making a single change in this file and then running format document would touch about half the lines in the file. Making git blame essentially useless.

I would say the output is far from an improvement, some examples:

image

image

image

image

image

imwints:
> Sure, if you can make sure this doesn't happen:

Will work on this 🕵🏽 

> In essence someone making a single change in this file and then running format document would touch about half the lines in the file. Making git blame essentially useless.

That's not right. There is `git-clang-format`, which is used in the workflow as well. There is also `.git-blame-ignore-revs` which is supported by Github's UI to ignore commits of mass re-formatting.

> I would say the output is far from an improvement, some examples:

It's not useful in all cases. These block can be ignored with `// clang-format on/off`.

deckstose:
> Sure, if you can make sure this doesn't happen:

I guess this is unavoidable. It just shows that there is a lot of inconsistency at the moment.

deckstose:
I've given this another shot. I've found the toggle for braced initializer lists so that config and menu mappings don't look completely f-ed up. Formatting is not applied right now, but when we do, we should add a file afterwards to tell git blame to ignore the commit.

vandabbin:
> I've given this another shot. I've found the toggle for braced initializer lists so that config and menu mappings don't look completely f-ed up. Formatting is not applied right now, but when we do, we should add a file afterwards to tell git blame to ignore the commit.

Personally I'm all for this. My editor (neovim) really really doesn't like the indentation of the namespaces and it is a constant annoyance. For some reason even though I have the config setting in neovim sets that it should be allowing this indentation. It just ignores it. Making me think a plugin is screwing with it. I bet a good .clang_format file would solve it though.

deckstose:
It always indents to the start of the line even if it should indented right? 😆

vandabbin:
Yup exactly! 😆

deckstose:
I hope I addressed all the issues raised in https://github.com/aristocratos/btop/pull/1146#issuecomment-2887049281

If someone else wants to take a look at the formatted files and see if there is anything obvious that's totally out of line, I'm glad to fix it

aristocratos:
Get this error when testing this with clang-format 18.1.8 (latest available in the regular repositories in Ubuntu 24.04)

```bash
.clang-format:19:20: error: invalid boolean
BinPackParameters: OnePerLine
                   ^~~~~~~~~~
```

Also, why is the PR removing a bunch of comments?

deckstose:
> Also, why is the PR removing a bunch of comments?

I have done the edits in work tree with a bunch of other staging edits, maybe they slipped in when I split them up. However: I didn't see their usefulness, as none of the other provided headers has any explanation why they where added (and IDE tools can provide you that information anyway). As it's unrelated, I will remove them 🐦

And the order of `<net/if.h>` and `<ifaddrs.h> will be managed by `clang-format`. Makes me wonder if we should add the formatting to this PR (or right after). Not sure.

> Get this error when testing this with clang-format 18.1.8 (latest available in the regular repositories in Ubuntu 24.04)

We don't support LLVM 18 anymore. I can recommend the https://apt.llvm.org/ site, this is how we get the LLVM releases in CI as well. Otherwise a container will work or I can push a branch with the formatted code, if you want me to.

deckstose:
I've added a commit with the formatted changes.

aristocratos:
Why are all tabs replaced with spaces? That's the opposite of what CONTRIBUTING.md states.

deckstose:
Ahh yes, that's what broke my editor regarding namespaces and indentation. It's not a big deal to switch it around, but I **strongly** recommend changing the contribution guidelines instead, as spaces are more consistent across all environments (and that's my goal with this) and tabs will always break things like alignment (try to view the Makefile with 2 or 8 space-wide tabs for example). As a project, I'd be opinionated about it.

I'm open to change this but this is one of the few points I have a stronger opinion about.

aristocratos:
>  as spaces are more consistent across all environments (and that's my goal with this) and tabs will always break things like alignment (try to view the Makefile with 2 or 8 space-wide tabs for example). 

That's why the indent size is explicitly defined in the `.editconfig`.

> As a project, I'd be opinionated about it.

If I didn't care about it I wouldn't have mentioned it in the CONTRIBUTING.md.

Please don't start arguing over tabs vs spaces, I've got better things to do...

deckstose:
Done :+1:

aristocratos:
I still don't think this is an improvement.
It also weirdly seems inconsistent with what is split up and what is compacted.

Function calls are split over multiple rows even if there's just 2-3 arguments.
Meanwhile some structs are compacted to be one-liners even if there is like 6-7 variables declared.

Any manually created structure in initializer-lists are completely destroyed and ends up worse.

And do really an if statement like for example ```if (bool) var = 5``` really need to split up in to two rows? That's just a waste of space (or newline in this particular case...).

The only actual fixes I see is the small changes like correcting indent characters and indentation, moving starting braces and moving the reference/pointer symbols.

deckstose:
> Function calls are split over multiple rows even if there's just 2-3 arguments.
Meanwhile some structs are compacted to be one-liners even if there is like 6-7 variables declared.

Can you point me to a specific part so I can take a look. This is configurable.

At the moment the tool tries to fit arguments on a single line given the max line length, and if it doesn't fit it will split it up into one arg per line.

deckstose:
> And do really an if statement like for example if `(bool) var = 5` really need to split up in to two rows? That's just a waste of space (or newline in this particular case...).

"Fixed"

src/btop_shared.hpp
```diff
@@ -120,28 +114,19 @@ namespace Gpu {
...
-				 temp_info = true,
-				 mem_total = true,
-				 mem_used = true,
-				 pcie_txrx = true,
-				 encoder_utilization = true,
-				 decoder_utilization = true;
+		bool gpu_utilization = true, mem_utilization = true, gpu_clock = true, mem_clock = true, pwr_usage = true,
+			 pwr_state = true, temp_info = true, mem_total = true, mem_used = true, pcie_txrx = true,
+			 encoder_utilization = true, decoder_utilization = true;
 	};
```
deckstose, line 145: @aristocratos you mean something like this?

aristocratos: Yeah, that's one example.

I'm gonna experiment a bit with the `.clang_format` file when I've got time and see if I can get it in a state where it's not actively breaking structure...

deckstose: I agree, that looks horrendous. If you've got more, please tell me! I've already spent some time learning the ins-and-outs of the clang-format settings, and I've _tried_ my best to make to overall code look sound. I've spent a couple of hours skimming the diff and finding structures that break and fixed them already. It's probably a waste of time if you start from the beginning again.

Besides the struct above, I think the overall result looks pretty good, hopefully fixing this is the last barrier?

deckstose: The specific problem here is that this declares multiple members on a single "line":

```cpp
int x = 1, y = 2;
```

It's fixed when the declarations are split up:


```cpp
int x = 1;
int y = 2;
```

PR #1154, opened 2025-05-12
Fix `btop` for earlier macOS and PowerPC

barracuda156:
There is a way to get temperature of CPU on PowerPC, but that has to be found out. For now, at least do not try to use x86-specific code on PowerPC, it does not work and forces the app to error out on launch for no good reason.

btop

### DIFF
--- a/src/btop_shared.hpp
+++ b/src/btop_shared.hpp
@@ -31,7 +31,9 @@ tab-size = 4
 #include <unistd.h>
 
 // From `man 3 getifaddrs`: <net/if.h> must be included before <ifaddrs.h>
+// Some platforms also require <sys/socket.h> to be included before <net/if.h>.
 // clang-format off
+#include <sys/socket.h>
 #include <net/if.h>
 #include <ifaddrs.h>
 // clang-format on


imwints (changes requested):
Can you add comments on why this is necessary, and ideally put the same information in the commits?

imwints:
Gentle ping

barracuda156:
> Gentle ping

Sorry for a delay, a lot of things in pipeline. Will do this tonight.

barracuda156:
@imwints I have updated the commits (with minor changes, upon reviewing documentation), and added brief explanation with references.

src/btop_shared.hpp
```diff
@@ -31,7 +31,9 @@ tab-size = 4
 #include <unistd.h>
 
 // From `man 3 getifaddrs`: <net/if.h> must be included before <ifaddrs.h>
+// Some platforms also require <sys/socket.h> to be included before <net/if.h>.
```
imwints, line 34: The comments should be inside the `clang-format` block: both above their respective header and `ifaddrs.h` at the end. Makes it easier to read.

imwints, line 34: The link in the commit message says that `net/if.h` also my need `sys/socket.h` on OpenBSD and FreeBSD. Just to make sure, can you add it their as well?

src/osx/btop_collect.cpp
```diff
@@ -228,6 +228,7 @@ namespace Cpu {
 				macM1 = true;
 			} else {
 #endif
+#ifndef __POWERPC__
```
imwints, line 231: Nit: If there are not differences to the `__ppc__` macro from other commit, I'd like to stick to one (makes it easier to grep for it)

src/osx/btop_collect.cpp
```diff
@@ -654,6 +662,15 @@ namespace Mem {
 			mem.stats.at("used") = (p.active_count + p.wire_count) * Shared::pageSize;
 			mem.stats.at("available") = Shared::totalMem - mem.stats.at("used");
 		}
+#else
+		vm_statistics p;
+		mach_msg_type_number_t info_size = HOST_VM_INFO_COUNT;
+		if (host_statistics(mach_host_self(), HOST_VM_INFO, (host_info_t)&p, &info_size) == KERN_SUCCESS) {
```
imwints, line 668: While we are at it, please use C++'s `static_cast` here and above.

src/osx/btop_collect.cpp
```diff
@@ -646,6 +653,7 @@ namespace Mem {
 		auto &mem = current_mem;
 		static bool snapped = (getenv("BTOP_SNAPPED") != nullptr);
 
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 1060 && !defined(__ppc__) && !defined(__i386__)
 		vm_statistics64 p;
 		mach_msg_type_number_t info_size = HOST_VM_INFO64_COUNT;
 		if (host_statistics64(mach_host_self(), HOST_VM_INFO64, (host_info64_t)&p, &info_size) == 0) {
```
imwints, line 659: Please use KERN_SUCCESS here as well.

imwints (commented):
Thanks for the references. Would it make sense to exclude the SMC code completely from being compiled?

src/osx/btop_collect.cpp
barracuda156: There are: `__ppc__` is a 32-bit ppc, `__POWERPC__` includes 32- and 64-bit both.

barracuda156:
> Thanks for the references. Would it make sense to exclude the SMC code completely from being compiled?

Yes, I believe so.

Issue #883, opened 2024-06-21
[REQUEST] Accept process filter as command line argument


bersace:
Hello,

Thanks for btop. It's awesome software.

**Is your feature request related to a problem? Please describe.**
When working with multiprocess development, I like using btop to scrutinize the behaviour of the app. e.g. Postgres. However, I always have to retype the same filter from one project to another.

**Describe the solution you'd like**
I suggest btop to accept process filter as CLI argument.

**Describe alternatives you've considered**
I didn't found.

**Additional context**
Using a CLI argument will help providing alias, share command, reuse shell history, etc.

Regards,
Étienne

imwints:
Can you test if this works as expected?

bersace:
Awesome. Does CI provide a binary artifact ?

imwints:
I think the artifact upload is broken at the moment ://

deckstose:
@bersace gentle ping :)

bersace:
following README right now.

deckstose:
If you need help don't hesitate to ask

bersace:
It's perfect. Thank you @imwints !

PR #1157, opened 2025-05-13
Add command line option to set an inital filter

imwints:
This will spawn btop with a filter already set, which can be edited and also deleted as usual.

Beware that the filter will ALWAYS also match btop itself, since the --filter flag on the command line will be matched.

Closes: https://github.com/aristocratos/btop/issues/883


imwints:
Can you test if this works as expected?

PR #1170, opened 2025-05-31
Themes: Add Catppuccin

SchweGELBin:
I've added the catppuccin themes from [here](https://github.com/catppuccin/btop).
I'm not the author, but I think this is okay to do with the MIT license.
I've added the copyright owner and link to the origin files in the commit messages.


deckstose:
We must add the license info in the files _at least_. Would be even better to ask upstream if they allow their work to be distributed under Apache-2.0.

PR #1172, opened 2025-06-01
Make 100ms the minimal refresh rate

### LINKED ISSUE #1171: [BUG] CLI option --update is being ignored with update_ms in the config
**Describe the bug**

When I updated from 1.4.0 to 1.4.3 the command `btop --update 3000` did not work like before: While having `update_ms` set to `100` in the `btop.conf` the command above used to override the config file. Now when the update time gets limited to 100ms. I can only set lower values to the CLI command.

**To Reproduce**

On 1.4.3 set `update_ms` to `100` in the `btop.conf` file and run:

```
btop --update 3000 # capped to 100
btop --update 3
```

**Expected behavior**

The CLI command should take priority.

 - btop++ version: 
    btop version: 1.4.3
    Compiled with: g++ (14.2.1)
    Configured with: cmake -DBTOP_STATIC=OFF -DBTOP_GPU=ON
 - Binary: static binary from release provided by nixos
 - Architecture: x86_64
 - Platform: Linux NixOS (with use of Home manager)
 - (Linux) Kernel: 6.12.24
 - Terminal used: kitty
 - Font used: -

**Additional context**

```
2025/06/01 (15:36:45) | ===> btop++ v.1.4.3
2025/06/01 (15:36:45) | DEBUG: Running in DEBUG mode!
2025/06/01 (15:36:45) | INFO: Logger set to DEBUG
2025/06/01 (15:36:45) | DEBUG: Setting LC_ALL=en_US.UTF-8
2025/06/01 (15:36:45) | INFO: Running on /dev/pts/4
2025/06/01 (15:36:45) | INFO: Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
2025/06/01 (15:36:45) | INFO: Failed to load librocm_smi64.so, AMD GPUs will not be detected: librocm_smi64.so.6: cannot open shared object file: No such file or directory
2025/06/01 (15:36:45) | DEBUG: Failed to find Intel GPU sysfs path, Intel GPUs will not be detected
2025/06/01 (15:36:45) | DEBUG: Shared::init() : Initialized.
2025/06/01 (15:36:45) | DEBUG: Loading theme file: /home/david/.config/btop/themes/stylix.theme
2025/06/01 (15:36:47) | DEBUG: Writing new config file
2025/06/01 (15:36:47) | INFO: Quitting! Runtime: 00:00:02
```

deckstose:
Closes: https://github.com/aristocratos/btop/issues/1171

### DIFF
--- a/src/btop_cli.cpp
+++ b/src/btop_cli.cpp
@@ -124,6 +124,9 @@ namespace Cli {
 				} catch (std::invalid_argument& e) {
 					error("Preset must be a positive number");
 					return OrRetCode { 1 };
+				} catch (std::out_of_range& e) {
+					error(fmt::format("Preset argument is out of range: {}", arg.data()));
+					return OrRetCode { 1 };
 				}
 				continue;
 			}
@@ -140,6 +143,9 @@ namespace Cli {
 					cli.updates = refresh_rate;
 				} catch (std::invalid_argument& e) {
 					error("Update must be a positive number");
+					return OrRetCode { 1 };
+				} catch(std::out_of_range& e) {
+					error(fmt::format("Update argument is out of range: {}", arg.data()));
 					return OrRetCode { 1 };
 				}
 				continue;
--- a/src/btop_cli.cpp
+++ b/src/btop_cli.cpp
@@ -136,7 +136,7 @@ namespace Cli {
 
 				auto arg = *it;
 				try {
-					auto refresh_rate = std::min(std::stoul(arg.data()), 100UL);
+					auto refresh_rate = std::max(std::stoul(arg.data()), 100UL);
 					cli.updates = refresh_rate;
 				} catch (std::invalid_argument& e) {
 					error("Update must be a positive number");


src/btop_cli.cpp
```diff
@@ -136,7 +136,7 @@ namespace Cli {
 
 				auto arg = *it;
 				try {
-					auto refresh_rate = std::min(std::stoul(arg.data()), 100UL);
+					auto refresh_rate = std::max(std::stoul(arg.data()), 100UL);
 					cli.updates = refresh_rate;
 				} catch (std::invalid_argument& e) {
```
aristocratos, line 144: While you're at it, you should add exception handling for `std::out_of_range` also.
https://en.cppreference.com/w/cpp/string/basic_string/stoul

aristocratos (changes requested):

aristocratos (approved):

AnsonX10:
Sorry to necropost, but it was fun and occasionally useful to set extremely fast update times. What was the reason for limiting it to 100ms?

deckstose:
This behavior existed before this MR, it was just fixing what was missed in a refactor.

I'm not entirely sure why this _exact_ number, but we need some time to collect info and draw to the screen, which takes some amount of time. Setting the refresh rate lower will result in a poor experience

AnsonX10:
> This behavior existed before this MR, it was just fixing what was missed in a refactor.
> 
> I'm not entirely sure why this _exact_ number, but we need some time to collect info and draw to the screen, which takes some amount of time. Setting the refresh rate lower will result in a poor experience

My computer could personally handle much faster refreshes than 100 ms. It's not a bad experience. It was already limited to 100 ms when changing it via the UI or keys (as it should be), but launching it with the -u flag to set a specific ms allowed it to go lower. This was useful for power users like myself. If someone wants to open it with a flag to see what the limits of their system are, or to get more time-accurate information, you should let them. I consider this a regression in functionality, and it makes the -u flag kinda pointless.

PR #1173, opened 2025-06-02
CMake: Create a library target for intel_gpu_top code

deckstose:


aristocratos (approved):

CMakeLists.txt
```diff
@@ -63,6 +63,11 @@ add_executable(btop
   src/btop_tools.cpp
 )
 
+if(BTOP_LTO)
+  check_ipo_supported()
+  set_target_properties(btop PROPERTIES INTERPROCEDURAL_OPTIMIZATION ON)
+endif()
+
```
khaneliman, line 70: Is this not a bug? You went from checking if it was supported before enabling to just enabling regardless. We are seeing a build failure https://github.com/NixOS/nixpkgs/pull/422218 and wanted to double check before I pass a flag to explicitly disable `BTOP_LTO`

deckstose: It's not enabled regardless since `check_ipo_supported` terminates CMake if LTO is unsupported. 

Yes, on platforms with no LTO you now must explicitly disable LTO via `BTOP_LTO`.

PR #1185, opened 2025-06-25
Fix auto-detection of CPU temp on Ampere boards

bexcran:
Ampere boards use the APM X-Gene Hardware Monitoring driver. The label the driver returns is "Soc Temperature" which the current code doesn't detect. Fix that by adding a special case for it. This prevents it from using the "CPU Power" sensor and reporting the CPU tempature as over 8000C (which is the power reported in mW).

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -445,7 +445,7 @@ namespace Cpu {
 
 						found_sensors[sensor_name] = Sensor { fs::path(basepath + "input"), temp, crit };
 
-						if (not got_cpu and (label.starts_with("Package id") or label.starts_with("Tdie"))) {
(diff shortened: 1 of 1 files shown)


src/linux/btop_collect.cpp
```diff
@@ -449,6 +449,10 @@ namespace Cpu {
 							got_cpu = true;
 							cpu_sensor = sensor_name;
 						}
+						else if (not got_cpu and (label.starts_with("SoC Temperature"))) {
```
deckstose: Why can't we add this to the branch above?

bexcran: I wasn't sure about increasing the complexity of that line.
I've updated the commit to just put it on the same line.

imwints (approved):

deckstose:
Thanks you!

PR #1218, opened 2025-08-06
feat: display container engine

deckstose:
These are just a couple of container engines I personally use. It should be pretty easy to add new ones if request. This could also be expanded to include VMs.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -19,6 +19,7 @@ tab-size = 4
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <iterator>
 #include <ranges>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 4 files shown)


src/btop_draw.cpp
```diff
@@ -599,6 +600,11 @@ namespace Cpu {
 			Input::mouse_mappings["-"] = {button_y, x + width - (int)update.size() - 7, 1, 2};
 			Input::mouse_mappings["+"] = {button_y, x + width - 5, 1, 2};
 
+			// Draw container engine name
+			if (detect_container().has_value()) {
+				fmt::format_to(std::back_inserter(out), "{}{}{}{}{}", Mv::to(button_y, x + 28), title_left, Theme::c("title"), detect_container().value(), title_right);
+			}
```
aristocratos: The `detect_container()` function should be run during the collect cycle since it file operations, not during draw.

deckstose:
I've moved the call to detect the container into `Shared::init`. I don't see any benefit in repeatedly checking for the relevant files. The container state shouldn't change in btop's run time.

Issue #1151, opened 2025-05-11
[BUG] btop shows all host CPUs in an LXC container, ignoring the container’s cpuset mask


eivamu:
### Describe the bug

When running btop inside an unprivileged Debian 12 LXC on Proxmox, btop displays all host CPU cores—even though the container is pinned to a limited set of cores via its cpuset cgroup. Other tools (htop, top) correctly hide or mark the disallowed cores.

### To reproduce

- Configure an unprivileged LXC (system container) with fewer CPU cores than the host
- Install Linux; compile and install btop
- Confirm cpuset mask: `cat /sys/fs/cgroup/cpuset.cpus.effective`
- Confirm process affinity `grep Cpus_allowed_list /proc/self/status`
- Start btop
- Observe that btop shows all CPUs, or at least more CPUs than the ones listed from the commands above; the unused CPUs stay at a reported 0% load

### Expected behavior

- btop should respect the container’s cpuset cgroup (or `sched_getaffinity()`), displaying only the allowed CPUs—or at minimum marking the others as offline—similar to htop and top.
- The aggregate load, by default displayed above the CPU cores in btop, should aggregate only from active CPU cores—at the moment it is unclear for the user whether this is the case.

### Screenshots
_(none)_

### Info

- btop++ version: 1.4.3+3f1ccd9
- Binary: self-compiled from GitHub “master”
- Architecture: x86_64 (uname -m)
- Platform: Linux (Debian 12 LXC on Proxmox)
- Kernel: 6.8.12-8-pve (uname -r)
- Terminal: _not relevant_
- Font: _not relevant_

### Additional context

- Container’s cpuset: 8,10,13,36
- Process affinity matches the cpuset.
- btop currently reads /proc/stat and renders every cpuN, without filtering by cgroup mask (or so it seems)

**Hypervisor config**

- CPU: 2x Intel E5-2650L (total of 20/40 cores/threads), disk: local ssd
- OS: Proxmox VE 8.4.1 / Debian 12 "bookworm"
- Kernel: 6.8.12-8-pve

**LXC config**

- CPU: 4 cores, memory: 8192, swap: 0, unprivileged: 1
- OS: Debian 12 "bookworm"
- Kernel: 6.8.12-8-pve

dom6770:
Yes, this is also the case for me.

But the memory statistic is also wrong. The values for Total and Free are correct, but the Used, Available, Cache values are completely wrong:

[...]

PR #1220, opened 2025-08-06
feat: respect effective cpus from cgroup

deckstose:
Parse `/sys/fs/cgroup/cpuset.cpus.effective` and highlight only active cpus with color in cpu widget.

Closes: https://github.com/aristocratos/btop/issues/1151

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -855,26 +855,31 @@ namespace Cpu {
 	#endif
 		max_row -= n_gpus_to_show;
 
+		auto is_cpu_enabled = [&cpu](const std::int32_t num) -> bool {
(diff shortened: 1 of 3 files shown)


src/btop_draw.cpp
```diff
@@ -819,26 +819,32 @@ namespace Cpu {
 
 		} catch (const std::exception& e) { throw std::runtime_error("graphs, clock, meter : " + string{e.what()}); }
 
+		auto active_cpus = detect_active_cpus();
+		auto is_cpu_enabled = [active_cpus](const std::int32_t num) {
+			return std::ranges::find(active_cpus, num) != active_cpus.end();
+		};
```
aristocratos: This should be done during collect not draw, so no extra latency is added during the call to draw. (Otherwise this would be called repeatedly during a window resize.)

deckstose:
ci is a gcc 11 issue

PR #1227, opened 2025-08-16
feat: display CPU power draw & fix GPU+load avg overwriting core info

Tom94:
This change was based on YuriiShkrobut's initial PR from here: https://github.com/aristocratos/btop/pull/595

Not only adds CPU power draw display, but also fixes missing CPU core information when there are too many cores for btop to display. Previously, the bottom lines were hidden by GPU(s) as well as by the load average.

### Screenshots at min width of 80

Before: (note missing cores 23, 47, and 70 upwards)
image

After:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,6 +334,7 @@ setcap:
 	@printf "\033[1;97mFile: $(DESTDIR)$(PREFIX)/bin/btop\n"
 	@printf "\033[1;92mSetting capabilities...\033[0m\n"
 	@setcap cap_perfmon=+ep $(DESTDIR)$(PREFIX)/bin/btop
+	@setcap cap_dac_read_search=+ep $(DESTDIR)$(PREFIX)/bin/btop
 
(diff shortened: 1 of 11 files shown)


Tom94:
Just updated the PR with a quick tweak (that I feel stupid for not thinking about earlier): CPU power draw is now displayed to the right of the top line, right after temperature. Makes it consistent with the way the GPU is displayed below. I updated the images in the OP accordingly.

aristocratos (changes requested):
The load average clips outside the box in the smallest possible configuration (only CPU visible and 60 cols wide), se below.
Otherwise looks good!

image

Tom94:
Thank you, fixed by changing "Load average:" to "Load avg:".

I didn't realize it could go as narrow as 60 cols by only showing cpu, whoops. Will make sure to test this in future contributions!

aristocratos (approved):
Awesome, thank you for contributing!

PR #1239, opened 2025-08-27
test

imwints:


deckstose:
@aristocratos This is just an experiment, but do you have any thoughts?

aristocratos:
Hey, I'm assuming you are asking what my thoughts are about replacing the musl builds with this?

To start of, static linking with glibc is strongly discouraged, see examples to as why: https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged

The reason we are currently using musl for the statically linked binaries is because of maximum compatibility. They will literally run on any linux system with a 2.6.39 kernel or newer.
That would not be true for glibc, and will likely create more issues instead.

If you want to upgrade the musl builds I suggest you look for cross platform toolchains based on Alpine, because I do believe the tools themselves are updated on Alpine, but our specific toolchain haven't been updated in a long time.

deckstose:
> Hey, I'm assuming you are asking what my thoughts are about replacing the musl builds with this?

Correct.

The `muslcc` images aren't updated in years and rely on an unsupported version of GCC.
I've looked around but nothing is prebuilt __and__ has such broad platform support as `muslcc` (although we build for some rather exotic architectures, maybe without any users).

deckstose:
Ready to go. I intend to add automated updates via renovate, but this doesn't work right now and can wait for now.

aristocratos (changes requested):
Please also add the i586 and i686 builds.

Otherwise, nice!

aristocratos (approved):

PR #1242, opened 2025-08-28
btop_menu: allow `k` and `j` key in help menu for up & down

botantony:
This should enable Vi bindings in help menu


deckstose:
That's already working if you enable Vim keybindings

botantony:
it works in options menu but not help page (the one you open by pressing `?`)

deckstose:
Interesting, didn't know you can scroll there...

PR #1260, opened 2025-09-21
Fix missing application entry in the snap distribution

brlin-tw:
This patch implements the necessary changes to allow the application desktop entry and icon to appear in the snap distribution user's desktop menu:

image

The SVG variant of the application is not used as currently it'll not rendered properly in the KDE Plasma desktop environment.

Refer-to: 448234 – Usage of Qt SVG renderer causes some 3rd-party app icons to be mis-rendered <https://bugs.kde.org/show_bug.cgi?id=448234>
Signed-off-by: 林博仁(Buo-ren Lin) <buo.ren.lin@gmail.com>

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,6 +25,7 @@ package-repositories:
 apps:
   btop:
     command: usr/local/bin/btop
+    desktop: usr/local/share/applications/btop.desktop
     environment:
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -62,3 +63,10 @@ parts:
     override-pull: |
       snapcraftctl pull
       snapcraftctl set-version "$(git describe --tags | sed 's/^v//' | cut -d "-" -f1)"
+
+    override-build: |
+      snapcraftctl build
+
+      # Patch icon path of the desktop entry
+      # WORKAROUND: Use PNG icon until KDE Plasma can correctly render the SVG one
```
deckstose, line 71: Please add a line `Bug: ` to the KDE bug tracker.

deckstose (commented):
Please put a

snap/snapcraft.yaml
brlin-tw: The requested change has been added.

deckstose (approved):

PR #1265, opened 2025-09-21
Fix incorrect list item paragraph indentation in the README

brlin-tw:
Refer to the following screenshots for the addressed problem(BEFORE/AFTER):

<div>before&nbsp;&nbsp;&nbsp;&nbsp;after</div>


deckstose:
Thanks!

Issue #1226, opened 2025-08-14
[REQUEST] Option to start with all processes collapsed


code-IM-perfect:
**Is your feature request related to a problem? Please describe.**

The tree view is very helpful because it bundles the plethora of subprocesses that plague the process list.

However I always find myself collapsing all the subprocesses of systemd as they are usually full applications and it isn't really meaningful to see their children because they take up so much space and I'm not able to take a bird's eye view of how much resources they are using overall which is what I actually want from the tree view.

**Describe the solution you'd like**

This can easily be fixed by adding a preference to start with all the process collapsed, this way I only need to expand the systemd process and all its children would already be collapsed.

Maybe even add an exception field, that way I wouldn't even have to expand systemd, but maybe that would be too much.

**Describe alternatives you've considered**

An alternative would be to add a keybind to collapse all the sub-children of the selected process, which would be helpful but maybe too specific.

**Additional context**

PR #1274, opened 2025-09-28
Tree children toggle

irabrine:
Added a keybind to collapse or expand trees of the selected process' children
i wanted this feature since i have to collapse them manually every reboot, and i saw an unengaged feature request at #1226 
it's not exactly starting them collapsed, but this way it gives people the ability to do it
i chose 'u' as the keybind with no logic whatsoever, feel free to change it to something that makes more sense.

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -402,11 +402,12 @@ namespace Input {
 						Config::set("show_detailed", false);
 					}
 				}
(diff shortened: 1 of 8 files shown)


O-ELMA:
Needs improvement, pls fix

aristocratos (approved):

PR #1275, opened 2025-09-29
fix escape character in btop_input

vandabbin:
Changing line in btop_input.cpp that compared first 2 characters of input to Fx::e. 

Instead it now compares first character to "\x1b" and that seems to work.


Korel:
@aristocratos Hello, I hope it's alright that I'm writing to a merged PR to ask this and it won't feel like a noise for you; is there a timeline for a release including this PR?

I have the same issue that this PR fixed on btop version 1.4.5 on macos version 26.0.1, tried on iterm2 and terminal. I cloned main (at the commit 0d02149cfcf5824b172c9ba83e22051580bcec31) and build manually and can confirm it works.

Issue #554, opened 2023-06-14
How does btop calculate CPU Frequency (and is it different from bpytop)?


randomcoder67:
I've noticed that in btop, the frequency will go between around 800Mhz and 2.5Ghz quite a lot. 

I tried manually getting the output of `/sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq`, adding them up and dividing by the number of threads, and it gives a more stable result. The result it gives is actually consistent with what is shown in bpytop, so I'm assuming that's how bpytop does it.  

Does btop use a different method to calculate the frequency than bpytop? I tried taking a look at the source code but I'm not familiar at all with C++ so I didn't get far. 

Apologies if this is the wrong place to ask this, I wasn't sure whether to put it here or in discussions.

aristocratos:
@randomcoder67 
Currently it tries `/sys/devices/system/cpu/cpufreq/policy0/scaling_cur_freq` first, if that fails, it parses `/proc/cpuinfo`. See code snippet below.

In both cases the cpu clock is read only from the first found cpu core.

Taking an average could be a better approach, _if_ you could trust that all cores have the same max speed. This might not be the case for example with cpu's that have efficiency cores. So even if all cores were fully loaded you wouldn't see your expected maximum core speed.

Another approach could be to find the highest core clock speed and use that, but then it might be misrepresented as if your cpu is never idling, since there is most likely always one core doing some work and then also having a higher clock speed than the rest.

Sorry, but don't really remember what my reasoning for using the first found core was (back when I wrote it), maybe that it would be more likely to not be misleading when taking all available cpu configurations in mind.

But I'm not sure what the best approach is, maybe having all three methods available to choose from?

https://github.com/aristocratos/btop/blob/e269046d3b97f717446580582ea4faa74f3f7ab0/src/linux/btop_collect.cpp#L409-L464

taifuranowar:
Can a low and high be added instead of average ? I am using an intel laptop processor and average makes it difficult to understand due to E cores. so instead of showing the frequency, it can show L: (min freq) and H: (highest frequency). In the option the current way can be as default, but the low-high can be set.
I can do a PR if you like..

PR #1277, opened 2025-09-30
feat: Introduce cpu frequency display modes

emiliopedrollo:
This allows for the user to select how the cpu frequency is calculated (relates to #554)

Options are:

 - **first**: The current way. Just fetches the frequency from the first available core;
 - **range**: Display the lowest and highest frequencies;
 - **lowest**: Display the lowest frequency;
 - **highest**: Display the highest frequency;
 - **average**: Sum all the cores frequencies and divides by the number of cores;

I don't have a device with macOS or BSD to test so I'm implementing this feature on Linux only. IF anyone with access to other OSes want to test it please do and comment here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1429,6 +1429,9 @@ base_10_sizes = False
 #* Show CPU frequency.
 show_cpu_freq = True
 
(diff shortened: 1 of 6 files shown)


emiliopedrollo:
image

PR #1282, opened 2025-10-05
Fix CPU frequency detection for LoongArch CPUs

### LINKED ISSUE #1281: Loongarch64 doesn't support cpu frequency monitoring
**Describe the bug**

Loongarch64 fails to read of /proc/cpuinfo fails to output the cpu frequency. 

**To Reproduce**

Have any Loongarch64 
ssh through konsole login
run btop and enable Show cpu freq

**Expected behavior**

Btop doesn't showcase the Cpu frequency at all. 

**Screenshots**

Image

Image

 - btop++ version: v1.4.5
 - Binary: self compiled 
 - Architecture: loongarch64
 - Platform: Linux 
 - (Linux) Kernel: Debian-6.12.32-loong64
 - Terminal used: Konsole 

**Additional Info** 

I've managed fix it by changing  if (cpufreq.peek() == 'c') to  if (cpufreq.peek() == 'C' || 'c') within line 575 btop_collect.cpp. Since a bug report haven't been made, I've decided to create an issue in order for the pull request to get reviewed.

Twopic2:
As mentioned in #1281, CPU frequency detection doesn't work on LoongArch systems because cpufreq.peek() is case sensitive and couldn't find **CPU MHz**. This PR improves the fallback `/proc/cpuinfo`. 

**Changes:**
- Modified `get_cpuHz()` function in `src/linux/btop_collect.cpp` to lines 576
- Added support for "CPU MHz" field in addition to "cpu MHz"
- Tested on Loongson 3A6000 system

**Testing:**
- CPU frequency now displays correctly on LoongArch
Screenshot_20251005_012546

Fixes #1281

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -637,7 +637,7 @@ namespace Cpu {
 				if (cpufreq.good()) {
 					while (cpufreq.ignore(SSmax, '\n')) {
 						// peek is caps sensitive so it was skipping 'CPU MHz'. This aims to fix it.
-						if (cpufreq.peek() == 'c' or 'C') {
+						if (cpufreq.peek() == 'c' || cpufreq.peek() == 'C') {
 							cpufreq.ignore(SSmax, ' ');
 							if (cpufreq.peek() == 'M') {
 								cpufreq.ignore(SSmax, ':');


src/linux/btop_collect.cpp
```diff
@@ -572,7 +572,8 @@ namespace Cpu {
 				ifstream cpufreq(Shared::procPath / "cpuinfo");
 				if (cpufreq.good()) {
 					while (cpufreq.ignore(SSmax, '\n')) {
-						if (cpufreq.peek() == 'c') {
+						// peek is caps sensitive so it was skipping 'CPU MHz'. This aims to fix it.
+						if (cpufreq.peek() == 'c' or 'C') {
```
deckstose: Use `||`.

What you really want to do is `cpufreq.peek() == 'c' || cpufreq.peek() == 'C'`. Right now the second block of the evaluation is `'C'` (it will not be compared), which will always be true.

```cpp
#include <iostream>

int main() {
  if ('C') { 
    std::cout << "Hello World!\n";
  }
  return 0;
}
```

Twopic2: Okay I'll fix this right away.

Twopic2: I just pushed my changes

Twopic2:
Just curious when will this be merged?

Issue #1293, opened 2025-10-14
CPU temp sensor ariphmetic exception


fgsfds1:
**Read the README.md and search for similar issues before posting a bug report!**

**Describe the bug**

Crash on trying to choose the CPU temp sensor via the settings menu (pressing "right arrow" to get to the "correct" one.

**To Reproduce**

- Notice that the automatically chosen temperature sensor for the CPU is wrong
- Go into settings, page 2 (CPU)
- Try going through CPU sensors
- Crash on reaching sensor 13 out of 37

**Expected behavior**

Crash doesn't happen.

**Screenshots**

Not applicable.

**Info (please complete the following information):**
 - btop++ version:
   - btop version: 1.4.5
   - Compiled with: g++ (15.2.1)
   - Configured with: make STATIC= GPU_SUPPORT=true RSMI_STATIC=
 - Binary: Arch Linux repo binary
 - Architecture: x86_64
 - Platform: Linux
 - (Linux) Kernel: 6.16.10-zen1-1-zen
 - Terminal used: kitty
 - Font used: Input

**Additional context**

- lm-sensors output:
```0 % sensors
nct6687-isa-0a20
Adapter: ISA adapter
VIN0:             992.00 mV (min =  +0.00 V, max =  +0.00 V)
VIN1:             992.00 mV (min =  +0.00 V, max =  +0.00 V)
VIN2:               1.31 V  (min =  +0.00 V, max =  +0.00 V)
VIN3:             688.00 mV (min =  +0.00 V, max =  +0.00 V)
VIN4:               1.01 V  (min =  +0.00 V, max =  +0.00 V)
VIN5:             368.00 mV (min =  +0.00 V, max =  +0.00 V)
VIN6:             544.00 mV (min =  +0.00 V, max =  +0.00 V)
VIN7:               1.54 V  (min =  +0.00 V, max =  +0.00 V)
VCC:                3.30 V  (min =  +0.00 V, max =  +0.00 V)
VSB:                3.36 V  (min =  +0.00 V, max =  +0.00 V)
AVSB:               3.36 V  (min =  +0.00 V, max =  +0.00 V)
VTT:                2.03 V  (min =  +0.00 V, max =  +0.00 V)
VBAT:               3.15 V  (min =  +0.00 V, max =  +0.00 V)
VREF:               1.81 V  (min =  +0.00 V, max =  +0.00 V)
fan1:             1069 RPM  (min =    0 RPM)
[...]

PR #1294, opened 2025-10-15
Quick color div-by-zero (max_temp=0) fix

fgsfds1:
This addresses issue #1293 in a quick and dirty-ish way. I've also looked around the file and found some more instances where division by zero if the max_temp is 0 could be an issue.


fgsfds1:
Don't know if it's better to set the "default" fraction color if the max is zero to "normal" (that would be 0) or "warning" or "critical" (max clamp value of 100). Right now it's zero.

Issue #1295, opened 2025-10-16
Load average overflows into parent cpu box


rcbellN51:
OS: FreeBSD 14.3-RELEASE-p3
btop version: 1.4.5
Compiled with: c++ (18.1.6)
Configured with: gmake STATIC=

Narrow (single column) CPU core usage tables could possibly use a minimum width, or a user setting in conf. Running as a VM with only 4 cores and no temps, the CPU core usage table isn't wide enough for the "Load avg:" line. As a result the line wraps, breaking both the CPU core internal box and larger CPU box.

<img width="791" height="326" alt="Image" src="https://github.com/user-attachments/assets/620e3c47-7c12-47e8-818f-160427444648" />

I understand this is probably an issue relatively few people will ever see, so I'd understand if it just went into the "nice to have" category. Alternatively, "Load avg:" could be moved to the bottom right of the overall CPU box (opposite of uptime).

#### Solution

Provide a minimum width for rendering load average.

danie-dejager:
Came here for the same issue happening on Linux from the latest 1.4.5 snap:

<img width="1035" height="155" alt="Image" src="https://github.com/user-attachments/assets/cda57ed1-92b6-4998-9a49-4bc8e046b7eb" />
The whole CPU block is unnecessarily small. The CPU name is also cut off.

vandabbin:
I notice this issue can appear for the gpu line as well when using a smaller window size.

<img width="1132" height="468" alt="Image" src="https://github.com/user-attachments/assets/f17f158e-25ad-4cc9-bb7e-26be8d6e15ba" />

vandabbin:
@danie-dejager 

> Came here for the same issue happening on Linux from the latest 1.4.5 snap:
>
> The whole CPU block is unnecessarily small. The CPU name is also cut off.

Hey I made a fix for the load average but I'm not sure how well it will deal with a long cpu name. Would you be able to test it and see how much of the name shows up?

PR #1304, opened 2025-10-21
fix cpu box width for load average and gpu line

vandabbin:
Fixes: #1295 

cpu core box now gets a width of 29 when there is only 1 column and temp isn't shown (this leaves room for all load averages to have 2 digits before the decimal places before filling the box

gpu lines went out of bounds of the box by 2 characters with thin, but tall windows. (99x41 as an example that it occured for me) fixed this by shrinking mem and temp meters for gpu to 4 characters each when b_width is smaller then 42

I have tested this on my M1 macbook with temp sensor disabled in btop to replicate the wrapping that was happening

I tested the GPU line on my linux laptop.

I haven't been able to make this fix fail yet, but if anyone else could test this and see how it works for them that would be great.

Example of fixed cpu usage box width
btop_cpubox

Comparison of before and after fix
Screenshot 2025-10-21 at 6 51 05 AM

Example of longer gpu status line
btop_gpu_long

Example of short gpu status line
btop_gpu_short

GPU line before change
502692987-f17f158e-25ad-4cc9-bb7e-26be8d6e15ba

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -542,7 +542,7 @@ namespace Cpu {
 		bool show_watts = (Config::getB("show_cpu_watts") and supports_watts);
 		auto single_graph = Config::getB("cpu_single_graph");
 		bool hide_cores = show_temps and (cpu_temp_only or not Config::getB("show_coretemp"));
-		const int extra_width = (hide_cores ? max(6, 6 * b_column_size) : 0);
+		const int extra_width = (hide_cores ? max(6, 6 * b_column_size) : (b_columns == 1 && !show_temps) ? 8 : 0);
 	#ifdef GPU_SUPPORT
 		const auto& show_gpu_info = Config::getS("show_gpu_info");
 		const bool gpu_always = show_gpu_info == "On";
@@ -656,6 +656,9 @@ namespace Cpu {
 				gpu_temp_graphs.resize(gpus.size());
 				gpu_mem_graphs.resize(gpus.size());
 				gpu_meters.resize(gpus.size());
+	
+				// Shrink gpu graph width in small boxes to prevent line width extending past box border
+				auto gpu_graph_width = b_width < 42 ? 4 : 5;
 
 				for (size_t i = 0; i < gpus.size(); i++) {
 					if (gpu_auto and v_contains(Gpu::shown_panels, i))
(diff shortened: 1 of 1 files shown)


src/btop_draw.cpp
```diff
@@ -542,7 +542,7 @@ namespace Cpu {
 		bool show_watts = (Config::getB("show_cpu_watts") and supports_watts);
 		auto single_graph = Config::getB("cpu_single_graph");
 		bool hide_cores = show_temps and (cpu_temp_only or not Config::getB("show_coretemp"));
-		const int extra_width = (hide_cores ? max(6, 6 * b_column_size) : 0);
+		const int extra_width = (hide_cores ? max(6, 6 * b_column_size) : (b_columns == 1 and not show_temps) ? 8 : 0);
```
vandabbin: I'm not sure if the `and not show_temps` is needed here.  Testing with the machines I have shows no differences between including it and not but maybe there is some configuration that it would be needed for?

tessus:
Works great. Tested on Linux and FreeBSD.

src/btop_draw.cpp
```diff
@@ -665,11 +665,11 @@ namespace Cpu {
 					//? GPU graphs/meters
 					auto width_left = b_width - 10 - (gpus.size() > 9 ? 2 : gpus.size() > 1 ? 1 : 0);
 					if (gpu.supported_functions.temp_info and show_temps) {
-						gpu_temp_graphs[i] = Draw::Graph{ 5, 1, "temp", gpu.temp, graph_symbol, false, false, gpu.temp_max, -23 };
+						gpu_temp_graphs[i] = Draw::Graph{ b_width < 42 ? 4 : 5, 1, "temp", gpu.temp, graph_symbol, false, false, gpu.temp_max, -23 };
 						width_left -= 11;
 					}
 					if (gpu.supported_functions.mem_used and gpu.supported_functions.mem_total and b_columns > 1) {
-						gpu_mem_graphs[i] = Draw::Graph{ 5, 1, "used", safeVal(gpu.gpu_percent, "gpu-vram-totals"s), graph_symbol };
+						gpu_mem_graphs[i] = Draw::Graph{ b_width < 42 ? 4 : 5, 1, "used", safeVal(gpu.gpu_percent, "gpu-vram-totals"s), graph_symbol };
```
deckstose: The calculation should be extracted into a variable with a descriptive name and maybe a comment for better readability.

Something like 

```cpp
// Shrink gpu graph width in small boxes because of yadayadaya...
auto gpu_graph_width = b_width < 42 ? 4 : 5
```

vandabbin: Ok I will make this change now!

vandabbin: Change made

src/btop_draw.cpp
deckstose:
```suggestion
		const int extra_width = (hide_cores ? max(6, 6 * b_column_size) : (b_columns == 1 && !show_temps) ? 8 : 0);
```

We are _programming_

vandabbin: funny that's actually how I originally had it when I first made this change and i changed it to `and not` because it seemed to be what was mostly used everywhere. I will make this change now

vandabbin: change made

lcheylus:
@vandabbin Thanks for your fix : tested on OpenBSD/amd64 (Virtual machine with KVM and 4 CPU) with version 1.4.6 => no more issue with display without temp for CPU :)

PR #1306, opened 2025-10-23
feat: Processes list pausing

vandabbin:
(Well kind of. It addresses the ability to pause the process list and examine it as is mentioned in a comment on this issue. I think the background update turned off in main menu is probably sufficient for a full pause with no collection)

This change sets a keybinding 'F' that pauses the processes list. It doesn't stop collection.
It stops new processes from being added and dead processes from being removed. It also disables sorting except when the sorting mode is changed or when switching between normal and tree view modes. While paused the user can still scroll through the processes list. Detailed view can still be opened and it displays current information even when the process is dead. When pausing is enabled it displays a banner on the last line of the processes list.

I think this feature is very useful. It really helps process navigation when they aren't all jumping around. 
It allows you to capture a moment in time and see what was running.

The BSDs need testing still. I have tested a bit in a FreeBSD VM. However it doesn't seem like the BSD builds are in great shape. I noticed several bugs in the Process Box on FreeBSD that exist in v1.4.5 (before any of my changes). I will be creating issues for them soon. It is possible they are because of the very minimal VM i was testing in though. As far as I could tell my changes were working on BSD though I may have missed some edge cases. 

I've been testing a lot on Linux and macOS and I think I have ironed out any issues on those platforms. If I missed anything though feedback would be great.

Screenshot of a paused process list with open detailed view (Changed to reflect change to pause banner text)
Screenshot_20251024_104440

Video example of pausing (Pause banner text has been changed but isn't shown in this video)

https://github.com/user-attachments/assets/a0b61b93-a88c-4310-bded-08cd354c26c3

Perhaps there is situations where persistence might be useful instead in which case maybe an additional feature should be added to allow new processes to appear but keep dead processes in the list. This could be a toggle option in the options menu. Not part of this PR though unless consensus is it should be added.


src/linux/btop_collect.cpp
```diff
@@ -3105,9 +3109,17 @@ namespace Proc {
...
+			if (not Runner::pause_proc_list) {
+				auto eraser = rng::remove_if(current_procs, [&](const auto& element){ return not v_contains(found, element.pid); });
+				current_procs.erase(eraser.begin(), eraser.end());
+			}
+			//? Reset cpu usage for dead processes if paused
+			else {
+				for (auto& r : current_procs) {
+					if (rng::find(found, r.pid) == found.end()) r.cpu_p = 0;
+				}
+			}
```
vandabbin: Currently I am not resetting the memory display of dead processes to 0 like I am with CPU usage.

I think it could be useful to see how much memory a process was using if it runs and disappears really quickly. And pausing lets you do that. If you think that it should be shown accurately though I can add that in.

src/linux/btop_collect.cpp
```diff
@@ -3105,9 +3110,17 @@ namespace Proc {
...
+			if (not pause_proc_list) {
+				auto eraser = rng::remove_if(current_procs, [&](const auto& element){ return not v_contains(found, element.pid); });
+				current_procs.erase(eraser.begin(), eraser.end());
+			}
+			//? Reset cpu usage for dead processes if paused
+			else {
+				for (auto& r : current_procs) {
+					if (rng::find(found, r.pid) == found.end()) r.cpu_p = 0;
+				}
+			}
```
vandabbin: Currently I am not resetting the memory usage display of dead processes to 0 like I am with CPU usage

I think it could be useful to see how much memory a process was using when it died if it runs and disappears really quickly. And pausing lets you do that. If you think that it should be shown accurately though I can add that in

or if you think this should be an option in the menu i can do that too

src/btop_draw.cpp
```diff
@@ -1965,10 +1968,16 @@ namespace Proc {
...
 				+ c_color + rjust(cpu_str, 4) + "  " + end;
 			if (lc++ > height - 5) break;
+			else if (lc > height - 5 and pause_proc_list) break;
 		}
 
 		out += Fx::reset;
 		while (lc++ < height - 5) out += Mv::to(y+lc+1, x+1) + string(width - 2, ' ');
+		if (pause_proc_list) {
+			out += Mv::to(y + height - 2, x + 1) + Theme::c("proc_pause_bg") + string(width - 2, ' ')
+				+ Mv::to(y + height - 2, x + ((width - 24) / 2)) + Theme::c("title") + Fx::b + "Processes List Paused"
```
vandabbin: I can't decide if this should say "Processes List" or "Process List"
I've switched back and forth a couple times on it...

Getting some other people's opinions would be great.

If it should be "Process List", I will change it in the Help menu as well

deckstose:
This looks really nice!

src/btop_draw.cpp
deckstose: It's `Process list paused`

vandabbin: Excellent thanks for that!

I will make that change now!

vandabbin: Change is made

src/freebsd/btop_collect.cpp
```diff
@@ -1076,14 +1077,17 @@ namespace Proc {
...
 		const size_t detailed_pid = Config::getI("detailed_pid");
 		bool should_filter = current_filter != filter;
 		if (should_filter) current_filter = filter;
 		bool sorted_change = (sorting != current_sort or reverse != current_rev or should_filter);
+		bool tree_mode_change = tree != is_tree_mode;
 		if (sorted_change) {
 			current_sort = sorting;
 			current_rev = reverse;
 		}
+		if (tree_mode_change) is_tree_mode = tree;
```
deckstose, line 1094:
```suggestion
		if (tree_mode_change) {
		    is_tree_mode = tree;
		}
```

Always include braces for readability.

src/freebsd/btop_collect.cpp
```diff
@@ -1076,14 +1077,17 @@ namespace Proc {
 		auto per_core = Config::getB("proc_per_core");
 		auto tree = Config::getB("proc_tree");
 		auto show_detailed = Config::getB("show_detailed");
+		auto pause_proc_list = Config::getB("pause_proc_list");
```
deckstose: Guess this can be `const`?

src/freebsd/btop_collect.cpp
```diff
@@ -1128,7 +1132,8 @@ namespace Proc {
 				//? Check if pid already exists in current_procs
 				bool no_cache = false;
 				auto find_old = rng::find(current_procs, pid, &proc_info::pid);
-				if (find_old == current_procs.end()) {
+				//? Only add new processes if not paused
+				if (find_old == current_procs.end() and not pause_proc_list) {
```
deckstose, line 1140:
```suggestion
				if (find_old == current_procs.end() && !pause_proc_list) {
```

as in the other PR.

src/freebsd/btop_collect.cpp
```diff
@@ -1186,9 +1191,17 @@ namespace Proc {
 				}
 			}
 
-			//? Clear dead processes from current_procs
-			auto eraser = rng::remove_if(current_procs, [&](const auto &element) { return not v_contains(found, element.pid); });
-			current_procs.erase(eraser.begin(), eraser.end());
+			//? Clear dead processes from current_procs if not paused
+			if (not pause_proc_list) {
+				auto eraser = rng::remove_if(current_procs, [&](const auto &element) { return not v_contains(found, element.pid); });
```
deckstose:
```suggestion
				auto eraser = rng::remove_if(current_procs, [&](const auto& element) { return not v_contains(found, element.pid); });
```

Left aligned reference symbols, please.

src/btop_input.cpp
```diff
@@ -327,7 +327,10 @@ namespace Input {
 					Config::flip("proc_tree");
 					no_update = false;
 				}
-
+				else if (is_in(key, "F")) {
+					Config::set("pause_proc_list", Config::getB("pause_proc_list") ? false : true);
```
deckstose: There is `Config::flip(const std::string& name)` to flip bool configs.

src/btop_draw.cpp
```diff
@@ -1965,10 +1968,16 @@ namespace Proc {
...
 			if (lc++ > height - 5) break;
+			else if (lc > height - 5 and pause_proc_list) break;
 		}
 
 		out += Fx::reset;
 		while (lc++ < height - 3) out += Mv::to(y+lc+1, x+1) + string(width - 2, ' ');
+		if (pause_proc_list) {
+			out += Mv::to(y + height - 2, x + 1) + Theme::c("proc_pause_bg") + string(width - 2, ' ')
+				+ Mv::to(y + height - 2, x + ((width - 22) / 2)) + Theme::c("title") + Fx::b + "Process list paused"
+				+ Fx::reset;
```
deckstose: If you feel comfortable please use `fmt::format` for new draw code.

Example:

```cpp
#include <iterator>

#include <fmt/format.h>

fmt::format_to(std::back_inserter(out), "{}{}{}", Fx::bold, "Hello World!", Fx::reset);
```

Here is the [syntax spec](https://fmt.dev/12.0/syntax/).

src/btop_draw.cpp
```diff
@@ -1599,11 +1600,13 @@ namespace Proc {
 		auto mem_bytes = Config::getB("proc_mem_bytes");
 		auto vim_keys = Config::getB("vim_keys");
 		auto show_graphs = Config::getB("proc_cpu_graphs");
+		auto pause_proc_list = Config::getB("pause_proc_list");
```
deckstose: Guess this can be `const`?

vandabbin:
@deckstose 
I saw the issue I had linked was removed. should I create a new issue to link with this PR or is it fine as is?

deckstose:
> I saw the issue I had linked was removed. should I create a new issue to link with this PR or is it fine as is?

It's fine, it's just not exactly as in the issue.

aristocratos:
> On first sight, again please remove the `and or not` stuff.

@deckstose
https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md

`and or not` is fine, since it's the convention used when I originally wrote btop.

aristocratos:
> This looks really nice!

Agreed!

src/freebsd/btop_collect.cpp
vandabbin: > Left aligned reference symbols, please.

I will make this change. that is a line I didn't personally write though. I just wrapped a if statement around it. but I will still do that change

src/btop_input.cpp
vandabbin: Oh neat! I missed that! I will use that instead

src/freebsd/btop_collect.cpp
aristocratos: I disagree.

src/freebsd/btop_collect.cpp
deckstose: It's fine for me that if you touch the code you can format it. I'll try to compile a set of rules in a `.clang-format` file, however I have problems getting it merged ://

deckstose:
> and or not is fine, since it's the convention used when I originally wrote btop.

I know but it reminds me of python 🥶

aristocratos:
> > and or not is fine, since it's the convention used when I originally wrote btop.
> 
> I know but it reminds me of python 🥶

Ok, that just makes me more confused. If you already know, why are you trying to enforce your own style (not backed by the repo) on other peoples PR's?

deckstose:
I thought it was _optional_ to use them, not a requirement. But I'll refrain from doing so in the future.

vandabbin:
> It's not inherently the problem of this PR but I don't like the code duplication in all the `btop_collect.cpp` files. Maybe we need to figure out a way of combining the code in `btop_shared.cpp`

I also found this to be a bit frustrating... but at least for the most part i can modify changes i made to all of them at once with multi-file search and replace lmfao

aristocratos:
> I thought it was _optional_ to use them, not a requirement.

They are optional yes. But when you request them to be changed this doesn't seem to be the case for someone submitting a PR, no?

> But I'll refrain from doing so in the future.

Thanks.

src/btop_draw.cpp
vandabbin: so something like this?

```
fmt::format_to(std::back_inserter(out), "{}{}{}{}{}{}{}{}",
				Mv::to(y + height - 2, x + 1), Theme::c("proc_pause_bg"), string(width - 2, ' '),
				Mv::to(y + height - 2, x + ((width - 22) / 2)), Theme::c("title"),
				Fx::b, "Process list paused", Fx::reset);
```

vandabbin: or like this

```
fmt::format_to(std::back_inserter(out), "{}{}{}{}{}{}Process list paused{}",
			Mv::to(y + height - 2, x + 1), Theme::c("proc_pause_bg"), string(width - 2, ' '),
			Mv::to(y + height - 2, x + ((width - 22) / 2)), Theme::c("title"),
			Fx::b, Fx::reset);
```

deckstose: Exactly.

You can also get rid of the `string(width - 2, ' ')` by using the align property.

It works like

```cpp
// Means: '<' align right, '{}' is the place holder for the width variable.
// And the fill char is " " (a space)
fmt::format_to(std::back_inserter(out), "{:<{}}", " ", width - 2); 
```

vandabbin:
@deckstose 
@aristocratos 

I made most of those changes. I left the `and, not, and or` stuff in place since it seems to be ok either way. And I also left the `if (tree_mode_change)`

I am not positive if I used the `fmt::format_to` correctly

also I will squash down the changes to 1 commit once I know these review changes I made are good

src/btop_draw.cpp
vandabbin: awesome! thanks!

deckstose:
> I am not positive if I used the fmt::format_to correctly

It works :)

vandabbin:
> One small thing is the left-right alignment of the pause message.
>
> On the left the bar aligns with the border, on the right it aligns with the scroll bar. Maybe we center the bar inside the proc box by moving the offset of by one?

Ahh thanks for catching that. I moved the offset over by 1

deckstose:
No I was wrong. If you scroll down in the proc list the bar of the selected process has the same position and width of your previous implementation, and looks off if we change that. That's why I removed the comment.

src/linux/btop_collect.cpp
vandabbin: @deckstose 
@aristocratos 

Did either of you have thoughts on this?

vandabbin:
> No I was wrong. If you scroll down in the proc list the bar of the selected process has the same position and width of your previous implementation, and looks off if we change that. That's why I removed the comment

@deckstose 
Oh! well you were still kind of right. When I changed the text from `Processes` to `Process` I misaligned the text by 1 to the left. its fixed now haha

src/linux/btop_collect.cpp
deckstose: That's the question of what `freezing` means to you. I'd say zero out the stats so it's obvious the process died. But I can see why someone would want to see what the usage was before it died. What is bottom doing?

vandabbin: I'm not sure what bottom does as I haven't used it. But htop keeps updating the cpu bars and other info in the top and fully freezes the process list so both cpu usage and mem usage show as  they were.

however in htop you cant open a detailed view and see the `DEAD` status so you don't have any idea at all about whether the process is still  running or not.

The primary reason I decided to reset the cpu usage of dead processes was that for some reason when testing with stress-ng as a quick running process if i paused it and then the process died. If i then switched to tree mode while still  paused the tree sort was broken and the stress-ng process appeared at the top above its parents in cpu direct sort mode (with cpu usage of dead processes reset however, the tree sort works properly and the stress-ng process was properly under its parents in the tree. 

In this screenshot both stress-ng processes are dead. 
Screenshot_20251025_064900

here is an example of that (this is with cpu usage not being reset)
Screenshot_20251025_070300

I don't really understand the tree generation code well enough to really get why that is happening

vandabbin:
So there are definitely still a few things I would like to iron out.

- I would like to improve the accuracy of detailed view for dead processes.

- Elapsed time isn't accurate if the detailed view isn't open when the process dies. instead it updates everytime it is opened. I'd like to see if I can fix this. If I can't figure out how to make it reflect the accurate elapsed time then I will make it so it doesn't update elapsed time for dead processes. 

- I'd like to iron out why the tree sorting breaks when cpu usage isn't reset. Even if we keep it resetting the cpu usage and decide to reset the memory usage too. I would like to understand it before making that decision.

- I think I need to go through another round of testing to make sure there isn't more things that I missed.

src/btop_draw.cpp
vandabbin: oh....now I understand why you want me to use that. I could do the centering of the text with that too! I'm making that change now. haha

vandabbin:
```
fmt::format_to(std::back_inserter(out), "{}{}{}{}{:^{}}{}", 
	Mv::to(y + height - 2, x + 1),               
	Theme::c("proc_pause_bg"), Theme::c("title"),           
	Fx::b, "Process list paused", width - 2,                
	Fx::reset);                                             
```

I like this much better

I should have spent more time reading that syntax link you sent earlier haha

aristocratos: @vandabbin 
You can also use named arguments like below if it becomes a lot of "brackets" to keep track of (or some arguments are reused).

https://github.com/aristocratos/btop/blob/485a52687a85fdfe2f428742c2ef52627adb4f40/src/btop.cpp#L146-L162

vandabbin:
@aristocratos 
@deckstose 

I have made the fixes for the issues found.

- For the tree sorting I had missed a block of code in the tree generation that was removing parent pids of  dead processes

- When cpu usage wasn't reset and tree sorting was working properly now, if proc_aggregate was set then the cpu usage of the parent processes of a dead process would just keep increasing.

- Detailed view now shows a mostly accurate elapsed time for dead processes (inaccuracy depends on update time)

- Detailed view also now show the cpu usage perentage of the dead process if the option to keep usage is set.

- Fixed the potential for  a dead process to have its stats overwritten by a new process with the same pid (I was never able to actually make this happen but I assume it *could* have happened)

- I have also added in a menu option to enable or disable the clearing of cpu usage and mem usage for dead processes

I haven't run into any other issues with my testing so far.

vandabbin:
> > It's not inherently the problem of this PR but I don't like the code duplication in all the `btop_collect.cpp` files. Maybe we need to figure out a way of combining the code in `btop_shared.cpp`
> 
> I also found this to be a bit frustrating... but at least for the most part i can modify changes i made to all of them at once with multi-file search and replace lmfao

Haha! This just bit me. Had a single character on a line that shouldn't have been there and it got copied to all three BSDs

vandabbin:
So I've written the code to add a 'Persistence Mode'

It basically just uses the code I wrote for pausing. The only difference between it and pausing is that it doesn't disable sorting and new processes are added to the list. I am not sure if I should add it too this PR or not. I think it could potentially break if a new process is assigned the pid of a dead process still in the list so I think it's probably best not to add it unless I figure that out.

The benefit to the persistent mode is that if you are still having trouble catching a quick running process with pause mode (since it has to be on the list when you pause) then persistent mode is a nice way to be able to "capture new processes" and then you could switch to pause mode or something like that

vandabbin:
@aristocratos 
Is it ok for me to `#include <unordered_set>` in the mac and bsd versions of btop_collect? I can do it a different way if required.

aristocratos:
> Is it ok for me to `#include <unordered_set>` in the mac and bsd versions of btop_collect? I can do it a different way if required.

That's no problem, it's STD.
Only pulling in external libraries not already included would warrant a discussion.

aristocratos:
@vandabbin 
Is this completed?
You requested a review, but looks like you were still working on the PR after that.

vandabbin:
Hi! Yes this is complete now. I did request that review earlier than I should have. My apologies.

aristocratos:
Np :)

It's a great feature.
Thanks for contributing!

vandabbin:
@aristocratos 
I'm not sure if this is needed. But I had a sudden realization that maybe it wasn't a good idea for me to leave this this way?

```
				//? Check if pid already exists in current_procs
				auto find_old = rng::find(current_procs, pid, &proc_info::pid);
				bool no_cache{};
				//? Only add new processes if not paused
				if (find_old == current_procs.end() and not pause_proc_list) {
					current_procs.push_back({pid});
					find_old = current_procs.end() - 1;
					no_cache = true;
				}
				else if (dead_procs.contains(pid)) continue;

				auto& new_proc = *find_old;
```

`find_old` is still pointing to `current_procs.end()` when it is assigned to new_proc for new processes while paused.

a little bit further down in the code it reads the stats for that pid and assigns it the what? an imaginary element at the end of the vector? I'm not an experienced programmer so I have no idea if this is an issue or not but I just realized I should ask about it while looking at the code working on another feature I wanted to add. I don't think it is a problem as long as it isn't somehow writing to memory it shouldn't be allowed to? I've had one instance of btop running paused for a full day with no errors or anything so it feels fine to me but if this is a problem I can change it so it isn't a problem. (I already figured out how)

I would change it to

``` diff
				//? Check if pid already exists in current_procs
				auto find_old = rng::find(current_procs, pid, &proc_info::pid);
				bool no_cache{};
				//? Only add new processes if not paused
-				if (find_old == current_procs.end() and not pause_proc_list) {
- 					current_procs.push_back({pid});
-					find_old = current_procs.end() - 1;
-					no_cache = true;
+				if (find_old == current_procs.end()) {
+					if (not pause_proc_list) {
+						current_procs.push_back({pid});
+						find_old = current_procs.end() - 1;
+						no_cache = true;
+					}
+					else continue;
				}
				else if (dead_procs.contains(pid)) continue;

				auto& new_proc = *find_old;
```

Issue #1321, opened 2025-10-27
sorting by name or command with reverse turned off seems backwards


vandabbin:
This isn't really a bug, I just think it feels weird.

If you sort by name or command, I would expect the default to start with  for example A - Z not Z - A

However the default seems to be starting with Z.

for the other sorting methods it makes sense to me to have the larger values come first with reverse turned off.

its just for these 2 methods that I think flipping them is better

<img width="419" height="406" alt="Image" src="https://github.com/user-attachments/assets/122b1226-f8dc-4bca-8f3f-ae90bf820597" />

PR #1322, opened 2025-10-27
Fix?: flip name and command sorting

vandabbin:
The name , command, and user sorting methods seemed to be sorting opposite of what would be expected

If you think it should stay the way it is though I will close this PR

Closes: #1321


vandabbin:
I initially tried this on macOS and it looked better there but now trying it on linux I see this isn't necessarily the perfect solution either... maybe if it sorted them so that Capital letters were grouped with their lowercase equivalents. so a case insensitive stable sort

Though looking at htop they don't do case insensitive but it isn't as obvious because of the long command strings rather then short name column

Screenshot_20251026_230555

PR #1325, opened 2025-10-28
test: add test harness with ctest

deckstose:


deckstose:
Lacks some documentation

aristocratos:
Can you add some more details on what this does and what's being tested?
And why we need it?

deckstose:
Sure thing.

#### What is does?

It moves `int main` into a separate compilation unit, so it's possible to link everything else into a library and link against test code (with it's own `main`). This library is internal to CMake only.

#### What is being testing

At the moment, nothing. It's just the changes required to link against all of `btop` without  `main`.

A test could look like this:

```shell
btop
┣━src
┣━tests
┃ ┣━test_cpu_names.cpp
┃ ┗━CMakeLists.txt 
┗━CMakeLists.txt
```

```cmake
# tests/CMakeLists.txt

# ...

add_executable(test_cpu_names test_cpu_names.cpp)
target_link_libraries(test_cpu_names libbtop_test)

include(GoogleTest)
gtest_discover_tests(test_cpu_names)
```

```cpp
// tests/test_cpu_names.cpp
#include <gtest/gtest.h>

#include "btop_shared.hpp"

TEST(CpuName, Ryzen) {
  ASSERT_STR_EQ(trim_name("AMD Ryzen 20 1337"), "Ryzen over 9000");
  // ...
}
```

#### Why we need it

It will give (at least me) more confidence of pushing correct code, e.g. we can compile a list of currently working CPU names and their expected values after transformation and be sure that none of these were changed when adding new code to adjust for new names.

This can help when refactoring the collect code to ensure it still behaves the same.

An additional benefit is that when you think about writing tests while implementing (or even writing tests beforehand), the code you write becomes more composable since you try to make the function as independent as possible, to easy testing. Adding tests for existing function will also warrant a refactor into a direction of separation of concerns.

It would also be interesting to write a GUI test framework with this. This would require to use something like a `TerminalBackend` that writes to a buffer instead of standard out directly, so that a `TestBackend` can be injected for tests and assert on the buffer.

deckstose:
Could have written that in the merge request directly haha

Issue #778, opened 2024-02-22
[REQUEST] Provide a way to search for processes while keeping the entire process tree visible (and follow them around)


mfrischknecht:
**Is your feature request related to a problem? Please describe.**
I regularly encounter situation where I'd like to find a process in btop's process list without wanting to hide other processes. As an example, consider a situation where one would like to take a closer look at a process spawned in an SSH session, in a tmux session or within a nix (as in NixOS) build process. In these situation it's usually very useful to also have a situational awareness of the ancestors of the process under consideration.

**Describe the solution you'd like**
The solution I personally would find most intuitive would be for the "delete filter" action (i.e. when pressing the `del` key after having searched for processes) not to clear the current process selection. Say I'm filtering for `bash` and have selected one specific process -- with this change pressing `del` would simply switch the process list to displaying the entire process tree again while keeping the selected `bash` instance still scrolled into view (ideally in the middle of the list) and the selection active. It would then be rather easy for me to look throught the process' immediate surrounding subtree.

Ideally, that behavior would also work the same if I changed to a view that gets re-sorted very often (say `direct cpu`) so I could "follow" a process that is active in bursts and thus currently vanishes from view very often. This should also make killing such a process much easier in such scenarios. Also, if this behavior would also work when switching between sorting strategies, one could more easily have a look at multiple aspects of the same process (say: cpu, memory, number of threads) without "losing" it.

**Describe alternatives you've considered**
As an alternative, I can imagine having a similar behavior tied to having the "info" bar of a process active could also work well, i.e. if one selects a process and presses `enter`, the info bar is displayed and the process is "selected" (i.e. marked) with a different color than the "normal" selection. The process could then stay similarly scrolled into view as described above _unless the user selects another process with the arrow keys_ (the selection vanishing upon a press on `enter` is already the current behavior). Having essentially two different selection types might make navigating frequently changing views like `direct cpu` simpler, as "following around" a process would still be a more concious (opt-in) decision (and opening a process' info already indicates interest in that specific process).

**Additional context**

gabyx:
This would be really welcome!

aspiers:
Yes great proposal, and this would be tremendously useful. The lack of solutions for this and #192 prevent btop from being really useful for process tree navigation, which is a real missed opportunity.

vandabbin:
Planning to work on this after I get a good solution for #192

No idea when I will get to this though

PR #1326, opened 2025-10-28
Feat: Process following

vandabbin:
Closes: #778 

Updated to reflect recent changes: 

- Added menu options to automatically enable process following for selected process from Detailed View

- Changed Pause keybinding to 'u'
- Changed tree-children-toggle (https://github.com/aristocratos/btop/commit/f8b42175a6590098b8b9c9a754413009e9e62ea8) keybinding to 'C'

- Added keybinding 'F' (Shift+F) to follow selected process

- Added Follow and Pause buttons

- Processes are followed through sorting, tree mode, and reverse mode changes.

- Process following disengages if the process is no longer in the list.
  - Either because it died or it was filtered out

- Changing the selection exits following mode unless the list is paused.
  - While paused, changing sorting method returns the selection to the followed process.
  - Unpausing also returns the selection to the process being followed.
  - Opening a new detailed view will change the process being followed if menu option for auto following detailed view is set.

- Pause banner has been upgraded to a Pause and Following banner.
  - Text on banner reflects current state.
  - Banner background color changes depending on mode.
  - Currently:
    - Red Banner (Process list paused)
    - Blue Banner (Following process)
    - Purple Banner (Paused list and following process)

**Detailed view following example** Outdated

https://github.com/user-attachments/assets/83944729-96c1-4efa-bee6-0be90a1ce0d8

I also added a proc_banner_fg theme attribute in case these properties are ever added to any of the themes and the title attribute's color wouldn't work.

Modifying and creating themes is outside my wheelhouse however

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -451,12 +451,12 @@ namespace Input {
 						}
 					}
 				}
-				else if (is_in(key, "+", "-", "space", "u") and Config::getB("proc_tree") and Config::getI("proc_selected") > 0) {
+				else if (is_in(key, "+", "-", "space", "C") and Config::getB("proc_tree") and Config::getI("proc_selected") > 0) {
 					atomic_wait(Runner::active);
 					auto& pid = Config::getI("selected_pid");
 					if (key == "+" or key == "space") Proc::expand = pid;
 					if (key == "-" or key == "space") Proc::collapse = pid;
-					if (key == "u")	Proc::toggle_children = pid;
+					if (key == "C")	Proc::toggle_children = pid;
 					no_update = false;
 				}
 				else if (is_in(key, "t", kill_key) and (Config::getB("show_detailed") or Config::getI("selected_pid") > 0)) {
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -185,7 +185,7 @@ namespace Menu {
 		{"Up, Down", "Select in process list."},
 		{"Enter", "Show detailed information for selected process."},
 		{"Spacebar", "Expand/collapse the selected process in tree view."},
-		{"u", "Expand/collapse the selected process' children."},
+		{"C", "Expand/collapse the selected process' children."},
 		{"Pg Up, Pg Down", "Jump 1 page in process list."},
 		{"Home, End", "Jump to first or last page in process list."},
 		{"Left, Right", "Select previous/next sorting column."},


vandabbin:
This is fine but I thought of something I want to add too it

vandabbin:
I would love some feedback if some other people could please try it and tell me what you think? Does it need any tweaks? Is there anything non-intuitive about the operation? 

Personally I prefer the detailed view following to the filter following but I think it is good to have both as options

vandabbin:
I have fixed up a couple issues that were hiding as well as removing a bit of redundant code. I think this has come together nicely. Though I still have a bit more polishing to do. 

Edit: I've gotten it as good as I can now until someone else looks at it.

vandabbin:
@aristocratos @deckstose 

Hi! Would you prefer I split these 2 things off to separate PRs or are they ok to stay in this one?

- Fixed TTY theme Pause banner background color. Was accidentally using the red foreground ansi code. not background. I didn't realize until I was modifying it for use with the following too.

- Made it so the hide button for the detailed view is active instead of greyed out when the selected detailed view process is highlighted in the process list by the selection bar.

aristocratos:
> Hi! Would you prefer I split these 2 things off to separate PRs or are they ok to stay in this one?

Please do, since they aren't related.

vandabbin:
@aristocratos @deckstose 

This is ready for review now. 

I'm pinging you more as an indicator that I am actually done making any changes to it since I went and modified it quite a few more times after the last time I said it was ready.

vandabbin:
I have a different version that allows you to move the selection off of the followed process without pausing. however with that version, you need to pause to scroll the list further in either direction from what is visible on the screen without unfollowing the process. I don't know which version would be preferable. Any input? Personally I think either way is fine so I need other opinions.

Benefit to locking the selection to the followed process that I can see:
  - Easier to unfollow a process if following a detailed view process. Just press enter without needing to make sure it is selected first. 
 
Benefit to not locking the selection to the followed process that I can see:
  - Easier to move from following one process to the other since you don't need to pause or unfollow first unless that process isn't visible.

I don't see much difference either way for the filtered process following.

here is a link to the other branch if you would like to try it out to see which you prefer
https://github.com/vandabbin/btop/tree/follow-process


https://github.com/user-attachments/assets/76db0e98-93e4-43a6-a66f-5bbb13347802

aristocratos:
@vandabbin 
Sorry for the delay, this looks good to me.
Maybe `proc_follow_detailed` should be `true` by default, seems sensible.

vandabbin:
Hi!

I can change proc_follow_detailed to true by default. 

I'll do that tonight

vandabbin:
@aristocratos 

> this looks good to me. Maybe `proc_follow_detailed` should be `true` by default, seems sensible.

I have made this change. I agree that it seems right for it to be enabled by default but didn't want to make that choice without some input from others.

aristocratos:
@vandabbin 
Ok, after having played around with it a little I have some thoughts.
Maybe the "pause" key should be changed to `u` and assigning `shift+F` to "Follow".
Then showing they key in the banners "pa[u]sed" and "[F]ollowing".
And adding buttons "pause" and "Follow" at the bottom after "Nice".
Auto follow on "enter" remains but auto follow on filtering should probably be removed when a dedicated key is assigned.

Also moving the selection (up or down key) should probably also end "following" mode.

I think that would make it a little more intuitive.

What do you think?

src/btop_draw.cpp
```diff
@@ -1621,18 +1624,51 @@ namespace Proc {
...
 		out.reserve(width * height);
 
+		//? Move current selection/view to the selected process when a process should be followed
+		if (follow_process and (not pause_proc_list or Config::getB("update_following"))) {
+			Config::set("update_following", false);
+			int loc = 1;
+			bool can_follow = false;
+			for (auto& p : plist) {
+				if (p.filtered or (proc_tree and p.tree_index == plist.size())) continue;
+				if (p.pid == (size_t)Config::getI("followed_pid")) {
```
aristocratos: `Config::getI("followed_pid")` can be assigned to a variable before the loop so we don't do the lookup inside the loop.

aristocratos (changes requested):

vandabbin:
@aristocratos 

> Ok, after having played around with it a little I have some thoughts. Maybe the "pause" key should be changed to `u` and assigning `shift+F` to "Follow". Then showing they key in the banners "pa[u]sed" and "[F]ollowing". And adding buttons "pause" and "Follow" at the bottom after "Nice". Auto follow on "enter" remains but auto follow on filtering should probably be removed when a dedicated key is assigned.

I think you are right about this. It makes it more clear how to activate the features and also how to unactivate them if they accidently activated it. 

> Also moving the selection (up or down key) should probably also end "following" mode.

Ok it is great to get an opinion on this. I think this is a good change to make. It I think more nicely solves the problem I felt I was trying to fix with that other branch I mentioned in a previous comment on this pr.

> I think that would make it a little more intuitive.
> 
> What do you think?

I appreciate your input on this. I'll get to making these tweaks happen

src/btop_draw.cpp
vandabbin: Good call. I'll make that happen

vandabbin:
@aristocratos 

I have made those changes and absolutely agree they improve intuitiveness.

aristocratos:
@vandabbin 
Awesome!

Maybe we don't need the "[F]" and "[u]" in the banners when we got the buttons on the bottom, noticed it didn't look that great.

Could also exclude the "pause" button from being greyed out when not browsing the process list, since the pause keybind is active all the time.

aristocratos:
Oh, also the button text should be all bold when the respective feature is enabled (that's how all other buttons that can be toggled work).

aristocratos:
Come to think of it, "pause" should probably be with the buttons at the top instead.
Since it's not affecting a "single" item in the list, but the whole list like the other buttons at the top.

vandabbin:
@aristocratos 

> Maybe we don't need the "[F]" and "[u]" in the banners when we got the buttons on the bottom, noticed it didn't look that great.

Yes after looking at it while using it I do think that this doesn't look as great. I have changed this back.

> Oh, also the button text should be all bold when the respective feature is enabled (that's how all other buttons that can be toggled work).

Ahh they should! I have now made them toggle between bold and regular.

> Could also exclude the "pause" button from being greyed out when not browsing the process list, since the pause keybind is active all the time. 
> Come to think of it, "pause" should probably be with the buttons at the top instead. Since it's not affecting a "single" item in the list, but the whole list like the other buttons at the top.

Ok yeah that is a great point.  I have now moved pause to the top and preventing it from becoming greyed out. It also allows the width of the process list to be slightly less while still keeping both those buttons visible. I do make them the first buttons to disappear still as the width shrinks. I think they are lower priority buttons then some of the others. Do you agree with that?


I also made a fix that if auto following the detailed view is disabled, closing the detailed view won't stop following a process already being followed

src/btop_draw.cpp
```diff
@@ -1892,10 +1941,11 @@ namespace Proc {
 
 			//? Set correct gradient colors if enabled
 			string c_color, m_color, t_color, g_color, end;
-			if (is_selected) {
+			if (is_selected or is_followed) {
 				c_color = m_color = t_color = g_color = Fx::b;
 				end = Fx::ub;
-				out += Theme::c("selected_bg") + Theme::c("selected_fg") + Fx::b;
+				const string highlight = is_followed ? Theme::c("followed_bg") + Theme::c("followed_fg") : Theme::c("selected_bg") + Theme::c("selected_fg");
```
vandabbin, line 1947: Do you think that maybe the followed process highlighting should show below the selected process highlighting instead? If done this way then the followed process would only be highlighted a different color when the list is paused and the selection is not the same as the followed process. I think I am leaning more towards this way now but a second opinion would be great. I think it only really matters while paused is being used. So it's not a big deal either way

```suggestion
				const string highlight = is_selected ? Theme::c("selected_bg") + Theme::c("selected_fg") : Theme::c("followed_bg") + Theme::c("followed_fg");
```

Screenshot of following process with selected highlighting priority over following
Screenshot 2025-12-29 at 9 47 28 PM
Screenshot of paused list, following process, with selection different from following
Screenshot 2025-12-29 at 9 47 19 PM

vandabbin:
@aristocratos 

Oh I just realized that the 'u' key is technically being used already by a feature that collapses/expands the children processes in tree mode. What do you think we should do about this? Should I change the pause key to something else or should the tree children toggle feature have it's binding changed instead?

From #1274
> Added a keybind to collapse or expand trees of the selected process' children i wanted this feature since i have to collapse them manually every reboot, and i saw an unengaged feature request at #1226 it's not exactly starting them collapsed, but this way it gives people the ability to do it i chose 'u' as the keybind with no logic whatsoever, feel free to change it to something that makes more sense.

It was introduced with this commit f8b4217 and the PR description does say that 'u' was chosen as a keybind with no logic behind it and to feel free to change it

Perhaps the tree children toggle feature could use "C" (Shift+C) instead? or "U" (Shift+U)?

aristocratos:
> Oh I just realized that the 'u' key is technically being used already by a feature that collapses/expands the children processes in tree mode. What do you think we should do about this?

Sorry, missed that.

> Perhaps the tree children toggle feature could use "C" (Shift+C) instead? or "U" (Shift+U)?

Changing tree children toggle is ok by me.

src/btop_draw.cpp
aristocratos: I think using a new color for following is actually good, since it makes it really obvious what's going on considering the banner has the same color.

But maybe the newly added theme options should be documented in the README.md so themers know that they exist.

vandabbin: Okay I do agree with that. I will leave it as is.

> But maybe the newly added theme options should be documented in the README.md so themers know that they exist.

Yes they should be.  I will see about doing that now

vandabbin:
@aristocratos 

Ok I have updated the README to include some information about the new theme attributes along with a message saying they haven't been added to any of the themes except for the default one so far.

The tree-children-toggle keybinding has been changed. I left that as a seperate commit but can squash that down if you would like. I wasn't sure if I should since it relates to a different feature

aristocratos:
> The tree-children-toggle keybinding has been changed. I left that as a seperate commit but can squash that down if you would like.

That's completely fine.

Nice work!

Will merge in a little bit.

vandabbin:
Thanks you for your input it was extremely helpful!

aspiers:
Awesome work!! Excited to try this 🚀

Issue #1327, opened 2025-10-29
In some cases when tree view is filtered, sort isn't stable?


vandabbin:
**Describe the bug**

I have reproduced this on v1.4.5 release. 

When filtering the tree view in some cases the sort isn't stable. In particular with the sorting methods that aren't sorted numerically but rather alphabetically. The sorting is perfectly stable with the full tree view and in most cases when filtering, however not all cases. 

Oddly specifically for the command sorting I could make it stable by switching to reverse mode. 

This ends up affecting the pause mode as well as when paused I am not disabling whichever part of the tree gen is messing this up. It ends up actually causing every sort method to switch around in the filtered tree view in these specific cases. With the full tree view and a lot of other filtered searches that I have tried. The pausing stops the reordering as it should.

I have been looking at the code. However, my understanding of the tree generation code is severely lacking. I have a theory it has something to do with the tree-indexing but I really don't know

**To Reproduce**

- Install or Compile v1.4.5
- Run btop and filter for something like fish or maybe bash or zsh or some other name that might have the right conditions to recreate this. 
-  switch to name, user, or command modes
- try switching reverse modes as well (specifically on command mode)

**Expected behavior**

Sorting by name, command, or user shouldn't be moving around like that.

and if whatever is causing this is resolved then hopefully without any changes the pause works in these situations too. 

**Screenshots**

- Video of btop v1.4.5 filtering for fish and switching between the sorting methods. 
-  showcases name, command, and user sorting being out of whack.

https://github.com/user-attachments/assets/3b35e116-fbc3-4624-94de-03cbfa943fd8

- Video example of how with a different filter search everything works fine even with pause

https://github.com/user-attachments/assets/1b2b29ca-ca96-4658-b63d-dfbf9d6208af

- Video example of pause being completely broken with the original filter

https://github.com/user-attachments/assets/1526d82d-f3cb-44df-96b8-6244a1aea430

PR #1328, opened 2025-10-30
Fix: Tree sort unstable when filtered

vandabbin:
Fixes: #1327 

I managed to figure it out!

changed final tree index `sort` to `stable_sort` and this seems to have fixed the issue.

It doesn't seem to have adversely affected the unfiltered tree sorting in any way. But it has fixed the issue of a filtered tree sort moving the last result up to the top and back to the bottom for `name`, `command`, and `user` sorting modes and for all modes with pause enabled

Video example of fix

https://github.com/user-attachments/assets/70cc93a9-0a48-4698-9bbc-bb1a85b36131

### DIFF
--- a/src/freebsd/btop_collect.cpp
+++ b/src/freebsd/btop_collect.cpp
@@ -1326,7 +1326,7 @@ namespace Proc {
 			}
 
 			//? Final sort based on tree index
-			rng::sort(current_procs, rng::less{}, & proc_info::tree_index);
+			rng::stable_sort(current_procs, rng::less {}, &proc_info::tree_index);
 
 			//? Move current selection/view to the selected process when collapsing/expanding in the tree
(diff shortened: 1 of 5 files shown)


src/freebsd/btop_collect.cpp
```diff
@@ -1326,7 +1326,7 @@ namespace Proc {
 			}
 
 			//? Final sort based on tree index
-			rng::sort(current_procs, rng::less{}, & proc_info::tree_index);
+			rng::stable_sort(current_procs, rng::less{}, & proc_info::tree_index);
```
deckstose:
```suggestion
			rng::stable_sort(current_procs, rng::less {}, &proc_info::tree_index);
```

vandabbin: Changed

Issue #976, opened 2024-11-18
[REQUEST] Display CPU Frequency for all individual cores?


junyuan-chen:
I often see the CPU frequency jumping between 600MHz and 5.6GHz. It turns out that the CPU frequency printed is just for the first core? Is it possible to have an option to let the frequency for all cores printed next to the core usage?

<img width="802" alt="Screenshot 2024-11-18 at 10 30 57 AM" src="https://github.com/user-attachments/assets/62e70b98-13fe-412e-9626-e7b45751fb79">
<img width="802" alt="Screenshot 2024-11-18 at 10 31 32 AM" src="https://github.com/user-attachments/assets/21d81040-1dfe-4fd1-980e-6d0d424eaaaa">

```
╰─ sudo cat /proc/cpuinfo | grep "MHz"
cpu MHz		: 600.000
cpu MHz		: 2949.487
cpu MHz		: 5582.021
cpu MHz		: 600.000
cpu MHz		: 5583.058
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 5401.742
cpu MHz		: 5407.256
cpu MHz		: 600.000
cpu MHz		: 2701.844
cpu MHz		: 5418.163
cpu MHz		: 5407.263
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 5583.116
cpu MHz		: 2843.269
cpu MHz		: 5583.066
cpu MHz		: 5583.046
cpu MHz		: 600.000
cpu MHz		: 5583.080
cpu MHz		: 5582.575
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 5407.213
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 600.000
cpu MHz		: 600.000
```

By the way, thank you for having this amazing project available!

jontrulson:
I second this... Looking at only the first core doesn't make any sense on a modern CPU.  In fact, I would prefer per-core clocks to the current per-core temps.

FlavioMili:
https://github.com/aristocratos/btop/pull/1593

if anyone wants to help for the other platforms

PR #1341, opened 2025-11-04
Add ability to view core frequencies for Linux

akhlebka:
Add ability to view core frequencies on Linux. CPU cores frequencies are hidden for each OS but Linux to prevent unsupported functionality from being activated on other platforms.
image

Added config update from settings, removed some of the magic numbers in core column width calculation.
In case when failed to get core frequency, zero is populated so we can track it and show **N/A** for a moment.

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -524,6 +524,11 @@ namespace Menu {
 				"Highest, the highest frequency.",
 				"",
 				"Average, sum and divide."},
+			{"show_cores_freq",
+				"Show each CPU core frequency.",
+				"",
+				"Can cause slowdowns on systems with many",
+				"cores and certain kernel versions."},
 		#endif
 			{"custom_cpu_name",
 				"Custom cpu model name in cpu percentage box.",
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -525,7 +525,7 @@ namespace Menu {
 				"",
 				"Average, sum and divide."},
 			{"show_cores_freq",
-				"Show each CPU core frequency.",
+				"Show each CPU core frequency next to core utilization.",
 				"",
 				"Can cause slowdowns on systems with many",
 				"cores and certain kernel versions."},
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -689,13 +689,10 @@ namespace Cpu {
 		return cpuhz;
 	}
 
-	vector<string> get_cpu_coresHz()
-	{
-		try
-		{
+	vector<string> get_cpu_coresHz() {
+		try {
 			vector<double> frequencies;
-			for (auto it = core_freq.begin(); it != Cpu::core_freq.end(); )
-			{
+			for (auto it = core_freq.begin(); it != Cpu::core_freq.end();) {
 				if (it->empty()) {
 					it = core_freq.erase(it);
 					continue;
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -723,7 +723,7 @@ namespace Cpu {
 			}
 		}
 		catch (const std::exception& e) {
-			Logger::warning("get_cpu_coresHz() : " + string{e.what()});
+			Logger::warning(fmt::format("get_cpu_coresHz() : {}", e.what()));
 			return vector<string>{};
 		}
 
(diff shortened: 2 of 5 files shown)


src/linux/btop_collect.cpp
```diff
@@ -688,6 +689,47 @@ namespace Cpu {
...
+				for (auto& freq : frequencies)
+				{
+					formated_frequencies.push_back(freq == 0.0 ? "" : normalize_frequency(freq));
+				}
+
+				return formated_frequencies;
+			}
+		}
+		catch (const std::exception& e) {
+			Logger::warning("get_cpu_coresHz() : " + string{e.what()});
```
deckstose:
```suggestion
			Logger::warning(fmt::format("get_cpu_coresHz() : {}", e.what()));
```

src/linux/btop_collect.cpp
```diff
@@ -688,6 +689,47 @@ namespace Cpu {
 		return cpuhz;
 	}
 
+	vector<string> get_cpu_coresHz()
+	{
```
deckstose: Please stick to opening braces on the same line.

akhlebka:
Closes #976

src/btop_menu.cpp
```diff
@@ -524,6 +524,11 @@ namespace Menu {
 				"Highest, the highest frequency.",
 				"",
 				"Average, sum and divide."},
+			{"show_cores_freq",
+				"Show each CPU core frequency.",
+				"",
+				"Can cause slowdowns on systems with many",
+				"cores and certain kernel versions."},
```
deckstose, line 531: Can you expand on that? Did you experience this?

akhlebka: Me personally no, I’ve sticked with the message for the original CPU frequency feature, due to some parts of the code are close to each other and the features are related.

src/linux/btop_collect.cpp
```diff
@@ -688,6 +689,44 @@ namespace Cpu {
...
+				else {
+					frequencies.push_back(core_hz);
+				}
+				++it;
+			}
+
+			if (not frequencies.empty()) {
+				vector<string> formated_frequencies;
+				for (auto& freq : frequencies)
+				{
```
deckstose:
```suggestion
				for (auto& freq : frequencies) {
```

src/linux/btop_collect.cpp
```diff
@@ -1171,6 +1215,11 @@ namespace Cpu {
 
 		cpu.active_cpus = std::make_optional(detect_active_cpus());
 
+		if (Config::getB("show_cores_freq") and not cpu_coreHz.empty())
+		{
```
deckstose:
```suggestion
		if (Config::getB("show_cores_freq") and not cpu_coreHz.empty()) {
```

src/linux/btop_collect.cpp
```diff
@@ -688,6 +689,44 @@ namespace Cpu {
...
+			for (auto it = core_freq.begin(); it != Cpu::core_freq.end();) {
+				if (it->empty()) {
+					it = core_freq.erase(it);
+					continue;
+				}
+
+				double core_hz = stod(readfile(*it, "0.0")) / 1000;
+				if (core_hz <= 1 or core_hz >= 999999999) {
+					Logger::warning("get_cpuHZ() : Failed to read /sys/devices/system/cpu/cpufreq/policy");
+					frequencies.push_back(0.0);
```
deckstose: Instead of pushing an empty value to signal a missing entry, this should be wrapped in an optional to better deliver the intent.

src/btop_menu.cpp
akhlebka: I've changed the message a bit to signal where the core frequency will be, but did not remove the part about slow down possibility.

src/linux/btop_collect.cpp
akhlebka: Updated as a part of latest comit

src/linux/btop_collect.cpp
akhlebka: Updated as a part of latest commit

src/linux/btop_collect.cpp
akhlebka: Added -1.0 as a signal and returning "N/A" right from collect() with empty check as a fallback in draw part

deckstose: This is the same but shifted by -1

akhlebka: Got what you mean, done

RafayAhmad7548:
why was this not merged?
any reason?

akhlebka:
@RafayAhmad7548 idk, it was open for 2 months. Then I deleted the repo and MR closed itself.

PR #1346, opened 2025-11-05
build(cmake): don't set a default build type

deckstose:
This was initially done to aid with fortifying source which required an optimized build. We since removed the logic do to avoid duplicating work of distribtions.

We are also much more likly to call cmake in development so we mostly don't want release builds by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,11 +22,6 @@ include(CMakeDependentOption)
 # Make our Find<Package>.cmake files available
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 
-# When the build type is not set we can't fortify
-if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


deckstose:
It was just a dirty workaround to make the manual `_FORTIFY_SOURCE` setting work.

`CMAKE_BUILD_TYPE` is not one of the variables the build system should make assumptions or preferences about. It should only be set by the user invoking CMake ([see this talk](https://www.youtube.com/watch?v=xNHKTdnn4fY)).

> So the default should probably cater to the more common case rather than the rare.

I agree, however I don't agree that the people invoking CMake are more likely to want to compile release builds all the time. These are likely only developers who want to build and test their work, not people who build the latest source release for installation.

The majority of downstream users which want release builds are distributions, however they've got their own build environments where things like `CMAKE_BUILD_TYPE` are already configured properly, so for them the default doesn't matter.

> The only reason to build a debug build would be for debugging.

Note that the default CMake build type is unset which is not the same as debug mode (with explicit debugging flags).

PR #1357, opened 2025-11-10
feat: add --themes-dir flag

adeci:
## Description

Adds `--themes-dir` command-line option to specify a custom directory for theme files.
When provided, this directory is searched before the default user and system directories.

## Motivation

The current fixed theme search paths don't work well for declarative and immutable system configurations that cannot or prefer not to place files in `$HOME`.

I am working with application wrappers that provide additional resources and configuration without modifying the package itself or writing to user directories. Currently, there's no clean way to provide custom btop themes in these scenarios since themes must be in predefined system or user paths, which would require either rebuilding the package to modify the outpath or impurely placing files in the user's home directory.

With `--themes-dir`, themes can be specified in any directory the user chooses, or bundled alongside the application and referenced directly.

## Changes

- Added `--themes-dir <dir>` CLI option with directory validation
- Custom directory takes priority in theme search order

## Usage

```bash
btop --themes-dir /path/to/custom/themes

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -910,6 +910,12 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
 		}
 	}
 
+	//? Set custom themes directory from command line if provided
+	if (cli.themes_dir.has_value()) {
+		Theme::custom_theme_dir = cli.themes_dir.value();
+		Logger::info("Using custom themes directory: " + Theme::custom_theme_dir.string());
+	}
+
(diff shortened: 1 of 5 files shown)


aristocratos (approved):

deckstose:
This is missing a man page entry.

adeci:
> This is missing a man page entry.

I'll author a man page entry and add to the README.md themes section, thanks for reminding me on this

PR #1374, opened 2025-11-23
docs(README): add mise alternative method installation

jylenhof:


deckstose:
From other similar requests:

> Thanks for contributing! I think the couple of examples are enough for user the guess that the package name is btop everywhere and just change the command from dnf or others to their distros command.

My views have changed a bit in this regard, however I think the `README.md` is too long at the moment, which doesn't help new users find what they are looking for.

I'd love to see this merged if we would split the installation instruction in a separate file like `INSTALL.md` or if we would create a github page (with something like [mkdocs material](https://github.com/squidfunk/mkdocs-material/)) with very detailed instruction for all platforms (my preferred approach). There we have a better chance to structure everything properly.

PR #1386, opened 2025-12-03
fix: check COLORTERM for truecolor before enabling TTY mode

### LINKED ISSUE #1383: [BUG] TTY mode incorrectly enabled based on device path instead of terminal capabilities
## Bug Description

btop's TTY mode auto-detection only checks if the device path starts with `/dev/tty`, ignoring actual terminal capabilities. This incorrectly forces 16-color mode on systems that support truecolor.

## Current Detection Logic

**src/btop.cpp:826-830**
```cpp
else if (Term::current_tty.starts_with("/dev/tty")) {
    Config::set("tty_mode", true);
    Logger::debug("Auto detect real TTY");
}
```

This doesn't check:
- `COLORTERM` environment variable
- `TERM_PROGRAM` or terminal capabilities
- Actual color support

## Impact

- Linux framebuffer console with virtio/DRM graphics (supports truecolor but forced to 16-color)
- SSH sessions displayed through truecolor terminals
- VMs with console piped through modern terminals (see #373)

When TTY mode is incorrectly enabled, user theme colors are completely discarded and replaced with hardcoded values in `TTY_theme`.

## Proposed Fix

Check `COLORTERM` and other terminal capability indicators before falling back to device path detection:

```cpp
static auto supports_truecolor() -> bool {
    if (const auto* colorterm = std::getenv("COLORTERM")) {
        std::string ct = str_to_lower(colorterm);
        if (ct == "truecolor" || ct == "24bit") return true;
    }
    
    if (const auto* term_program = std::getenv("TERM_PROGRAM")) {
        std::string tp = term_program;
        if (tp == "iTerm.app" || tp == "vscode" || tp == "WezTerm") return true;
    }
    
    if (const auto* term = std::getenv("TERM")) {
        std::string t = term;
        if (t.find("truecolor") != std::string::npos || 
            t.find("24bit") != std::string::npos) return true;
    }
    
    return false;
}

static auto configure_tty_mode(std::optional<bool> force_tty) {
    if (force_tty.has_value()) {
        Config::set("tty_mode", force_tty.value());
    }
    else if (Term::current_tty.starts_with("/dev/tty") && !supports_truecolor()) {
        Config::set("tty_mode", true);
    }
}
```

## Additional Issue: Themes Ignored in TTY Mode

When TTY mode is enabled (even correctly), user themes are completely ignored. The `generateTTYColors()` function replaces all user colors with hardcoded `TTY_theme` values.

**src/btop_theme.cpp:354-374**
```cpp
void generateTTYColors() {
    colors = TTY_theme;  // Overwrites user theme
}
```

This means if a user sets all box borders to the same color, they appear different in TTY mode (hardcoded green, yellow, magenta, red).

**Expected**: User theme colors should be converted to nearest 16-color ANSI codes instead of being ignored.

## Current Workaround

Use `btop +t` flag or set shell alias.

i-am-logger:
## Summary
Fixes #1383

TTY mode detection now checks for truecolor support via environment variables before falling back to 16-color TTY mode. This allows modern terminals on /dev/tty* devices (like framebuffer consoles with virtio graphics) to use full truecolor support.

## Changes
- Added `supports_truecolor()` function that checks:
  - `COLORTERM` environment variable for "truecolor" or "24bit"
  - `TERM_PROGRAM` for known truecolor-capable terminals (iTerm.app, vscode, WezTerm, Hyper)
  - `TERM` variable for truecolor indicators (truecolor, 24bit, direct)
- Modified `configure_tty_mode()` to only enable TTY mode if the device is `/dev/tty*` AND truecolor is not supported
- Added comprehensive test suite (20 tests) for truecolor detection

## Testing
All tests pass:
- COLORTERM detection (truecolor, 24bit, case-insensitive)
- TERM_PROGRAM detection (iTerm, VSCode, WezTerm, Hyper)
- TERM variable detection (xterm-truecolor, xterm-24bit, xterm-direct)
- Real-world scenarios (SSH, framebuffer, legacy terminals)


deckstose:
There are **lots** if unrelated changes. I guess they come from using AI. Closing for now until a solution for this is discussed in the linked issue.

PR #1390, opened 2025-12-04
themes: load/save by name if possible

### LINKED ISSUE #443: [REQUEST] Store built-in themes differently in config file
When sharing configuration files across different machines (and OSs), having a hard-coded path is not ideal. Having my configuration file shared between a Linux and a macOS, the only diff I have between the two is the `color_theme` entry. On Linux:
```
color_theme = "/usr/share/btop/themes/nord.theme"
```

On macOS:

```
color_theme = "/opt/homebrew/Cellar/btop/1.2.9/share/btop/themes/nord.theme"
```

(Side-wnote: what happens on macOS when btop is updated? I suspect this path becomes invalid?)

Avoid any hard-coded path. One option could be to copy/symlink the built-in themes in `$XDG_CONFIG_PATH/btop/themes` on startup or when selecting a built-in theme. Another one could be to only store the name for built-in themes rather than the full path.

N/A

**Additional context**
N/A

katexochen:
This fixes #443 and a downstream issue in nixpkgs: https://github.com/NixOS/nixpkgs/issues/460344

Short description of the latter issue: I nixpkgs, different versions of btop are installed under different paths. If a user selects a theme and it is saved to the config by its absolute path, that path will be broken after an update.

To solve this, we allow writing the theme name only into the config, and we save the theme by name if possible.

When selecting a theme, we check if it is the first (with respect to the load oder priority: custom > user > system) match for this filename in the list of available themes. If it is, we save it by filename instead of using the full path. If there is another theme with the same name and higher priority, we save using the full path.

This should also be compatible with the previous behavior of being able to load themes by name from $XDG_CONFIG_PATH/btop/themes.

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -34,6 +34,9 @@ tab-size = 4
 #include <iostream>
 #include <unordered_map>
 #include <utility>
+#include <ranges>
+#include <algorithm>
+#include <iterator>
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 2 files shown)


eclairevoyant:
This works for what it intends to do, but seems to cause 2 issues:

1. All the non-internal themes (anything outside of "Default" or "TTY") have the ID "38/37".
2. The theme order is now lost.

katexochen:
Thanks for the review @eclairevoyant, I think I fixed the issue, could you check again?

eclairevoyant:
First issue is fixed, sort order is still off. Prior to this PR, the external themes (theme # 3 onward) are sorted asciibetically. So `HotPurpleTrafficLight`, `adapta`, `adwaita-dark`, ... after this PR, the sort order is unclear to me.

katexochen:
> First issue is fixed, sort order is still off. Prior to this PR, theme 3 onward is sorted lexically. So `HotPurpleTrafficLight`, `adapta`, `adwaita-dark`, ... after this PR, the sort order is unclear to me.

Thanks, fixed by adding sorting to the theme list. I kept the order case sensitive as it was before.

eclairevoyant (commented):
Not familiar with c++, but functionality-wise lgtm

eclairevoyant:
@aristocratos mind reviewing this? Also why was my prior comment marked as spam? I built and tested the code and was _confirming that it worked as intended._

katexochen:
Rebased to solve the merge conflict.

aristocratos (approved):

aristocratos:
@katexochen Thanks for contributing!

@eclairevoyant Your review comment was marked as spam because you posted a review of C++ code where you immediately proclaimed that you don't know C++, making the review pointless.

It's ok to post that a PR you tested works.
But reviewing code when you don't know the programming language in a repository where you're not a collaborator or even contributor doesn't make much sense.

eclairevoyant:
We had a back-and-forth where the functionality did not work the first time, nor the second time, I hope you can see that.

I can read C++, it's surely not that complicated, I've been writing code in various languages for 20 years, but I don't know the acceptable coding practices in 2026 because I don't write C++ on a regular basis, so I abbreviated that as "I don't know C++", and I was simply confirming that the 3rd time worked as intended after my testing, rather than, yknow, _silence for months._ (And presumably the author appreciated the confirmation given their reaction.) It's not spam to give the 3rd review if the first 2 reviews weren't spam...

aristocratos:
@eclairevoyant Nobody has any issue with you commenting on the PR. But there is a big difference in making a comment and doing a code review, which you did on Dec 8.

The comment attached to the code review was marked as spam since you proclaimed "Not familiar with c++", which obviously would make the code review meaningless.

> [...] yknow, silence for months.

This is a non sponsored open source project i maintain in my spare time (you know for free), for anyone to to use (for free).

So I don't know, call customer support if you wanna complain...

eclairevoyant:
> But there is a big difference in making a comment and doing a code review

There really isn't lmao it's just a comment.

image


It's not like I approved the PR (I don't even have access to do so). But whatever, it's a level of pedantry I can't engage with.

PR #1393, opened 2025-12-06
Fix: humanized units shortening

### LINKED ISSUE #1371: [BUG] Incorrect conversion of swap memory size when resizing window
**Describe the bug**

Incorrect conversion of swap memory size when resizing window.

btop version is 1.4.5

**To Reproduce**

I have 10 GiB of swap memory.

Btop says I have "9.99 GiB".

When I make the window small enough to trigger the memory conversion (no spaces, fewer digits/characters), btop says I have "1.0T".

**Expected behavior**

Btop says I have "10G"

**Screenshots**

Correct:

Image

Wrong:

Image

yinghao-w:
Fixes #1371 
First commit removes some unreachable code. (Please let me know if  this should be split off into its own PR).
Second commit fixes the formatting issue. It correctly shortens `9.99` to `10` without bumping up the unit, and preserves the shortening of other numbers as before.

### DIFF
--- a/src/btop_tools.cpp
+++ b/src/btop_tools.cpp
@@ -463,16 +463,20 @@ namespace Tools {
 		}
 
 		if (shorten) {
-			auto f_pos = out.find(".");
-			if (f_pos == 1 and out.size() > 3) {
+			auto has_sep = out.find(".") != string::npos;
+			if (has_sep) {
 				out = fmt::format("{:.1f}", stod(out));
 			}
-			else if (f_pos != string::npos) {
-				out = fmt::format("{:.0f}", stod(out));
-			}
 			if (out.size() > 3) {
-				out = fmt::format("{:d}.0", out[0] - '0');
-				start++;
+				// if out is of the form xy.z
+				if (has_sep) {
+					out = fmt::format("{:.0f}", stod(out));
+				}
+				// if out is of the form xyzw (only when not using base 10)
+				else {
+					out = fmt::format("{:d}.0", out[0] - '0');
+					start++;
+				}
 			}
 			out.push_back(units[start][0]);
 		}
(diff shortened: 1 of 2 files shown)


yinghao-w:
> Nit: I'd rather have the comment say "single precision, four digits, etc" than repeat what the code already tells

Thank you for the review. I'm not too sure what exact change to the comments you would prefer? If the comments are not necessary or clutter up the code, I'll remove them if you'd prefer that.

deckstose:
I was thinking about it and it should be fine as it is. What I meant was instead of writing:

```
// if out is in the form xy.z
// ->
// Num is single precision
```

but this is equally good I guess.

I think you can store `out.find('.')` in a local variable and reuse it in the `if` blocks.

yinghao-w:
Thank you for clarifying. I'll leave the comments then if you're okay with it, and make the change.

src/btop_tools.cpp
```diff
@@ -471,15 +463,17 @@ namespace Tools {
...
+			auto f_pos = out.find(".");
+			// if out is of the form xy.z
+			if (out.size() > 3 and f_pos != string::npos) {
 				out = fmt::format("{:.0f}", stod(out));
 			}
-			if (out.size() > 3) {
-				out = fmt::format("{:d}.0", out[0] - '0');
+			// if out is of the form xyzw (only when not using base 10)
+			else if (out.size() > 3 and f_pos == string::npos) {
+				out= fmt::format("{:d}.0", out[0] - '0');
```
deckstose:
```suggestion
				out = fmt::format("{:d}.0", out[0] - '0');
```

src/btop_tools.cpp
```diff
@@ -471,15 +463,17 @@ namespace Tools {
 		}
 
 		if (shorten) {
-			auto f_pos = out.find(".");
-			if (f_pos == 1 and out.size() > 3) {
+			if (out.find(".") != string::npos) {
```
deckstose: You can the cached variable here as well.

I'd do something like 

```cpp
auto contains_decimal_separator = out.find('.') != out.npos;

if (!contains_decimal_separator) {
  /* ... */
}
```

yinghao-w: Yes you're right, the exact position doesn't matter, just the presence.

yinghao-w:
Applied suggestions and slight refactor.

Issue #1394, opened 2025-12-07
Add --default-config flag to generate a default config to stdout


nickjj:
**Is your feature request related to a problem? Please describe.**

I manage my btop config file in my dotfiles and I want to quickly see what changed between versions. This could be new btop config items or different defaults.

**Describe the solution you'd like**

It would be cool to be able to run `btop --default-config` or some comparable command to have btop output a default config file for this version and print it to stdout. btop itself wouln't run here, it just outputs a config.

Now I could diff my git committed config vs this output and see what changed.

**Describe alternatives you've considered**

I run `btop -c /tmp/btop-conf` which launches btop with a default config and then writes the file out to disk. Then I diff this file vs the git committed file and delete the tmp file afterwards. It's a few extra steps.

**Additional context**

Totally unrelated but the default config in the README file is for v1.2.2 which is quite outdated.

deckstose:
Would be great if you can take a look at the PR and provide feedback if something is missing/doesn't work.

nickjj:
Thanks so much for looking into it. I dropped in a comment into the PR at https://github.com/aristocratos/btop/pull/1399#issuecomment-3649465759.

PR #1399, opened 2025-12-13
feat: add cli option to dump default config

deckstose:
Closes: https://github.com/aristocratos/btop/issues/1394

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -838,7 +838,7 @@ namespace Config {
 			} else if (ints.contains(name)) {
 				fmt::format_to(std::back_inserter(buffer), std::locale::classic(), "{:L}", ints[name]);
 			} else if (bools.contains(name)) {
-				fmt::format_to(std::back_inserter(buffer), "{}", bools[name] ? "True" : "False");
+				fmt::format_to(std::back_inserter(buffer), "{}", bools[name] ? "true" : "false");
 			}
 			fmt::format_to(std::back_inserter(buffer), "\n");
 		}
--- a/README.md
+++ b/README.md
@@ -1287,7 +1287,6 @@ If you want to disable building tests, pass `-DBUILD_TESTING=OFF` to the configu
 	sudo snap connect btop-desktop:removable-media
 	```
 
-
 ## Configurability
 
 All options changeable from within UI.
@@ -1295,7 +1294,7 @@ Config and log files stored in `$XDG_CONFIG_HOME/btop` or `$HOME/.config/btop` f
 
 #### btop.conf: (auto generated if not found)
 
-```bash
+```toml
 #? Config file for btop v.1.4.5
 
 #* Name of a btop++/bpytop/bashtop formatted ".theme" file, "Default" and "TTY" for builtin themes.
@@ -1303,14 +1302,14 @@ Config and log files stored in `$XDG_CONFIG_HOME/btop` or `$HOME/.config/btop` f
 color_theme = "Default"
 
 #* If the theme set background should be shown, set to False if you want terminal background transparency.
-theme_background = True
+theme_background = true
 
 #* Sets if 24-bit truecolor should be used, will convert 24-bit colors to 256 color (6x6x6 color cube) if false.
-truecolor = True
+truecolor = true
 
 #* Set to true to force tty mode regardless if a real tty has been detected or not.
 #* Will force 16-color mode and TTY theme, set all graph symbols to "tty" and swap out other non tty friendly symbols.
-force_tty = False
+force_tty = false
 
 #* Define presets for the layout of the boxes. Preset 0 is always all boxes shown with default settings. Max 9 presets.
 #* Format: "box_name:P:G,box_name:P:G" P=(0 or 1) for alternate positions, G=graph symbol to use for box.
@@ -1320,13 +1319,13 @@ presets = "cpu:1:default,proc:0:default cpu:0:default,mem:0:default,net:0:defaul
 
 #* Set to True to enable "h,j,k,l,g,G" keys for directional control in lists.
 #* Conflicting keys for h:"help" and k:"kill" is accessible while holding shift.
-vim_keys = False
+vim_keys = false
 
 #* Rounded corners on boxes, is ignored if TTY mode is ON.
-rounded_corners = True
+rounded_corners = true
 
 #* Use terminal synchronized output sequences to reduce flickering on supported terminals.
-terminal_sync = True
+terminal_sync = true
 
 #* Default symbols to use for graph creation, "braille", "block" or "tty".
 #* "braille" offers the highest resolution but might not be included in all fonts.
@@ -1358,40 +1357,40 @@ update_ms = 2000
 proc_sorting = "cpu lazy"
 
 #* Reverse sorting order, True or False.
-proc_reversed = False
+proc_reversed = false
 
 #* Show processes as a tree.
-proc_tree = False
+proc_tree = false
 
 #* Use the cpu graph colors in the process list.
-proc_colors = True
+proc_colors = true
 
 #* Use a darkening gradient in the process list.
-proc_gradient = True
+proc_gradient = true
 
 #* If process cpu usage should be of the core it's running on or usage of the total available cpu power.
-proc_per_core = False
+proc_per_core = false
 
 #* Show process memory as bytes instead of percent.
-proc_mem_bytes = True
+proc_mem_bytes = true
 
 #* Show cpu graph for each process.
-proc_cpu_graphs = True
+proc_cpu_graphs = true
 
 #* Use /proc/[pid]/smaps for memory information in the process info box (very slow but more accurate)
-proc_info_smaps = False
+proc_info_smaps = false
 
 #* Show proc box on left side of screen instead of right.
-proc_left = False
+proc_left = false
 
 #* (Linux) Filter processes tied to the Linux kernel(similar behavior to htop).
-proc_filter_kernel = False
+proc_filter_kernel = false
 
 #* In tree-view, always accumulate child process resources in the parent process.
-proc_aggregate = False
+proc_aggregate = false
 
 #* Should cpu and memory usage display be preserved for dead processes when paused.
-keep_dead_proc_usage = False
+keep_dead_proc_usage = false
 
 #* Sets the CPU stat shown in upper half of the CPU graph, "total" is always available.
 #* Select from a list of detected attributes from the options menu.
@@ -1402,28 +1401,28 @@ cpu_graph_upper = "Auto"
 cpu_graph_lower = "Auto"
 
 #* Toggles if the lower CPU graph should be inverted.
-cpu_invert_lower = True
+cpu_invert_lower = true
 
 #* Set to True to completely disable the lower CPU graph.
-cpu_single_graph = False
+cpu_single_graph = false
 
 #* Show cpu box at bottom of screen instead of top.
-cpu_bottom = False
+cpu_bottom = false
 
 #* Shows the system uptime in the CPU box.
-show_uptime = True
+show_uptime = true
 
 #* Shows the CPU package current power consumption in watts. Requires running `make setcap` or `make setuid` or running with sudo.
-show_cpu_watts = True
+show_cpu_watts = true
 
 #* Show cpu temperature.
-check_temp = True
+check_temp = true
 
 #* Which sensor to use for cpu temperature, use options menu to select from list of available sensors.
 cpu_sensor = "Auto"
 
 #* Show temperatures for cpu cores also if check_temp is True and sensors has been found.
-show_coretemp = True
+show_coretemp = true
 
 #* Set a custom mapping between core and coretemp, can be needed on certain cpus to get correct temperature for correct core.
 #* Use lm-sensors or similar to see which cores are reporting temperatures on your machine.
@@ -1435,10 +1434,10 @@ cpu_core_map = ""
 temp_scale = "celsius"
 
 #* Use base 10 for bits/bytes sizes, KB = 1000 instead of KiB = 1024.
-base_10_sizes = False
+base_10_sizes = false
 
 #* Show CPU frequency.
-show_cpu_freq = True
+show_cpu_freq = true
 
 #* How to calculate CPU frequency, available values: "first", "range", "lowest", "highest" and "average".
 freq_mode = "first"
@@ -1448,7 +1447,7 @@ freq_mode = "first"
 clock_format = "%X"
 
 #* Update main ui in background when menus are showing, set this to false if the menus is flickering too much for comfort.
-background_update = True
+background_update = true
 
 #* Custom cpu model name, empty string to disable.
 custom_cpu_name = ""
@@ -1458,43 +1457,43 @@ custom_cpu_name = ""
 disks_filter = ""
 
 #* Show graphs instead of meters for memory values.
-mem_graphs = True
+mem_graphs = true
 
 #* Show mem box below net box instead of above.
-mem_below_net = False
+mem_below_net = false
 
 #* Count ZFS ARC in cached and available memory.
-zfs_arc_cached = True
+zfs_arc_cached = true
 
 #* If swap memory should be shown in memory box.
-show_swap = True
+show_swap = true
 
 #* Show swap as a disk, ignores show_swap value above, inserts itself after first disk.
-swap_disk = True
+swap_disk = true
 
 #* If mem box should be split to also show disks info.
-show_disks = True
+show_disks = true
 
 #* Filter out non physical disks. Set this to False to include network disks, RAM disks and similar.
-only_physical = True
+only_physical = true
 
 #* Read disks list from /etc/fstab. This also disables only_physical.
-use_fstab = True
+use_fstab = true
 
 #* Setting this to True will hide all datasets, and only show ZFS pools. (IO stats will be calculated per-pool)
-zfs_hide_datasets = False
+zfs_hide_datasets = false
 
 #* Set to true to show available disk space for privileged users.
-disk_free_priv = False
+disk_free_priv = false
 
 #* Toggles if io activity % (disk busy time) should be shown in regular disk usage view.
-show_io_stat = True
+show_io_stat = true
 
 #* Toggles io mode for disks, showing big graphs for disk read/write speeds.
-io_mode = False
+io_mode = false
 
 #* Set to True to show combined read/write io graphs in io mode.
-io_graph_combined = False
+io_graph_combined = false
 
 #* Set the top speed for the io graphs in MiB/s (100 by default), use format "mountpoint:speed" separate disks with whitespace " ".
 #* Example: "/mnt/media:100 /:20 /boot:1".
@@ -1506,10 +1505,10 @@ net_download = 100
 net_upload = 100
 
 #* Use network graphs auto rescaling mode, ignores any values set above and rescales down to 10 Kibibytes at the lowest.
-net_auto = True
+net_auto = true
 
 #* Sync the auto scaling for download and upload to whichever currently has the highest scale.
-net_sync = True
+net_sync = true
 
 #* Starts with the Network Interface specified here.
 net_iface = ""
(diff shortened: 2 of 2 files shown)


nickjj:
I don't know C++ but I did compile this code locally on Arch within WSL 2 and everything appears to work as intended, thanks! Nice touch on coloring the output too.

## Feedback

- In `--help`, consider moving `--default-config    Print default config to standard output` to be output below `-c` for grouping up related commands, not having a short flag feels ok since it could be a less frequently used command and it reserves letters for future flags
    - If this were changed, changing the flag order in the readme and man page would also go with it?

## Compiling

```sh
$ PREFIX=./hmm GPU_SUPPORT=false make

 ██████╗ ████████╗ ██████╗ ██████╗
 ██╔══██╗╚══██╔══╝██╔═══██╗██╔══██╗   ██╗    ██╗
 ██████╔╝   ██║   ██║   ██║██████╔╝ ██████╗██████╗
 ██╔══██╗   ██║   ██║   ██║██╔═══╝  ╚═██╔═╝╚═██╔═╝
 ██████╔╝   ██║   ╚██████╔╝██║        ╚═╝    ╚═╝
 ╚═════╝    ╚═╝    ╚═════╝ ╚═╝      Makefile v1.6
PLATFORM     ?| Linux
ARCH         ?| x86_64
GPU_SUPPORT  :| true
CXX          ?| g++ (15.2.1)
CC           ?| cc (15.2.1)
THREADS      :| 4
REQFLAGS     !| -std=c++23
WARNFLAGS    :| -Wall -Wextra -pedantic
OPTFLAGS     :| -O2
LDCXXFLAGS   :| -pthread -DFMT_HEADER_ONLY -D_GLIBCXX_ASSERTIONS -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG -D_FILE_OFFSET_BITS=64 -fexceptions -fstack-clash-protection -fcf-protection -fstack-protector -DGPU_SUPPORT
CXXFLAGS     +| $(REQFLAGS) $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)
LDFLAGS      +| $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS)


Command 'lowdown' not found: skipping generating man page btop.1
Building btop++ (v1.4.5) Linux x86_64
Configuring obj/config.h
Compiling src/linux/intel_gpu_top/igt_perf.c
Compiling src/linux/intel_gpu_top/intel_device_info.c
Compiling src/linux/intel_gpu_top/intel_name_lookup_shim.c
Compiling src/linux/intel_gpu_top/intel_gpu_top.c
6%   -> obj/intel_device_info.c.o     (32KiB) (00s)
12%  -> obj/intel_name_lookup_shim.c.o(8.0KiB) (00s)
18%  -> obj/igt_perf.c.o              (8.0KiB) (00s)
24%  -> obj/intel_gpu_top.c.o         (20KiB) (00s)
Compiling src/btop.cpp
Compiling src/btop_cli.cpp
Compiling src/btop_config.cpp
Compiling src/btop_draw.cpp
30%  -> obj/btop_cli.o                (260KiB) (20s)
Compiling src/btop_input.cpp
36%  -> obj/btop_config.o             (656KiB) (33s)
Compiling src/btop_menu.cpp
42%  -> obj/btop.o                    (768KiB) (37s)
Compiling src/btop_shared.cpp
48%  -> obj/btop_input.o              (576KiB) (21s)
Compiling src/btop_theme.cpp
60%  -> obj/btop_menu.o               (712KiB) (21s)
60%  -> obj/btop_draw.o               (1.3MiB) (55s)
Compiling src/btop_tools.cpp
Compiling src/main.cpp
66%  -> obj/btop_theme.o              (424KiB) (14s)
72%  -> obj/main.o                    (8.0KiB) (01s)
Compiling src/linux/btop_collect.cpp
78%  -> obj/btop_shared.o             (576KiB) (19s)
src/btop_tools.cpp: In function ‘size_t Tools::wide_ulen(std::string_view)’:
src/btop_tools.cpp:218:30: warning: ‘template<class _Codecvt, class _Elem, class _Wide_alloc, class _Byte_alloc> class std::__cxx11::wstring_convert’ is deprecated [-Wdeprecated-declarations]
  218 |                         std::wstring_convert<std::codecvt_utf8<wchar_t>> conv;
      |                              ^~~~~~~~~~~~~~~
In file included from /usr/include/c++/15.2.1/locale:47,
                 from /usr/include/c++/15.2.1/bits/fs_path.h:36,
                 from /usr/include/c++/15.2.1/filesystem:54,
                 from src/btop_tools.cpp:22:
/usr/include/c++/15.2.1/bits/locale_conv.h:262:33: note: declared here
  262 |     class _GLIBCXX17_DEPRECATED wstring_convert
      |                                 ^~~~~~~~~~~~~~~
src/btop_tools.cpp: In function ‘std::string Tools::uresize(std::string, size_t, bool)’:
src/btop_tools.cpp:248:38: warning: ‘template<class _Codecvt, class _Elem, class _Wide_alloc, class _Byte_alloc> class std::__cxx11::wstring_convert’ is deprecated [-Wdeprecated-declarations]
  248 |                                 std::wstring_convert<std::codecvt_utf8<wchar_t>> conv;
      |                                      ^~~~~~~~~~~~~~~
/usr/include/c++/15.2.1/bits/locale_conv.h:262:33: note: declared here
  262 |     class _GLIBCXX17_DEPRECATED wstring_convert
      |                                 ^~~~~~~~~~~~~~~
84%  -> obj/btop_tools.o              (588KiB) (16s)
90%  -> obj/linux/btop_collect.o      (1.3MiB) (32s)

Linking and optimizing binary...
100% -> bin/btop                      (2.2MiB) (01s)

Build complete in (01m:31s)
```

```sh
nick@kaizen ~/tmp/btop (push-txruxupkwpmo) $ PREFIX=./hmm GPU_SUPPORT=false make install
Installing binary to: ./hmm/bin/btop
Installing doc to: ./hmm/share/doc/btop
Installing themes to: ./hmm/share/btop/themes
Installing desktop entry to: ./hmm/share/applications/btop.desktop
Installing PNG icon to: ./hmm/share/icons/hicolor/48x48/apps/btop.png
Installing SVG icon to: ./hmm/share/icons/hicolor/scalable/apps/btop.sv
```

Here's confirmation that the new `--default-config` flag exists:

```
$ hmm/bin/btop --help
Usage: btop [OPTIONS]

Options:
  -c, --config <file>     Path to a config file
  -d, --debug             Start in debug mode with additional logs and metrics
  -f, --filter <filter>   Set an initial process filter
      --force-utf         Override automatic UTF locale detection
  -l, --low-color         Disable true color, 256 colors only
  -p, --preset <id>       Start with a preset (0-9)
  -t, --tty               Force tty mode with ANSI graph symbols and 16 colors only
      --themes-dir <dir>  Path to a custom themes directory
      --no-tty            Force disable tty mode
  -u, --update <ms>       Set an initial update rate in milliseconds
      --default-config    Print default config to standard output
  -h, --help              Show this help message and exit
  -V, --version           Show a version message and exit (more with --version)
```

## Outputs

Running it produces this output. The expectation is it produces the default values for this version. I cannot confirm if all of those are the expected defaults but it runs to completion and exits successfully:

### Colors

image

I also redirected the output to a file and the ANSI codes weren't in the output which is good. It came out as plain text, the same happens if you pipe the output of the command to `less` or some other tool.

### Commands

```
$ hmm/bin/btop --default-config | wc -l
269
```

```btop
$ hmm/bin/btop --default-config
#? Config file for btop v.1.4.5

#* Name of a btop++/bpytop/bashtop formatted ".theme" file, "Default" and "TTY" for builtin themes.
#* Themes should be placed in "../share/btop/themes" relative to binary or "$HOME/.config/btop/themes"
color_theme = "Default"

#* If the theme set background should be shown, set to False if you want terminal background transparency.
theme_background = true

#* Sets if 24-bit truecolor should be used, will convert 24-bit colors to 256 color (6x6x6 color cube) if false.
truecolor = true

#* Set to true to force tty mode regardless if a real tty has been detected or not.
#* Will force 16-color mode and TTY theme, set all graph symbols to "tty" and swap out other non tty friendly symbols.
force_tty = false

#* Define presets for the layout of the boxes. Preset 0 is always all boxes shown with default settings. Max 9 presets.
#* Format: "box_name:P:G,box_name:P:G" P=(0 or 1) for alternate positions, G=graph symbol to use for box.
#* Use whitespace " " as separator between different presets.
#* Example: "cpu:0:default,mem:0:tty,proc:1:default cpu:0:braille,proc:0:tty"
presets = "cpu:1:default,proc:0:default cpu:0:default,mem:0:default,net:0:default cpu:0:block,net:0:tty"

#* Set to True to enable "h,j,k,l,g,G" keys for directional control in lists.
#* Conflicting keys for h:"help" and k:"kill" is accessible while holding shift.
vim_keys = false

#* Rounded corners on boxes, is ignored if TTY mode is ON.
rounded_corners = true

#* Use terminal synchronized output sequences to reduce flickering on supported terminals.
terminal_sync = true

#* Default symbols to use for graph creation, "braille", "block" or "tty".
#* "braille" offers the highest resolution but might not be included in all fonts.
#* "block" has half the resolution of braille but uses more common characters.
#* "tty" uses only 3 different symbols but will work with most fonts and should work in a real TTY.
#* Note that "tty" only has half the horizontal resolution of the other two, so will show a shorter historical view.
graph_symbol = "braille"

# Graph symbol to use for graphs in cpu box, "default", "braille", "block" or "tty".
graph_symbol_cpu = "default"

# Graph symbol to use for graphs in gpu box, "default", "braille", "block" or "tty".
graph_symbol_gpu = "default"

# Graph symbol to use for graphs in cpu box, "default", "braille", "block" or "tty".
graph_symbol_mem = "default"

# Graph symbol to use for graphs in cpu box, "default", "braille", "block" or "tty".
graph_symbol_net = "default"

# Graph symbol to use for graphs in cpu box, "default", "braille", "block" or "tty".
graph_symbol_proc = "default"

#* Manually set which boxes to show. Available values are "cpu mem net proc" and "gpu0" through "gpu5", separate values with whitespace.
shown_boxes = "cpu mem net proc"

#* Update time in milliseconds, recommended 2000 ms or above for better sample times for graphs.
update_ms = 2000

#* Processes sorting, "pid" "program" "arguments" "threads" "user" "memory" "cpu lazy" "cpu direct",
#* "cpu lazy" sorts top process over time (easier to follow), "cpu direct" updates top process directly.
proc_sorting = "cpu lazy"

#* Reverse sorting order, True or False.
proc_reversed = false

#* Show processes as a tree.
proc_tree = false

#* Use the cpu graph colors in the process list.
proc_colors = true

#* Use a darkening gradient in the process list.
proc_gradient = true

#* If process cpu usage should be of the core it's running on or usage of the total available cpu power.
proc_per_core = false

#* Show process memory as bytes instead of percent.
proc_mem_bytes = true

#* Show cpu graph for each process.
proc_cpu_graphs = true

#* Use /proc/[pid]/smaps for memory information in the process info box (very slow but more accurate)
proc_info_smaps = false

#* Show proc box on left side of screen instead of right.
proc_left = false

#* (Linux) Filter processes tied to the Linux kernel(similar behavior to htop).
proc_filter_kernel = false

#* In tree-view, always accumulate child process resources in the parent process.
proc_aggregate = false

#* Should cpu and memory usage display be preserved for dead processes when paused.
keep_dead_proc_usage = false

#* Sets the CPU stat shown in upper half of the CPU graph, "total" is always available.
#* Select from a list of detected attributes from the options menu.
cpu_graph_upper = "Auto"

#* Sets the CPU stat shown in lower half of the CPU graph, "total" is always available.
#* Select from a list of detected attributes from the options menu.
cpu_graph_lower = "Auto"

#* If gpu info should be shown in the cpu box. Available values = "Auto", "On" and "Off".
show_gpu_info = "Auto"

#* Toggles if the lower CPU graph should be inverted.
cpu_invert_lower = true

#* Set to True to completely disable the lower CPU graph.
cpu_single_graph = false

#* Show cpu box at bottom of screen instead of top.
cpu_bottom = false

#* Shows the system uptime in the CPU box.
show_uptime = true

#* Shows the CPU package current power consumption in watts. Requires running `make setcap` or `make setuid` or running with sudo.
show_cpu_watts = true

#* Show cpu temperature.
check_temp = true

#* Which sensor to use for cpu temperature, use options menu to select from list of available sensors.
cpu_sensor = "Auto"

#* Show temperatures for cpu cores also if check_temp is True and sensors has been found.
show_coretemp = true

#* Set a custom mapping between core and coretemp, can be needed on certain cpus to get correct temperature for correct core.
#* Use lm-sensors or similar to see which cores are reporting temperatures on your machine.
#* Format "x:y" x=core with wrong temp, y=core with correct temp, use space as separator between multiple entries.
#* Example: "4:0 5:1 6:3"
cpu_core_map = ""

#* Which temperature scale to use, available values: "celsius", "fahrenheit", "kelvin" and "rankine".
temp_scale = "celsius"

#* Use base 10 for bits/bytes sizes, KB = 1000 instead of KiB = 1024.
base_10_sizes = false

#* Show CPU frequency.
show_cpu_freq = true

#* How to calculate CPU frequency, available values: "first", "range", "lowest", "highest" and "average".
freq_mode = "first"

#* Draw a clock at top of screen, formatting according to strftime, empty string to disable.
#* Special formatting: /host = hostname | /user = username | /uptime = system uptime
clock_format = "%X"

#* Update main ui in background when menus are showing, set this to false if the menus is flickering too much for comfort.
background_update = true

#* Custom cpu model name, empty string to disable.
custom_cpu_name = ""

#* Optional filter for shown disks, should be full path of a mountpoint, separate multiple values with whitespace " ".
#* Only disks matching the filter will be shown. Prepend exclude= to only show disks not matching the filter. Examples: disk_filter="/boot /home/user", disks_filter="exclude=/boot /home/user"
disks_filter = ""

#* Show graphs instead of meters for memory values.
mem_graphs = true

#* Show mem box below net box instead of above.
mem_below_net = false

#* Count ZFS ARC in cached and available memory.
zfs_arc_cached = true

#* If swap memory should be shown in memory box.
show_swap = true

#* Show swap as a disk, ignores show_swap value above, inserts itself after first disk.
swap_disk = true

#* If mem box should be split to also show disks info.
show_disks = true

#* Filter out non physical disks. Set this to False to include network disks, RAM disks and similar.
only_physical = true

#* Read disks list from /etc/fstab. This also disables only_physical.
use_fstab = true

#* Setting this to True will hide all datasets, and only show ZFS pools. (IO stats will be calculated per-pool)
zfs_hide_datasets = false

#* Set to true to show available disk space for privileged users.
disk_free_priv = false

#* Toggles if io activity % (disk busy time) should be shown in regular disk usage view.
show_io_stat = true

#* Toggles io mode for disks, showing big graphs for disk read/write speeds.
io_mode = false

#* Set to True to show combined read/write io graphs in io mode.
io_graph_combined = false

#* Set the top speed for the io graphs in MiB/s (100 by default), use format "mountpoint:speed" separate disks with whitespace " ".
#* Example: "/mnt/media:100 /:20 /boot:1".
io_graph_speeds = ""

#* Set fixed values for network graphs in Mebibits. Is only used if net_auto is also set to False.
net_download = 100

net_upload = 100

#* Use network graphs auto rescaling mode, ignores any values set above and rescales down to 10 Kibibytes at the lowest.
net_auto = true

#* Sync the auto scaling for download and upload to whichever currently has the highest scale.
net_sync = true

#* Starts with the Network Interface specified here.
net_iface = ""

#* "True" shows bitrates in base 10 (Kbps, Mbps). "False" shows bitrates in binary sizes (Kibps, Mibps, etc.). "Auto" uses base_10_sizes.
base_10_bitrate = "Auto"

#* Show battery stats in top right if battery is present.
show_battery = true

#* Which battery to use if multiple are present. "Auto" for auto detection.
selected_battery = "Auto"

#* Show power stats of battery next to charge indicator.
show_battery_watts = true

#* Set loglevel for "~/.local/state/btop.log" levels are: "ERROR" "WARNING" "INFO" "DEBUG".
#* The level set includes all lower levels, i.e. "DEBUG" will show all logging info.
log_level = "WARNING"

#* Measure PCIe throughput on NVIDIA cards, may impact performance on certain cards.
nvml_measure_pcie_speeds = true

#* Measure PCIe throughput on AMD cards, may impact performance on certain cards.
rsmi_measure_pcie_speeds = true

#* Horizontally mirror the GPU graph.
gpu_mirror_graph = true

#* Set which GPU vendors to show. Available values are "nvidia amd intel"
shown_gpus = "nvidia amd intel"

#* Custom gpu0 model name, empty string to disable.
custom_gpu_name0 = ""

#* Custom gpu1 model name, empty string to disable.
custom_gpu_name1 = ""

#* Custom gpu2 model name, empty string to disable.
custom_gpu_name2 = ""

#* Custom gpu3 model name, empty string to disable.
custom_gpu_name3 = ""

#* Custom gpu4 model name, empty string to disable.
custom_gpu_name4 = ""

#* Custom gpu5 model name, empty string to disable.
custom_gpu_name5 = ""
```

## Diff

I then diff'd it against my existing v1.4.4 config with `diff ~/.config/btop/btop.conf <(~/tmp/btop/hmm/bin/btop --default-config)` and saw a lot of `True` vs `true` and related changes since it looks like booleans are lowercase now so this is likely expected? I also saw new additions such as but not limited to:

```diff
+#* How to calculate CPU frequency, available values: "first", "range", "lowest", "highest" and "average".
+freq_mode = "first"
```

## stdout test

This confirms the output produced by the new flag gets written to stdout and not stderr:

```sh
# This returns nothing as expected since we redirect stdout to /dev/null
$ hmm/bin/btop --default-config 1>/dev/null

# This returns everything as expected since we redirect stderr to /dev/null
$ hmm/bin/btop --default-config 2>/dev/null
[redacted all of the config items listed above]
```

## btop config

- Running `hmm/bin/btop` runs successfully but seemed to have picked up my `~/config/btop/btop.conf` file, is that intended?
- Running `hmm/bin/btop --default-config > btop.conf` produces a new default config
- Running `hmm/bin/btop -c btop.conf` runs successfully and uses the passed in config
- Running `echo hello >> btop.conf` appends a dummy line to the end of the written config
    - Running `hmm/bin/btop --default-config -c btop.conf` runs successfully and prints the default config
    - Running `hmm/bin/btop -c btop.conf --default-config` runs successfully and prints the default config
    - For the above 2 tests, I set `truecolor = false` in `btop.conf` but the `true` version was printed
- Running `hmm/bin/btop --default-config --help` prints the default config which is ok but would it make sense for `--help` to always win if there's competing flags? I notice if you flip the order then the help gets produced which makes me think the flags are read in order, that's why I didn't add this as feedback since it's ok either way

src/btop_config.cpp
```diff
@@ -828,4 +819,29 @@ namespace Config {
...
+			}
+
+			fmt::format_to(std::back_inserter(buffer), "{} = ", name);
+			// Lookup default value by name and write it out.
+			if (strings.contains(name)) {
+				fmt::format_to(std::back_inserter(buffer), R"("{}")", strings[name]);
+			} else if (ints.contains(name)) {
+				fmt::format_to(std::back_inserter(buffer), std::locale::classic(), "{:L}", ints[name]);
+			} else if (bools.contains(name)) {
+				fmt::format_to(std::back_inserter(buffer), "{}", bools[name] ? "true" : "false");
```
aristocratos, line 841: Why is this different from how bools are represented in string format in the rest of the program (with capitalized first letter)?

https://github.com/aristocratos/btop/blob/c2b477ec9fa701d892326cccc87b870aee1d6e6b/src/btop_config.cpp#L614-L622

https://github.com/aristocratos/btop/blob/c2b477ec9fa701d892326cccc87b870aee1d6e6b/src/btop_config.cpp#L795-L796

deckstose: As explained in the commit message, this makes our config `toml`. This is great for interop with other tools and libraries.

aristocratos: We are not parsing it as `toml`, so having it appear to be toml compatible would be bad since we would fail to read any string not formatted as a "basic" single line string.

I see no issue with changing format to `toml`, but in that case proper parsing for `toml` would need to be added (literals vs basic string, multiline, escaping, etc...).

deckstose: I get what you mean, but since we don't advertise it (no toml extension, no mention in the readme), I don't see a problem. It just opens up the possibility to use a [toml parser](https://github.com/marzer/tomlplusplus), autocompletion with a schema, etc, in the future.

nickjj: Does using toml in the future also help with the idea of maybe allowing config merging / overrides? For example being able to have all defaults handled by btop without a config file but if you define 3 options in your own user config, btop will merge those 3 options in, override their defaults and continue using the other btop-defined defaults?

PR #1401, opened 2025-12-14
Feat: add mouse dragging of scrollbar

yinghao-w:
Adds the ability to drag the proc box scrollbar with the mouse.
Question: should `Proc::scroll_pos` be an `atomic<int>`? I see that `scroll_pos` is updated on the drawing thread, then read by `Input::process` on the main thread, but the other `Proc::` variables used in `btop_input.cpp` are not atomic.

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -355,8 +355,9 @@ namespace Input {
 					const auto& [col, line] = mouse_pos;
 					const int y = (Config::getB("show_detailed") ? Proc::y + 8 : Proc::y);
 					const int height = (Config::getB("show_detailed") ? Proc::height - 8 : Proc::height);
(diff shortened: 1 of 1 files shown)


src/btop_input.cpp
```diff
@@ -375,11 +382,22 @@ namespace Input {
...
-						else
+						else if (key.starts_with("mouse_scroll_")) {
 							goto proc_mouse_scroll;
+						}
+						else if (key == "mouse_drag" and dragging_scroll) {
+							goto proc_scroll_drag;
+						}
+					}
+					else if (key == "mouse_drag" and dragging_scroll) {
+						proc_scroll_drag:
```
deckstose: Any way you can manage this neatly without introducing a new `goto`?

yinghao-w: Sure I think so, if I refactor the parent if-chain.

PR #1402, opened 2025-12-17
feat: add config option to disable mouse events

deckstose:


deckstose:
We do not need to keep track of current mouse state, right? There is no issue with sending `mouse_on` to a terminal that already has mouse support enabled?

aristocratos (approved):

aristocratos:
> We do not need to keep track of current mouse state, right? There is no issue with sending `mouse_on` to a terminal that already has mouse support enabled?

No issue, it's ignored if already in the state you're trying to set.

PR #1409, opened 2025-12-22
fix: remove unused asprintf implementation function

deckstose:


deckstose:
It's implemented in both musl and glibc. I don't see why this would be required. I've tested it on both glibc and musl, and on arm.

deckstose:
Otherwise we shouldn't re-implement that anyways but define `_DEFAULT_SOURCE`

aristocratos (commented):
Have you tested this on all supported platforms and compilers?

I would assume this was added because it was needed at some point, so you need to make sure it isn't anymore if you want to remove it.

deckstose:
What other platforms did you have in mind?

aristocratos:
The supported ones...
Linux, Macos, XXX_BSD

deckstose:
This is Linux only

aristocratos (approved):

PR #1425, opened 2025-12-28
Improve Makefile readability by symbolizing ANSI

dotysan:
All messy inline `\033[` jazz is removed.

In refactoring the Makefile, I did stumble across a couple bugs. However, I tried not to change any logic in this branch except for some inconsistent formatting. For example, when outputting `Message...` the trailing ... were colorized inconsistently. So I switched all trailing ... to non-formatted for consistency. Could also leave it bold, if maintainers prefer that?

Also, the `$(P)` symbol is no longer needed.

And at least one unreferenced `$$TSTAMP` was removed.

And the new `file_with_size()` hides du stderr. Is that right?

And the new `step_duration()` replaces a reference to `date` with `$(DATE_CMD)`. But I didn't do this globally. There are a few more, such as in `@TSTAMP=`.

I've spot tested on linux only.


ansi_banner.utf8
```diff
@@ -0,0 +1 @@
...96m██████\033[38;5;240m╗ \033[38;5;196m████████\033[38;5;240m╗ \033[38;5;196m██████\033[38;5;240m╗ \033[38;5;196m██████\033[38;5;240m╗\n \033[38;5;160m██\033[38;5;239m╔══\033[38;5;160m██\033[38;5;239m╗╚══\033[38;5;160m██\033[38;5;239m╔══╝\033[38;5;160m██\033[38;5;239m╔═══\033[38;5;160m██\033[38;5;239m╗\033[38;5;160m██\033[38;5;239m╔══\033[38;5;160m██\033[38;5;239m╗   \033[38;5;160m██\033[38;5;239m╗    \033[38;5;160m██\033[38;5;239m╗\n \033[38;5;124m██████\033[38;5;238m╔╝   \033[38;5;124m██\033[38;5;238m║   \033[38;5;124m██\033[38;5;238m║   \033[38;5;124m██\033[38;5;238m║\033[38;5;124m██████\033[38;5;238m╔╝ \033[38;5;124m██████\033[38;5;238m╗\033[38;5;124m██████\033[38;5;238m╗\n \033[38;5;88m██\033[38;5;237m╔══\033[38;5;88m██\033[38;5;237m╗   \033[38;5;88m██\033[38;5;237m║   \033[38;5;88m██\033[38;5;237m║   \033[38;5;88m██\033[38;5;237m║\033[38;5;88m██\033[38;5;237m╔═══╝  ╚═\033[38;5;88m██\033[38;5;237m╔═╝╚═\033[38;5;88m██\033[38;5;237m╔═╝\n \033[38;5;52m██████\033[38;5;236m╔╝   \033[38;5;52m██\033[38;5;236m║   ╚\033[38;5;52m██████\033[38;5;236m╔╝\033[38;5;52m██\033[38;5;236m║        ╚═╝    ╚═╝\n \033[38;5;235m╚═════╝    ╚═╝    ╚═════╝ ╚═╝      \033[1;3;38;5;240mMakefile v1.6\033[0m
```
deckstose: This just pollutes the repository

deckstose (commented):
I don't see much value in this.

ansi_banner.utf8
aristocratos: How so?

aristocratos: The last part "Makefile v1.6" should probably be added to the `BANNER_FILE` variable instead so it's not missed if updating the Makefile.

aristocratos (commented):
Good effort to simplify, but needs to be tested on all platforms before merging.

ansi_banner.utf8
deckstose: I could have said that in a nicer way:

_If_ we split the banner out in a file, this should go in a dedicated folder, conventionally this is `assets` (the pictures in the README from `Img` should go in that folder, too).

Saying that I don't see why adding a dedicated file for a one line variable is necessary (it's not used anywhere else, that's the only reason I would deem this useful).

The escape codes into variables looks nice tho

dotysan: > _If_ we split the banner out in a file, this should go in a dedicated folder, conventionally this is `assets` (the pictures in the README from `Img` should go in that folder, too).

Agreed! I originally put it in assets/ but then "simplified" for this PR.

Your idea of moving all assets (including `Img/`) makes perfect sense.

The bigger solution is to have that ANSI banner sequence appear once as a single source of truth. But it's already embedded in the cpp... so I didn't want to override that.

aristocratos (approved):

PR #1434, opened 2026-01-04
Change color of 'tab' in options menu for parity

TheSovietPancakes:
This PR changes the color of the word "tab" in the top left corner of the options menu. This is to keep parity with other keyboard shortcut "indicators", which are highlighted in blue to let the user know that pressing that key will do an action.
Also note that the arrow intersecting the box symbols is a quirk of my font, and this change did not cause that. :)

Before change:
image

After change:
image


deckstose (dismissed):
Thanks 💯  I'll ask that on every occasion: do you mind also converting the string allocation you modify to switch to `fmt::format` (the last parameter to `Draw::createBox`)?

TheSovietPancakes:
Yep, just changed 👍

deckstose:
One step at a time 😎

PR #1441, opened 2026-01-06
Add Russian localization to the XDG desktop entry

adem4ik:


wehrwolfmann:
Not my PR — just a user who would like it in, adding a data point in case it simply went quiet.

It still applies cleanly to main today, and `desktop-file-validate` is silent on the result; I checked locally against current master. GitHub also still reports it as mergeable after eight months, so nothing has drifted.

For context: the pt_BR entry (#1471) was opened a week after this one and merged the same day, so there does not seem to be anything wrong with the change itself.

Right now the desktop entry is the one place btop shows English to Russian-speaking users, and this is four lines of metadata with no effect on the build. Would be good to see it land.

Issue #1348, opened 2025-11-05
RFC: Remove redundant `btop_*` prefix from source files


deckstose:
1. It's redundant
2. For TUI editor users with sufficient auto comp, it makes it a little bit more annoying to open a project file
    1. Press `s<tab>` -> `src/`
    2. _Press `b<tab>` -> `src/btop_`_
    3. Press `<char><tab>` -> `src/btop_draw.`
    4. Choose header or source file (`h<tab>` or `c<tab`)
 
The second step can be avoided. Seems a little nit-picky, however I do that quite a lot.

vandabbin:
For what it is worth, I would agree with this. The program name repeated in every src file feels unnecessary

PR #1449, opened 2026-01-08
refactor: remove superfluous btop prefix from source files

deckstose:
Closes: https://github.com/aristocratos/btop/issues/1348


aristocratos:
That's fine on my part, but I guess it's gonna get a bit messy to get in (lots of broken PR's)?

deckstose:
I think we are overestimating this a bit, we don't have lots of large, active, open PR's (that aren't already broken), but yes, they will break. However it's easy to rename the files and fix the couple of includes.

But that's why I'd like to bundle that with the formatting PR.

Issue #1451, opened 2026-01-09
Some mouse clicks in proc box dont disengage following mode


vandabbin:
When I made a change to the process following mode to make changing the selection disengage following mode, I missed a couple points in the mouse input code that were preventing following mode from exiting.

If you clicked on the detailed view box, the process list top border,  or the label line following mode wouldn't disengage. If the detailed box was open then it would not be selected and following mode would stay engaged.

PR #1452, opened 2026-01-09
Minor improvements to process following

vandabbin:
Fixes: #1451

When making the change to the process following PR that made moving the selection disengage following mode, I missed a couple points in the mouse input code that was preventing following mode from being disengaged when clicking on the detailed info box or on the upper border of the proc box/label line.

This alteration fixes that.

edit: 
Also made a change so that if the detailed view is open/opened and
following mode engages. If the detailed view process is the one being
followed then the selection is set to 0 and the detailed view buttons
are active instead of the bottom bar buttons.

added Follow button to detailed view

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -361,7 +361,8 @@ namespace Config {
 		{"selected_depth", 0},
 		{"proc_start", 0},
 		{"proc_selected", 0},
-		{"proc_last_selected", 0}
+		{"proc_last_selected", 0},
+		{"proc_followed", 0},
 	};
 	std::unordered_map<std::string_view, int> intsTmp;
 
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1564,16 +1564,22 @@ namespace Proc {
 	string box;
 
 	int selection(const std::string_view cmd_key) {
-		if ((Config::getB("follow_process") and not Config::getB("pause_proc_list"))) {
-			Config::flip("follow_process");
-			Config::set("followed_pid", 0);
-			redraw = true;
-		}
 		auto start = Config::getI("proc_start");
 		auto selected = Config::getI("proc_selected");
 		auto last_selected = Config::getI("proc_last_selected");
 		const int select_max = (Config::getB("show_detailed") ? (Config::getB("proc_banner_shown") ? Proc::select_max - 9 : Proc::select_max - 8) :
 																(Config::getB("proc_banner_shown") ? Proc::select_max - 1 : Proc::select_max));
+
+		if (Config::getB("follow_process")) {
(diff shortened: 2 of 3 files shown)


aristocratos:
I also noticed that the buttons in the details box is greyed out when following.
So there should also probably be a conditional there so the buttons are active if in following mode and the "details" box is showing the process that is being followed.

vandabbin:
@aristocratos 

> I also noticed that the buttons in the details box is greyed out when following. So there should also probably be a conditional there so the buttons are active if in following mode and the "details" box is showing the process that is being followed.

Ok yeah I can fix that as well. I think I had made a conscious choice to leave it like that after learning that the mouse_mapping for the buttons on the top bar and bottom bar were shared and to have both working at the same time required a bit of extra code. If you think that would be better I will do that. Or were you thinking that the bottom bar should be greyed out instead when following is active and the detailed view is open for that process?

aristocratos:
I think it's just an issue about the "selection" still being active when entering following mode (or pressing enter when the option for auto follow is enabled).

I think the best approach would be that when entering following mode (with `F` or `enter`) the selection is removed. And then if pressing `up` or `down` will resume selection from the position of the followed process. If pressing `enter` again while following and detailed view is visible, follow mode is disabled, detailed view is closed and selection goes back to the previously followed process.

vandabbin:
@aristocratos 

> I think it's just an issue about the "selection" still being active when entering following mode (or pressing enter when the option for auto follow is enabled).
> 
> I think the best approach would be that when entering following mode (with `F` or `enter`) the selection is removed. And then if pressing `up` or `down` will resume selection from the position of the followed process. If pressing `enter` again while following and detailed view is visible, follow mode is disabled, detailed view is closed and selection goes back to the previously followed process.

Ah okay I see what you are saying. I'll go through and make it work that way instead

vandabbin:
@aristocratos 

Okay so I have made these improvements. 

when the detailed view is open and is the same process as the followed process the detailed view box is selected & active.
moving away from the followed process returns selection to followed process location and then moves from there.

Oh there is one more thing I found to fix now that this has been changed the Follow button on bottom bar isnt active for detailed view following or maybe I should just add it as a button in the detailed view as well?

vandabbin:
I decided to add the Follow button to the detailed view buttons as well

vandabbin:
> I think the best approach would be that when entering following mode (with `F` or `enter`) the selection is removed. 

I found it easier for the selection to only be removed if the detailed view is open and is the same as the followed process. So if the user starts following a process by pressing 'F' and it isn't the detailed view process then it continues to make the selection the same as the followed. This made it easier to keep the bottom bar buttons active when following a process with the detailed view closed as there are many checks for (selected > 0).

Maybe this is what you meant all along though 😆

PR #1408, opened 2025-12-22
feat(linux): add Intel Xe GPU driver support

deveworld:
## Summary

Add support for Intel Xe GPU driver (Lunar Lake, etc). #1407 

## Changes

- Add dynamic PMU device discovery (`find_xe_pmu_device`, `find_i915_pmu_device`)
  - Supports both `xe` and `xe_XXXX_XX_XX.X` PMU device names
- Add Xe PMU event constants (`XE_PMU_ENGINE_ACTIVE_TICKS`, etc.)
- Add Xe-specific engine discovery in `discover_engines()`
- Add `pmu_init_xe()` for Xe PMU initialization
- Change hardcoded `"i915"` device to dynamic detection
- Add Xe-specific utilization calculation (`active_ticks / total_ticks`)
- Add power availability check for GPUs without RAPL `energy-gpu`

## Tests

- [x] Tested on Intel Lunar Lake (Core Ultra 5 228V, device ID 0x64A0)
- [x] Build succeeds
- [x] GPU utilization and frequency display correctly
- [ ] i915 systems (no hardware available, but code path unchanged)

PR #1457, opened 2026-01-10
Add Intel Xe GPU driver support

### LINKED ISSUE #1407: [Feature Request] Add Intel Xe driver support (Lunar Lake, etc.)

btop cannot detect Intel GPUs using the Xe driver (Lunar Lake, Battlemage, and future Intel GPUs).
Even with root privileges, the GPU is not visible in btop.

I'm using a Samsung Galaxy Book 5 Pro with Intel Core Ultra 5 228V (Lunar Lake, device ID 0x64A0).
The system uses the `xe` driver (not i915), which is confirmed working:

```
$ lspci -nn | grep VGA
00:02.0 VGA compatible controller [0300]: Intel Corporation Lunar Lake [Intel Arc Graphics 130V / 140V] [8086:64a0] (rev 04)

$ ls /sys/bus/event_source/devices/ | grep xe
xe_0000_00_02.0
```

-  Dynamically discover PMU device names (`xe`, `xe_*`, or `i915_*`) instead of hardcoding `i915`
- Add Xe PMU event handling (`engine-active-ticks`, `engine-total-ticks`, `gt-actual-frequency`, `gt-c6-residency`)
- Calculate Xe utilization as `active_ticks / total_ticks * 100`
- Keep i915 behavior unchanged

- Use `intel_gpu_top` standalone
- Wait for an i915 compatibility layer (no announced plans)

**Additional context**

There are Xe helper functions in `src/linux/intel_gpu_top/igt_perf.c` (`xe_perf_device()`, `xe_perf_type_id()`, `perf_xe_open()`) and Lunar Lake PCI IDs in `src/linux/intel_gpu_top/xe_pciids.h`, but integration is missing:

- `src/linux/btop_collect.cpp` hardcodes `const char* device = "i915";`
- `src/linux/intel_gpu_top/intel_gpu_top.c` only discovers `*-busy` events (i915-specific) and uses `I915_PMU_*` constants
- No `XE_PMU_*` constants are defined

Likely affected devices: Intel Lunar Lake (Core Ultra 200V), Intel Battlemage (Arc B-series), and future Xe-driver GPUs.

deveworld:
## Summary

Adds support for Intel GPUs using the **Xe kernel driver**, required for newer Intel hardware such as **Lunar Lake**, **Battlemage**, and newer Arc / Arc Pro devices, which the legacy i915 path cannot detect.

Closes #1407

## What Changed

- Dedicated `Xe` collection path in `src/linux/btop_collect.cpp` with multi-GPU discovery and per-instance i915/xe routing
- Minimal `src/linux/xe_drm.h` with the Xe UAPI definitions btop needs
- Utilization from fdinfo (`drm-cycles-*` / `drm-total-cycles-*`), falling back to gtidle sysfs, then Xe PMU perf events
- GPU clock from Xe sysfs frequency nodes
- Dedicated VRAM via `DRM_IOCTL_XE_DEVICE_QUERY` (skips SYSMEM-only devices so shared system memory is not shown as VRAM)
- GPU power for discrete Arc via hwmon (`power1_average` / `power1_input`)
- RC/MC split graphs, without modifying `src/linux/intel_gpu_top/`
- Works without root: when the DRM render node can't be opened (user not in the `render` group), VRAM is skipped but gtidle/fdinfo utilization and clock still work, so the GPU is shown instead of being dropped

## Implementation Notes

Per maintainer feedback on #1408:
- **Does NOT modify** any files in `src/linux/intel_gpu_top/`
- Deduplicates fdinfo by `drm-client-id`; uses the per-class max of `drm-total-cycles-*` as the shared GPU timestamp
- The `/proc` fdinfo scan is prefiltered via `/proc/*/fd` symlinks and iterates with `error_code` (no per-inaccessible-process exceptions)

## Supported Metrics (Xe driver)

| Metric | Support | Source |
| ------- | ------- | ------ |
| GPU Utilization | ✅ | fdinfo, with gtidle/PMU fallback |
| GPU Clock Speed | ✅ | Xe sysfs frequency |
| Memory Usage | ✅ (dedicated VRAM only) | `DRM_XE_DEVICE_QUERY_MEM_REGIONS` |
| Power Usage | ✅ discrete (hwmon) / ❌ integrated (no sensor) | `power1_average` / `power1_input` |
| Temperature | ❌ | Not implemented |

## Performance

fdinfo collection is capped to once per 500 ms and runs on the collector thread. Measured on Lunar Lake (Arc 140V, ~540 procs / ~4150 fds, unprivileged): **~24–48 ms per scan (≈32 ms avg, ~16–25 ms warm)**. Per-scan duration is logged at `--debug` (`Xe: fdinfo scan <ms>`).

## Known Limitations

- fdinfo is DRM-client based, so unprivileged runs can undercount processes owned by other users
- Integrated GPUs expose no GPU power sensor (hwmon/PMU), so power is reported only on discrete cards
- Multi-engine class capacity normalization (e.g. Battlemage VCS=2 / DG2 CCS=4 partial use) is deferred to a follow-up

## Validation

- ✅ `make GPU_SUPPORT=true`
- ✅ `cmake -S . -B build-test -DBTOP_GPU=ON && cmake --build build-test && ctest --test-dir build-test`
- ✅ Tested on Intel Lunar Lake (Arc 140V, xe): detection, fdinfo utilization, clock, name lookup, and graceful no-render-group fallback (no crash)
- Community testing in #1407 covered Lunar Lake and Battlemage as the implementation evolved PMU → gtidle → fdinfo


TheSovietPancakes:
[Please do not reformat code if you can](https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md)

deveworld:
> [Please do not reformat code if you can](https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md)

Thanks for the feedback!
I've restored fmt::format usage and the clang diagnostic pragma. 
It should now match the original code style

TheSovietPancakes:
I was referring to the code style. Please continue using fmt, since this is preferred. But I was referring to things like changing the indentation of completely unrelated code. This makes for a larger diff and can sometimes get your PR rejected outright in some repositories.
(@deckstose Ngl this would be a really great usecase for .clang-format, it's starting to warm up on me 😆)
If you can make one big change to fix Intel, without touching other code, that would be awesome 😄

MrChuw:
I tested on my B580 sparkle, it works as expected, but I have a suggestion, how about using ```nvtop``` if available, the ```nvtop -i (or -i)``` option provides an output in json format with memory and power usage and also temperature and more.



Here a example of nvtop output.
```json
[
  {
   "device_name": "Battlemage G21 (Arc B580)",
   "gpu_clock": "1200MHz",
   "mem_clock": null,
   "temp": "54C",
   "fan_speed": "0RPM",
   "power_draw": "43W",
   "gpu_util": "11%",
   "encode_decode": "0%",
   "mem_util": "54%",
   "mem_total": "12809404416",
   "mem_used": "6956257280",
   "mem_free": "5853147136",
   "processes" : [
     {
       "pid": "1",
       "cmdline": "/usr/lib/systemd/systemd --switched-root --system --deserialize=56",
       "kind": "graphic",
       "user": "root",
       "gpu_usage": "0%",
       "gpu_mem_bytes_alloc": null,
       "gpu_mem_usage": null,
       "encode_decode": "0%"
     },
     {
       "pid": "894",
       "cmdline": "/usr/lib/systemd/systemd-logind",
       "kind": "graphic",
       "user": "root",
       "gpu_usage": "0%",
       "gpu_mem_bytes_alloc": null,
       "gpu_mem_usage": null,
       "encode_decode": "0%"
     }
   ]
  }
]

```

PrakarshPanwar:
Works well on my both laptops, Tiger Lake-H and Raptor Lake-P iGPUs. No need for sudo access to read GPU stats. When it will be merged?

deveworld:
Rebased this on current `upstream/main`, cleaned up the Xe path, and updated the PR description so it matches the code again.

The branch now prefers `fdinfo` for utilization, keeps `gtidle`/PMU as fallback, handles `drm-engine-capacity-*` when available, and only reports actual dedicated VRAM. I also fixed the remaining local Xe build warnings.

I reran `make -j4` and `cmake -S . -B build-test && cmake --build build-test -j4 && ctest --test-dir build-test --output-on-failure`.

The main limitation still left is that unprivileged `fdinfo` collection can undercount protected clients.

MrChuw:
latest commit, making the percentage wrong.

deveworld:
Thanks, @MrChuw 

The last Xe fdinfo change was skewing the percentage. I had added `drm-engine-capacity-*` into the calculation, and that made the reported value too low for cases like the one in your screenshot.

That part is removed, so the latest head should be back to the expected raw fdinfo-style value.

If you get a chance, please try the latest commit again and let me know how it looks.

MrChuw:
The last commit where the GPU utilization on my B580 was right was 45aa98fe08bec82db096b797d98674111b955394 

the latest commit still with skewed utilization 922a37e43b098bde231a03e8379628d0b186f885

src/linux/btop_collect.cpp
```diff
@@ -1807,124 +1876,1176 @@ namespace Gpu {
...
+				uint64_t main_total_cycles = 0;
+				uint64_t media_total_cycles = 0;
+				bool has_main = false;
+				bool has_media = false;
+			};
+
+			// Collect GPU cycles from all processes' fdinfo with client-id deduplication.
+			// Keep raw per-class totals so the resulting utilization matches the semantics
+			// used by existing DRM tools such as nvtop.
+			static FdinfoCycles collect_fdinfo_cycles(const string& pci_slot) {
```
aristocratos: Whats the performance impact of this? (How long does the collection cycle for GPU take when running btop with `--debug` when using the `fdinfo` method?)

deveworld:
Rebased and pushed.

@aristocratos, `collect_fdinfo_cycles`: 110–150 ms -> ~90 ms/sample after `229e6f6` (`/proc/*/fd` prefilter). `FDINFO_SAMPLE_MIN_NS = 500 ms` caps calls. Lunar Lake 228V, ~5000 open fds / ~90 drm fds.

@MrChuw, B580 % restored via `8ad127e` (per-class `max()` on `drm-total-cycles` instead of sum). Other commits fix late-fdinfo RC/MC activation (`f1ecf86`), gtidle false-100% on short ticks (`d244373`), plus dead-code / whitespace / style cleanup.

Deferred to a follow-up PR: multi-engine class capacity normalization (Battlemage VCS=2 / DG2 CCS=4 partial-use underreport).

src/linux/btop_collect.cpp
deveworld: about 24-48 ms per scan on Lunar Lake, throttled to once per 500 ms on the collector thread
Added a --debug line (Xe: fdinfo scan <ms>) to measure it on any device

deveworld:
Force-pushed, rebased onto current main (resolves the merge conflict), squashed to one commit, and integrated with the new AMD Asysfs backend.

New since last review:
- No root / render group needed. The DRM fd is now optional; if it can't be opened, VRAM is skipped but gtidle/fdinfo utilization and clock still work, so the GPU is shown instead of dropped.
- Power for discrete Arc via hwmon. Integrated GPUs expose no GPU power sensor (no hwmon, no PMU energy event), so iGPU power isn't available.
- fdinfo perf: about 24-48 ms/scan on Lunar Lake, throttled to 500 ms; the /proc scan no longer throws on unreadable processes; per-scan time logged at --debug.

Deferred: CCS / multi-engine capacity normalization; i915 still needs PMU permissions.

temidaradev:
looks accurate so far 
image

jlfranklin:
Looks good on a Debian/forky system:

```
CPU: 16-core Intel Core Ultra 7 356H (-MCP-)
speed/min/max: 400/400/4700:3700:3300 MHz Kernel: 7.1.8+deb14.1-amd64 x86_64
Up: 3d 17h 18m Mem: 9.15/15.03 GiB (60.9%) Storage: 476.94 GiB (16.7% used)
Procs: 703 Shell: Bash inxi: 3.3.41
```

image

stolk:
Works.

The usage seems to have a low pass filter on it, though? I hammer it going from 0 to 100 in a fraction of a second, but:

image_2026-08-29_150559434

Boladi888:
I've got a little different problem on my asrock B70 pro 32gb

On my system, btop can see the GPU and utilization, but wattage shows as `0.00W`.  The sensor itself is not dead though. It looks like power is available, just not in the form the current Xe path is expecting.

System:
0000:03:00.0 VGA compatible controller [0300]: Intel Corporation Battlemage G31 [Intel Graphics] [8086:e223]
kernel: 7.0.0-31-generic
btop: 1.4.6, GPU_SUPPORT=true

The relevant hwmon node looks like this:
/sys/class/hwmon/hwmon7/name: xe
/sys/class/hwmon/hwmon7/energy1_label: card
/sys/class/hwmon/hwmon7/energy1_input: <increments>
/sys/class/hwmon/hwmon7/energy2_label: pkg
/sys/class/hwmon/hwmon7/power1_label: card
/sys/class/hwmon/hwmon7/power1_cap: 275000000
/sys/class/hwmon/hwmon7/power1_input: missing
/sys/class/hwmon/hwmon7/power1_average: missing

Sampling energy1_input manually gives a real nonzero value. In one sample I got:
Battlemage card energy delta: 414 J over 2.01 s => 206.33 W

So the practical issue is not that Battlemage has no power data. It is that this card exposes cumulative energy for card, and the current path appears to look only for instantaneous power1_average or power1_input.

Would it make sense for the Xe backend to fall back to deriving watts from energy*_input where energy*_label == card when power1_average and power1_input are absent?  Basically sample energy twice, divide by elapsed time, and treat that as the same source of truth.  Power is not missing here.  It is just wearing the wrong sysfs hat.


For images: include one cropped screenshot of btop showing `GPU ... 0.00W`, but don’t use the full screenshot unless you’re fine with the visible process list and local IP. I’d also include a terminal screenshot or pasted text showing the `hwmon7` files and the manual `206 W` energy-delta calculation. That second one is the stronger evidence.

image

BBaoVanC:
I just now tested it on Arch Linux with my Intel Core Ultra 325 iGPU, and I don't see VRAM or power usage. On my desktop with an Arc B580, I see VRAM usage fine but it still doesn't show power. nvtop is able to see VRAM on both devices, and power on the B580 (still not on the Panther Lake iGPU, but that might not be btop's fault)

Issue #701, opened 2023-12-28
[REQUEST] Apple Silicon GPU? 🥺


aidan-gibson:
https://github.com/tlkh/asitop does this via [powermetrics](https://www.unix.com/man-page/osx/1/powermetrics/) (sudo required). I don't mind the sudo requirement, it seems unavoidable. 

This seemed like a viable sudoless alternative (https://github.com/op06072/NeoAsitop, which relies on https://github.com/dehydratedpotato/socpowerbud) but neither worked on my machine and it seems to be a widespread issue.

martinoturrina:
This would be so good!

emanuelez:
I have an implementation of this here: https://github.com/emanuelez/btop/tree/apple-silicon

It was LLM generated and I don't think it's fair to push for review some code that I don't fully understand. I was simply scratching an itch and this works for me so anybody please feel free to:

- review the code and make a proper PR. I don't expect any credit or attribution really.
- build it and run it. It works quite nicely for me but it does require sudo

martinoturrina:
> * but it does require sudo

Seems like macmon does most of the things needed without sudo, so maybe that could be interesting to take a look to see if it's possible to avoid requiring sudo for btop also

emanuelez:
Well spotted! Here's a sudo-less version that comes with the same disclaimers as my previous comment: https://github.com/emanuelez/btop/tree/gpu-apple-silicon-nosudo

<img width="1506" height="381" alt="Image" src="https://github.com/user-attachments/assets/1370fbcb-8202-48d4-8901-4ec280089dcb" />

martinoturrina:
To me the code looks ok, but I'm also not an expert. 

One concern is that usually GPU on linux is shown under the CPU cores in a more conceived way, the screen you posted is a bit different layout which is not bad but overall maybe is best to keep the standard layout

<img width="1252" height="224" alt="Image" src="https://github.com/user-attachments/assets/610abcb3-e078-4580-a4dd-a3e778a387b0" />

emanuelez:
I was just pressing 5 to show more of the GPU stats :)

<img width="1506" height="293" alt="Image" src="https://github.com/user-attachments/assets/dbce38d1-1491-4a3c-8b2f-280b5bf1df57" />

martinoturrina:
I think you can submit the PR and mark it as AI generated, I think is a good contribution even if the code was generated

Ruaneri-Portela:
[...]

PR #1458, opened 2026-01-10
Fixes and Improvement for Pause and Following modes

vandabbin:
Fixes: #1459

### Reselection Issues with Pause and Follow

- I found an edge case I had missed before for the very end of the process list (not view). For the process list pausing feature, if the process selected was the very last item in the whole list then the selection was moved up one. Now it shifts the list keeping the last item in view. And vice versa when the list is paused and the very last item is selected, unpausing would also move it up one item in the list. For the process following feature the same thing would happen when unfollowing (if not using the detailed view follow) but for the last few items (upto halfway up the view)
  - This was fixed by adding 1 to either `selected` or `start` when required
  

- When process following with detailed view. unfollowing by clicking the process and then clicking again to close detail view threw the view off because following was disengaged.
  - Fixed with the modification to the code block that handles following. It now also restores the selection and view when exiting detailed view also. (I had a different fix for this but the fix for the next issue made it irrelevant and was removed)

- If following a process with detailed view and it is one of the first or last few items in the whole list (not view) then if you change the reverse sort mode and then unfollow the process, instead of the selection returning to the process that was being followed it returns to where the process would have been if the sorting wasn't reversed.
  - I struggled to figure out a fix for this but landed on reusing the code block that locates the followed process to restore the selection when detailed view closes
  - I actually quite like this fix since it also centers the selection if possible when closing the detailed view.

- The change that removed selection for detailed view following caused an issue where if the list was paused holding `up` would cause the selection to loop from the list position where the followed process was after reaching the top of the list
  - This was fixed by adding some additional conditions before changing the selection back to the followed process when `selected` is 0
  - This gets rid of this issue so that when paused scrolling to the top of the list doesn't loop back to the followed process location

- When following a process and the view was at the end of the list, if the process died and following disengaged, a blank line would be left at the bottom of the list until the next collection cycle.
  - fixed this by making `select_max` not a const and adding 1 to it when the following mode disengaged.

- When the followed process is the first process in the list, pressing up doesn't redraw the screen and get rid of the following banner until the next collection cycle.

  - I fixed this by moving `bool changed = false` to the top of the selection function and setting `changed` to true when reselecting the followed process before moving back up and `selected` is 0 again.

- When following a process clicking on the banner with the mouse would exit following mode and select the last line. I am not sure if this is desirable or not.

  - I have fixed this with an extra condition in the mouse input logic
  - This can be reverted if you feel it is good for clicking the banner to exit following in that way.
<hr>

### Issue with tree expand/collapse when following

- When I made the change that set `selected = 0` when following the detailed process. I did not realize that this would prevent the user from being able to expand or collapse the tree of the process being followed without first unfollowing it.

  - I fixed this by slightly reworking the input code for expand/collapse with some extra conditions.
<hr>

### Improvements to follow mode list centering

- The followed process was centered using `select_max / 2`. this works but would leave it off center by 2 if `select_max` was odd.
  - it has been updated to use `select_max / 2` if select_max is even which has the followed process off center as close as it can be but slightly closer to the top of list. if select_max is odd then it uses `select_max / 2 + 1` which places the process dead center.

- If `follow-detailed` is disabled and a process is being followed while paused, opening or closing the detailed view doesn't recenter the followed process.
  - fixed this by relocating a line that sets `update_following` 
<hr>

### Misc improvements to follow mode

- Added a line in the `F` input handling code that sets the selection to the followed process before exiting follow mode. This means that if you press `F` after opening a detailed view and following that it will exit follow mode and move `selected` to the followed process location instead of leaving `selected` as 0. This is equivalent to pressing `down` after opening a detailed view when not following the process

- Since when following the detailed view process the selection can be moved up or down from the followed process, the up arrow for the select button should not be greyed out if following the detailed view process and `should_selection_return_to_followed` is false.

- Added a comment in the selection function to explain what the
  following mode code block does there.
<hr>

### Small optimization for Pause mode

- I moved `Config::getB("keep_dead_proc_usage")` outside of the for loop that goes through all the current procs while paused.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1561,6 +1561,7 @@ namespace Proc {
 	Draw::Graph detailed_mem_graph;
 	int user_size, thread_size, prog_size, cmd_size, tree_size;
 	int dgraph_x, dgraph_width, d_width, d_x, d_y;
+	bool previous_proc_banner_state = false;
 
 	string box;
 
@@ -1568,11 +1569,19 @@ namespace Proc {
 		auto start = Config::getI("proc_start");
 		auto selected = Config::getI("proc_selected");
 		auto last_selected = Config::getI("proc_last_selected");
+		bool changed = false;
 		int select_max = (Config::getB("show_detailed") ? (Config::getB("proc_banner_shown") ? Proc::select_max - 9 : Proc::select_max - 8) :
 																(Config::getB("proc_banner_shown") ? Proc::select_max - 1 : Proc::select_max));
 
+		// Return the selection from the detailed view to the followed process before moving the selection
+		// Disengage following mode when moving the selection unless paused
 		if (Config::getB("follow_process")) {
-			if (selected == 0) selected = Config::getI("proc_followed");;
+			if (Config::getB("show_detailed") and selected == 0 and Config::getB("should_selection_return_to_followed")
+			and Config::getI("detailed_pid") == Config::getI("followed_pid")) {
+				selected = Config::getI("proc_followed");
+				Config::set("should_selection_return_to_followed", false);
+				changed = true;
+			}
 			if (not Config::getB("pause_proc_list")) {
 				Config::flip("follow_process");
 				Config::set("followed_pid", 0);
@@ -1625,7 +1634,6 @@ namespace Proc {
 			start = clamp((int)round((double)mouse_y * (numpids - select_max - 2) / (select_max - 2)), 0, max(0, numpids - select_max));
 		}
 
-		bool changed = false;
 		if (start != Config::getI("proc_start")) {
 			Config::set("proc_start", start);
 			changed = true;
@@ -1653,13 +1661,14 @@ namespace Proc {
 		auto follow_process = Config::getB("follow_process"); 
 		int followed_pid = Config::getI("followed_pid");
 		int followed = Config::getI("proc_followed");
+		bool should_selection_return_to_followed = Config::getB("should_selection_return_to_followed");
 		auto proc_banner_shown = pause_proc_list or follow_process;
 		Config::set("proc_banner_shown", proc_banner_shown);
 		start = Config::getI("proc_start");
 		selected = Config::getI("proc_selected");
 		const int y = show_detailed ? Proc::y + 8 : Proc::y;
 		const int height = show_detailed ? Proc::height - 8 : Proc::height;
-		const int select_max = show_detailed ? (proc_banner_shown ? Proc::select_max - 9 : Proc::select_max - 8) : 
+		int select_max = show_detailed ? (proc_banner_shown ? Proc::select_max - 9 : Proc::select_max - 8) : 
 												(proc_banner_shown ? Proc::select_max - 1 : Proc::select_max);
 		auto totalMem = Mem::get_totalMem();
 		int numpids = Proc::numpids;
@@ -1668,13 +1677,15 @@ namespace Proc {
 		out.reserve(width * height);
 
 		//? Move current selection/view to the selected process when a process should be followed
-		if (follow_process and (not pause_proc_list or Config::getB("update_following"))) {
+		//? Restore view and selection to the detailed view process when detailed view is closed
+		const int restore_detailed_pid = Config::getI("restore_detailed_pid");
+		if ((follow_process and (not pause_proc_list or Config::getB("update_following"))) or restore_detailed_pid > 0) {
 			Config::set("update_following", false);
 			int loc = 1;
 			bool can_follow = false;
 			for (auto& p : plist) {
 				if (p.filtered or (proc_tree and p.tree_index == plist.size())) continue;
-				if (p.pid == (size_t)followed_pid) {
+				if (p.pid == (size_t)(restore_detailed_pid > 0 ? restore_detailed_pid : followed_pid)) {
 					can_follow = true;
 					break;
 				}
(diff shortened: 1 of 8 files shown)


vandabbin:
So with many hours of fiddling around trying to find any bugs with the Pause or Following modes, I believe that I have found them all. Though I could be wrong. (edit: I was)

This PR addresses all of the remaining issues that I found except for one that I am still trying to figure out the best way to solve.

The one remaining issue is that if following a process with detailed view and it is one of the first or last few items in the whole list (not view) then if you change the reverse sort mode and then unfollow the process, instead of the selection returning to the process that was being followed it returns to where the process would have been if the sorting wasn't reversed. I've tried several different ideas for fixing this but all of them caused further issues or just plain didn't work 😆. So I am still working on it.

vandabbin:
Ok so I figured out a fix for that last remaining issue. I decided to reuse the code block that follows the process to also restore it when detailed view is closed. With this I don't know of any more issues with the follow or pause modes. I will continue to search for them though. (edit: I found some)

vandabbin:
I apologize for the number of additional fix PRs I have made for the following mode.

vandabbin:
hmm it seems something i changed introduced an issue. I need to fix it before this is ready.

Im not quite sure what is causing it yet. must be some sort of race condition?

if you repeatedly press enter on the same process (opening and closing the detailed view) the selection eventually stays at 0 instead of returning to the selection......

edit:

Ahh!! I figured it out. I should have had `Config::set("restore_detailed_pid", 0);` in an else block

vandabbin:
> Ahh!! I figured it out. I should have had `Config::set("restore_detailed_pid", 0);` in an else block

of course it wasn't actually an `else` that was needed...but a seperate `if`. I had a bit of a brainfart.

vandabbin:
I am debating changing `proc_banner_shown` to `proc_banner_state` to better match the `prev_proc_banner_state` variable that was added. Thoughts?

src/linux/btop_collect.cpp
```diff
@@ -3170,14 +3170,16 @@ namespace Proc {
...
+					if (rng::find(dead_procs, r.pid) == dead_procs.end()) {
+						if (rng::find(found, r.pid) == found.end()) {
+							r.death_time = round(uptime) - (r.cpu_s / Shared::clkTck);
+							r.state = 'X';
+							dead_procs.emplace(r.pid);
+							//? Reset cpu usage for dead processes if paused and option is set
+							if (!Config::getB("keep_dead_proc_usage")) {
+								r.cpu_p = 0.0;
+								r.mem = 0;
+							}
```
vandabbin: Thinking about this again though I thought `dead_procs.emplace(r.pid);` would only add to `dead_procs` if that pid is not already in `dead_procs`. And since that code is only run if the pid isn't found in the set of newly found process pids and the only pids being checked are in `current_procs` which doesn't get any new processes added to it while paused, I don't really understand how `dead_procs` is growing out of control? Am I missing something or is this issue that is mentioned by someone else not actually an issue? If that is the case I should remove this change because I'm not sure that it is actually any better than the previous implementation

https://github.com/aristocratos/btop/issues/701#issuecomment-3742871313
Screenshot 2026-01-13 at 4 12 14 AM

edit: I'm realizing now that I think that issue list was generated by an LLM so I'm not sure if this is even real haha

vandabbin: Hmm...I asked github's copilot my first question and it agrees with me that even with the original implementation `dead_procs` should never be able to grow larger than `current_procs` and `current_procs` shouldn't get larger while paused.

of course LLMs have the tendency to agree with the person using them so I'm still not really totally sure....

vandabbin: I've decided to remove this change. I don't think it is necessary. And I can't figure out how dead_procs is growing like claimed

src/btop_draw.cpp
```diff
@@ -1860,11 +1887,19 @@ namespace Proc {
...
-			out += Mv::to(y + height - 1, x + 1) + title_left_down + Fx::b + hi_color + Symbols::up + Theme::c("title") + " select " + down_button + Fx::ub + title_right_down
-				+ title_left_down + Fx::b + t_color + "info " + hi_color + Symbols::enter + Fx::ub + title_right_down;
+			fmt::format_to(std::back_inserter(out), 
+				"{move}{title_left}{bold}{up_button}{select_color}{select}{down_button}{unbold}{title_right}"
+				"{title_left}{bold}{t_color}{info}{hi_color}{enter_symbol}{unbold}{title_right}",
+				"move"_a = Mv::to(y + height - 1, x + 1), "bold"_a = Fx::b, "unbold"_a = Fx::ub,
+				"title_left"_a = title_left_down, "title_right"_a = title_right_down,
+				"select_color"_a = Theme::c("title"), "t_color"_a = t_color, "hi_color"_a = hi_color,
+				"up_button"_a = up_button, "down_button"_a = down_button, "enter_symbol"_a = Symbols::enter,
+				"select"_a = " select ", "info"_a = "info ");
```
vandabbin: I made this change because I had to modify it to add the `up_button` variable and I know there is a goal to replace all these with the use of the fmt library

src/btop_draw.cpp
```diff
@@ -1860,11 +1887,18 @@ namespace Proc {
...
 			int mouse_x = x + 14;
-			out += Mv::to(y + height - 1, x + 1) + title_left_down + Fx::b + hi_color + Symbols::up + Theme::c("title") + " select " + down_button + Fx::ub + title_right_down
-				+ title_left_down + Fx::b + t_color + "info " + hi_color + Symbols::enter + Fx::ub + title_right_down;
+			fmt::format_to(std::back_inserter(out), 
+				"{move}{title_left}{bold}{up_button}{select_color}select{down_button}{unbold}{title_right}"
+				"{title_left}{bold}{t_color}info{hi_color}{enter_symbol}{unbold}{title_right}",
+				"move"_a = Mv::to(y + height - 1, x + 1), "bold"_a = Fx::b, "unbold"_a = Fx::ub,
+				"title_left"_a = title_left_down, "title_right"_a = title_right_down,
+				"select_color"_a = Theme::c("title"), "t_color"_a = t_color, "hi_color"_a = hi_color,
+				"up_button"_a = up_button, "down_button"_a = down_button, "enter_symbol"_a = Symbols::enter);
```
vandabbin: I made this change because I had to modify it to add the up_button variable and I remember you saying fmt format should be used for new draw operations, this one isn't new but it was modified so I figured I'd do this. I can revert it if undesirable

vandabbin: Well I discovered that for some reason using `fmt::format_to` here messes up the button mappings. It looks right to me but even when the buttons are highlighted on the bottom bar they only occasionally do what they are supposed to when you click them, and the rest of the time it sets the selection to 0. This is happening for all the buttons after the info button. I am going to remove this change

vandabbin:
```c
fmt::format_to(std::back_inserter(out), "{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}",
	Mv::to(y + height - 1, x + 1), title_left_down, Fx::b, hi_color, up_button,
	Theme::c("title"), " select ", down_button, Fx::ub, title_right_down, title_left_down,
	Fx::b, t_color, "info ", hi_color, Symbols::enter, Fx::ub, title_right_down);
```

I tried a simpler version like this and the same thing. I wonder why that is

src/btop_input.cpp
```diff
@@ -395,6 +399,8 @@ namespace Input {
 									process("enter");
 									return;
 								}
+								else if (Config::getB("proc_banner_shown") and line == y + height - 2)
+									return;
```
vandabbin, line 403: When following a process clicking on the banner with the mouse would exit following mode and select the last line. I am not sure if this is desirable or not. This can be reverted if you feel it is good for clicking the banner to exit following in that way. Although this also has the benefit of preventing clicking the banner in pause mode from scrolling the view down one line but i could just change `proc_banner_shown` to `pause_proc_list`

aristocratos: That's fine by me if you want it to function that way, don't really have an opinion on it.

aristocratos:
Is the PR finished?

vandabbin:
> Is the PR finished?

Yup I haven't been able to find any more issues

Issue #1461, opened 2026-01-10
Down button arrow doesn't update highlighting when it should


vandabbin:
**Describe the bug**

When moving the selection to the end of the list the down arrow highlighting doesn't change until something else sets redraw to true

**To Reproduce**

move selection to last process in list. try moving up or down from there and see the highlighting not change. 
if you change the sorting order you can make it change.

**Expected behavior**

the highlighting should update properly when the selection reaches or leaves the last process in the list.

PR #1462, opened 2026-01-10
Fix down arrow highlighting

vandabbin:
Fixes: #1461 

The down arrow for the select button on the bottom bar of the process list needs `redraw = true;` in order to properly refresh highlighting.

However moving the selection around does not redraw that section.

This fix sets redraw to true if the selection reaches or leaves the last process on the list.

when merged with #1458 this block of code should be the one directly above the redraw block and below the selection edge cases block

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1548,6 +1548,7 @@ namespace Proc {
 	int x, y, width = 20, height;
 	int start, selected, select_max;
 	bool shown = true, redraw = true;
+	bool is_last_process_in_list = false;
 	int selected_pid = 0, selected_depth = 0;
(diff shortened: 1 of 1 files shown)


vandabbin:
@deckstose 

> Nit: I think the name of `at_list_end` can be more descriptive.

ok I will change it to something better

what do you think of `selected_last_process` or `last_process_in_list`?

deckstose:
I always prefix bools with `is_`

vandabbin:
@deckstose 

> I always prefix bools with `is_`

cool. I've changed it to `is_last_process_in_list`

deckstose (approved):

PR #1469, opened 2026-01-13
Add ROCm SMI library version 1.x support

vedantadhobley:
Ubuntu 24.04's rocm-smi package (5.7.0-1) reports library version as 1.0.0 via rsmi_version_get(), causing AMD GPU detection to fail.

This treats version 1.x the same as version 6/7, using the v6 API.

Tested on:
- AMD Ryzen AI Max+ 395 / Radeon 8060S (Framework Desktop)
- Ubuntu 24.04.1 LTS


src/linux/btop_collect.cpp
```diff
@@ -1559,7 +1559,7 @@ namespace Gpu {
 				if ((rsmi_dev_gpu_clk_freq_get_v5 = (decltype(rsmi_dev_gpu_clk_freq_get_v5))load_rsmi_sym("rsmi_dev_gpu_clk_freq_get")) == nullptr)
 					return false;
 			// In the release tarballs of rocm 6.0.0 and 6.0.2 the version queried with rsmi_version_get is 7.0.0.0
-			} else if (version.major == 6 || version.major == 7) {
+			} else if (version.major == 6 || version.major == 7 || version.major == 1) {
 				if ((rsmi_dev_gpu_clk_freq_get_v6 = (decltype(rsmi_dev_gpu_clk_freq_get_v6))load_rsmi_sym("rsmi_dev_gpu_clk_freq_get")) == nullptr)
 					return false;
```
aristocratos, line 1564: If it's version 5.7.0 that reports `1`, why is it added to the branch for RSMI version 6?

deckstose: I've tested this on a spare machine and it worked for me... If we get another issue for this we might need to differentiate between ROCm 5 and 7, which both seem to report version 1.x (???).

PR #1470, opened 2026-01-13
Apple Silicon GPU support 

Ruaneri-Portela:
With help from the folks at Issues #701 , I organized it using only IOKit, which provides the data without root.

It has only been tested on my M4 Pro platform, so anyone willing to test it would be very welcome.

Apparently, we can use the IOReport framework to obtain energy data for the CPU, such as usage in watts and clock speed.

We can also use the same framework to add the ANE (Apple Neural Engine). 

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(BTOP_LTO)
 endif()
 
 if(APPLE)
-  target_sources(libbtop PRIVATE src/osx/btop_collect.cpp src/osx/sensors.cpp src/osx/smc.cpp)
+  target_sources(libbtop PRIVATE src/osx/btop_collect.cpp src/osx/sensors.cpp src/osx/smc.cpp src/osx/iokit.cpp src/osx/gpu.cpp)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR CMAKE_SYSTEM_NAME STREQUAL "MidnightBSD")
   target_sources(libbtop PRIVATE src/freebsd/btop_collect.cpp)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
@@ -161,35 +161,36 @@ find_package(Threads REQUIRED)
 target_link_libraries(libbtop Threads::Threads)
 
 # Enable GPU support
-if(LINUX AND BTOP_GPU)
+if(BTOP_GPU AND (LINUX OR APPLE))
   target_compile_definitions(libbtop PUBLIC GPU_SUPPORT)
+  if(LINUX)
+    if(BTOP_RSMI_STATIC)
+      # ROCm doesn't properly add it's folders to the module path if `CMAKE_MODULE_PATH` is already
+      # set
+      # We could also manually append ROCm's path here
+      set(_CMAKE_MODULE_PATH CMAKE_MODULE_PATH)
+      unset(CMAKE_MODULE_PATH)
 
-  if(BTOP_RSMI_STATIC)
-    # ROCm doesn't properly add it's folders to the module path if `CMAKE_MODULE_PATH` is already
-    # set
-    # We could also manually append ROCm's path here
-    set(_CMAKE_MODULE_PATH CMAKE_MODULE_PATH)
-    unset(CMAKE_MODULE_PATH)
+      # NOTE: This might be problematic in the future if other sub projects depend on this or if
+      # btop starts producing libraries
+      # Build a static ROCm library
+      set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
 
-    # NOTE: This might be problematic in the future if other sub projects depend on this or if
-    # btop starts producing libraries
-    # Build a static ROCm library
-    set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
+      add_subdirectory(lib/rocm_smi_lib EXCLUDE_FROM_ALL)
 
-    add_subdirectory(lib/rocm_smi_lib EXCLUDE_FROM_ALL)
+      add_library(ROCm INTERFACE)
+      # Export ROCm's properties to a target
+      target_compile_definitions(ROCm INTERFACE RSMI_STATIC)
+      target_include_directories(ROCm INTERFACE lib/rocm_smi_lib/include)
+      target_link_libraries(ROCm INTERFACE rocm_smi64)
 
-    add_library(ROCm INTERFACE)
-    # Export ROCm's properties to a target
-    target_compile_definitions(ROCm INTERFACE RSMI_STATIC)
-    target_include_directories(ROCm INTERFACE lib/rocm_smi_lib/include)
-    target_link_libraries(ROCm INTERFACE rocm_smi64)
+      set(CMAKE_MODULE_PATH _CMAKE_MODULE_PATH)
 
-    set(CMAKE_MODULE_PATH _CMAKE_MODULE_PATH)
-
-    target_link_libraries(libbtop ROCm)
-  endif()
-  if(NOT BTOP_STATIC)
-    target_link_libraries(libbtop ${CMAKE_DL_LIBS})
+      target_link_libraries(libbtop ROCm)
+    endif()
+    if(NOT BTOP_STATIC)
+      target_link_libraries(libbtop ${CMAKE_DL_LIBS})
+    endif()
   endif()
 endif()
 
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -4,7 +4,7 @@
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
 
-	   http://www.apache.org/licenses/LICENSE-2.0
+           http://www.apache.org/licenses/LICENSE-2.0
 
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,58 +16,65 @@ indent = tab
 tab-size = 4
 */
 
+#include <cstddef>
+
 #include <Availability.h>
-#include <CoreFoundation/CoreFoundation.h>
-#include <IOKit/IOKitLib.h>
 #include <arpa/inet.h>
 #include <libproc.h>
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <IOKit/IOKitLib.h>
 #include <mach/mach.h>
 #include <mach/mach_host.h>
 #include <mach/mach_init.h>
+#include <mach/mach_time.h>
 #include <mach/mach_types.h>
 #include <mach/processor_info.h>
 #include <mach/vm_statistics.h>
-#include <mach/mach_time.h>
 // BUGS
 //     If both <net/if.h> and <ifaddrs.h> are being included, <net/if.h> must be
 //     included before <ifaddrs.h>.
 // from: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getifaddrs.3.html
+#include "../btop_config.hpp"
+#include "../btop_log.hpp"
+#include "../btop_shared.hpp"
+#include "../btop_tools.hpp"
+
+#include <cmath>
+#include <cstring>
+#include <fstream>
+#include <mutex>
+#include <numeric>
+#include <ranges>
+#include <stdexcept>
+#include <string>
+#include <unordered_set>
+#include <utility>
+
 #include <net/if.h>
 #include <ifaddrs.h>
 #include <net/if_dl.h>
 #include <netdb.h>
+#include <netinet/in.h> // for inet_ntop
 #include <netinet/tcp_fsm.h>
 #include <pwd.h>
 #include <sys/socket.h>
 #include <sys/statvfs.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
-#include <netinet/in.h> // for inet_ntop
 #include <unistd.h>
-#include <stdexcept>
-#include <utility>
-
-#include <cmath>
-#include <fstream>
-#include <mutex>
-#include <numeric>
-#include <ranges>
-#include <regex>
-#include <string>
-#include <unordered_set>
 
 #include <fmt/format.h>
 
-#include "../btop_config.hpp"
-#include "../btop_log.hpp"
-#include "../btop_shared.hpp"
-#include "../btop_tools.hpp"
-
 #if __MAC_OS_X_VERSION_MIN_REQUIRED > 101504
-#include "sensors.hpp"
+#    include "sensors.hpp"
 #endif
+#include "gpu.hpp"
+#include "iokit.hpp"
 #include "smc.hpp"
 
+<<<<<<< HEAD
+=======
 #if defined(GPU_SUPPORT)
 #include <dlfcn.h>
 #include <mach/mach_time.h>
@@ -115,6 +122,7 @@ extern "C" {
 #define kIOMainPortDefault kIOMasterPortDefault
 #endif
 
+>>>>>>> official
 using std::clamp, std::string_literals::operator""s, std::cmp_equal, std::cmp_less, std::cmp_greater;
 using std::ifstream, std::numeric_limits, std::streamsize, std::round, std::max, std::min;
 namespace fs = std::filesystem;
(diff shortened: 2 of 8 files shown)


deckstose:
So this adds a powemetrics api, and then removes it?

btop.desktop
```diff
@@ -5,12 +5,15 @@ Name=btop++
...
 Comment[it]=Monitoraggio delle risorse: mostra utilizzo e statistiche per CPU, dischi, rete e processi
 Comment[zh_TW]=顯示處理器、主記憶體、磁碟、網路與進程的使用與統計數據的資源監視器應用程式
+Comment[pt_BR]=Monitoramente de atividades com estatisticas e graficos por CPU, disco, rede e processos
 Icon=btop
 Exec=btop
 Terminal=true
 Categories=System;Monitor;ConsoleOnly;
 Keywords=system;process;task
 Keywords[zh_TW]=系統;進程;處理程序;任務
+Keywords[pt_BR]=sistema;processos;tarefas
```
deckstose, line 19: Please open a different PR for that. I can merge that immediately

deckstose (changes requested):
Most of the code is not properly formatted.

There is a huge amount of AI assisted code in there, this is what I was getting from the issue comments (?).

Most of the code needs _COMMENTS_. Especially the function pointer magic.

Ruaneri-Portela:
I had kept the power metrics to read the CPU data. But I think it's outside the scope to just add the GPU. I removed it for now.

With a little work, you can get the data via IOKit.

I took Emanuele Zattin's code, which was allegedly written by him using AI, and reworked some things that I thought were relevant.

I can redo it by scheduling with AI.

Ruaneri-Portela:
I added some comments, but I think it might not be enough, and I haven't formatted it properly yet.

deckstose:
I don't care about the boolean operators `! == not` etc, it's more about indentation, line length and naming (`snake_case`) functions.

Also IMO comments should not include the question mark (although it's done in other parts of the code). It's fine either way, but redundant in a way. And I'm not sure if editors with their commenting feature support it.

vandabbin:
> Also IMO comments should not include the question mark (although it's done in other parts of the code). It's fine either way, but redundant in a way. And I'm not sure if editors with their commenting feature support it.

whoops that how I was commenting code here because almost all the comments I already saw in the code had it. I will keep that in mind in the future

deckstose:
No worries, I have merged code with this style

deckstose:
Please mark this ready for review when this _is_ ready

Ruaneri-Portela:
I made the changes, but I'm not sure if it's any good. 

If you could take a look

deckstose:
I will take a look but it's not going to be anytime soon

aristocratos:
Seems to work as intended on a Mac Mini M1 running Tahoe.

But would be great to have some testing done on other Mac hardware.

What of the code is AI generated? Because that code would need to be checked so it hasn't been "lifted" from some other project (and might have missing attribution or incompatible licensing).

Ruaneri-Portela:
> Seems to work as intended on a Mac Mini M1 running Tahoe.
> 
> But would be great to have some testing done on other Mac hardware.
> 
> What of the code is AI generated? Because that code would need to be checked so it hasn't been "lifted" from some other project (and might have missing attribution or incompatible licensing).


My hardware is an M4 Pro, and that's where I tested it.

Regarding the code being marked by AI, and because of Emanuele Zattin's contribution, parts of what the code estimates to be CPU usage were inspired by AI. I say inspired because I didn't request a code per se, but rather the procedure.

Our reference code is https://github.com/dehydratedpotato/socpowerbud
Whose code is MIT.

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1461,118 @@ namespace Tools {
 		return 0.0;
 	}
 }  // namespace Tools
+
+
+namespace Gpu {
+#ifdef GPU_SUPPORT
```
deckstose, line 2143: Namespace should be inside ifdef

src/osx/btop_collect.cpp
```diff
@@ -68,16 +69,16 @@ tab-size = 4
...
+#include "iokit.hpp"
+#include "gpu.hpp"
 
 using std::clamp, std::string_literals::operator""s, std::cmp_equal, std::cmp_less, std::cmp_greater;
 using std::ifstream, std::numeric_limits, std::streamsize, std::round, std::max, std::min;
 namespace fs = std::filesystem;
 namespace rng = std::ranges;
 using namespace Tools;
 
+
```
deckstose, line 168: Needless newline.

src/osx/btop_collect.cpp
```diff
@@ -68,16 +69,16 @@ tab-size = 4
 #endif
 #include "smc.hpp"
```
deckstose, line 75: Needless newline.

src/osx/btop_collect.cpp
```diff
@@ -191,6 +210,33 @@ namespace Shared {
 		//? Init for namespace Mem
 		Mem::old_uptime = system_uptime();
 		Mem::collect();
+
+		#ifdef GPU_SUPPORT
```
deckstose: Please put macros at the start of the line.

src/osx/btop_collect.cpp
```diff
@@ -301,17 +347,18 @@ namespace Cpu {
...
 		int mib[] = {CTL_HW, HW_CPU_FREQ};
 
 		if (sysctl(mib, 2, &freq, &size, nullptr, 0) < 0) {
-			// this fails on Apple Silicon macs. Apparently you're not allowed to know
 			return "";
 		}
-		return std::to_string(freq / 1000.0 / 1000.0 / 1000.0).substr(0, 3);
+
+		// Convert from Hz to GHz and return a numeric string without units
+		const double freq_ghz = static_cast<double>(freq) / 1000000000.0;
```
deckstose:
```suggestion
		const auto freq_ghz = static_cast<double>(freq) / 1'000'000'000.0;
```

src/osx/btop_collect.cpp
```diff
@@ -638,10 +660,12 @@ namespace Mem {
 										while (cmp_greater(disk.io_activity.size(), width * 2)) disk.io_activity.pop_front();
 									}
 								}
-								CFRelease(properties);
 							}
 						}
 					}
+
+					if(properties)
+						CFRelease(properties);
```
deckstose: This screems to me like it wants to be put inside a RAII guard.

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
 		return 0.0;
 	}
 }  // namespace Tools
+
+
+namespace Gpu {
+
+#ifdef GPU_SUPPORT
```
deckstose, line 2143: Same as above.

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
...
+    namespace IOAccelerator {
+        //? Initialize Apple Silicon GPU monitoring Although the chips always have 1 GPU, I assume we can reuse them later on Intel Macs.
+        bool init() {
+            const size_t index = gpus.size();
```
deckstose:
```suggestion
            const auto index = gpus.size();
```

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
...
+            auto & io_gpus = io_gpu.get_gpus();
```
deckstose:
```suggestion
            auto& io_gpus = io_gpu.get_gpus();
```

Left hand pointer and references.

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
...
+
+			if (io_gpus.empty()) {
+				 return false;
+			}
+               
+			initialized = true;
+
+			for(size_t i = 0; i < io_gpus.size() ; i++){
+				gpus.emplace_back();  
+				std::string name = fmt::format("{} ({})", io_gpus[i].get_name(), io_gpus[i].get_core_count());
```
deckstose:
```suggestion
				auto name = fmt::format("{} ({})", io_gpus[i].get_name(), io_gpus[i].get_core_count());
```

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
...
+				collect<true>(&gpus[index], i);
+			}
+            return true;
+        }
+
+        //? Shutdown Apple Silicon GPU monitoring
+        bool shutdown() {
+            initialized = false;
+            return true;
+        }
```
deckstose, line 2168: Maybe I'm missing something but the return type feels useless here.

src/osx/btop_collect.cpp
```diff
@@ -1438,3 +1462,154 @@ namespace Tools {
...
+			}
+
+			io_gpus[index].refresh();
+			auto gpu_data = io_gpus[index].get_statistics();
+
+			gpus_slice->gpu_percent.at("gpu-totals").push_back(gpu_data.device_utilization);
+
+			gpus_slice->mem_total = gpu_data.alloc_system_memory;
+			gpus_slice->mem_used  = gpu_data.in_use_system_memory;
+			long long mem_percent = 0;
```
deckstose:
```suggestion
			auto mem_percent = 0LL;
```

src/osx/gpu.cpp
```diff
@@ -0,0 +1,670 @@
...
+#include "iokit.hpp"
+
+//? Others
+static uint64_t get_time_ns() {
+    static mach_timebase_info_data_t timebase = {0, 0};
+    if (timebase.denom == 0) {
+        mach_timebase_info(&timebase);
+    }
+    return mach_absolute_time() * timebase.numer / timebase.denom;
+}
```
deckstose: Can you explain why this is here? It's not immediately obvious to me why we need this over the C++ chrono facilities.

src/osx/gpu.cpp
```diff
@@ -0,0 +1,670 @@
...
+            }
+        }
+        CFRelease(services);
+    }
+    CFRelease(matching);
+    CFRelease(system);
+    return gpu_temp;
+}
+
+void GPUActivities::map_key_to_usage_number(GPUActivities::Usage &usage, const std::string &key, int64_t value) {
```
deckstose:
```suggestion
void GPUActivities::map_key_to_usage_number(GPUActivities::Usage& usage, const std::string& key, int64_t value) {
```

src/osx/gpu.cpp
```diff
@@ -0,0 +1,670 @@
...
+    static const std::unordered_map<std::string, int64_t GPUActivities::Usage::*> map = {
+            {"accumulatedGPUTime", &GPUActivities::Usage::accumulated_gpu_time},
+            {"lastSubmittedTime", &GPUActivities::Usage::last_submitted_time},
+    };
+
+    if (auto it = map.find(key); it != map.end()) {
+        usage.*(it->second) = value;
```
deckstose: This needs to be rewritten in a more readable syntax.

src/osx/gpu.cpp
```diff
@@ -0,0 +1,670 @@
...
+    const auto pidPos = s.find("pid ");
+    const auto commaPos = s.find(',');
+
+    if (pidPos == std::string::npos or commaPos == std::string::npos) return;
+
+    try {
+        proc = std::stoul(s.substr(pidPos + 4, commaPos - (pidPos + 4)));
+        name = s.substr(commaPos + 2);
+    } catch (...) {
+        return;
```
deckstose: Missing log?

src/osx/gpu.cpp
```diff
@@ -0,0 +1,670 @@
...
+            CFIndex count = CFDictionaryGetCount(usage_stats);
+
+            std::vector<const void *> keys(count);
+            std::vector<const void *> values(count);
+
+            CFDictionaryGetKeysAndValues(usage_stats, keys.data(),
+                                         values.data());
+            GPUActivities::Usage new_usage{};
+            for (CFIndex i = 0; i < count; ++i) {
+                CFStringRef keyRef = static_cast<CFStringRef>(keys[i]);
```
deckstose:
```suggestion
                auto key_ref = static_cast<CFStringRef>(keys[i]);
```

everywhere please.

CMakeLists.txt
```diff
@@ -193,6 +193,10 @@ if(LINUX AND BTOP_GPU)
   endif()
 endif()
 
+if(APPLE AND BTOP_GPU)
+target_compile_definitions(libbtop PUBLIC GPU_SUPPORT)
+endif()
```
deckstose: Combine this branch with the Linux branch with `GPU_SUPPORT` defined

deckstose (changes requested):
I've pointed out a couple of patterns, they should apply everywhere.

I can't say much about the actual implementation, but it looks good over all.

@aristocratos 

I wonder if it is necessary to have a struct indicating no GPU support, the undefined `GPU_SUPPORT` macro should be enough, right?

https://github.com/aristocratos/btop/blob/0c11d7349497fffcb70a693c71a508f38711398c/src/btop_shared.hpp#L192-L196

_EDIT:_ Resolved in https://github.com/aristocratos/btop/pull/1495.

vandabbin:
> Seems to work as intended on a Mac Mini M1 running Tahoe.
> 
> But would be great to have some testing done on other Mac hardware.

I tested it just now on my M1 Macbook Pro running 15.7.3 and it also seems to be working. I don't have any newer apple hardware to test with though.

However maybe I'm wrong and its just apple silicon proving how efficient it is but I have trouble believing that I was getting 1 watt power usage playing 4k video. It fluctuated between .73 and 1.03

Screenshot 2026-01-19 at 3 43 34 PM

also I'm not so sure that the power meter is displaying very nicely? Those numbers just don't seem to match well. when the usage drops the meter spikes. and then half the meter is .05 watts but the full meter is 1 watt? If it is supposed to scale like that It would be nice if there was some sort of indicator. (ex. the way net has numbers to indicate the scale)

<details>
<summary><b>video example</b></summary>

https://github.com/user-attachments/assets/d6ad5eb8-d9da-4f37-853b-7a101adbbd89
</details>

Ruaneri-Portela:
> > Seems to work as intended on a Mac Mini M1 running Tahoe.
> > But would be great to have some testing done on other Mac hardware.
> 
> I tested it just now on my M1 Macbook Pro running 15.7.3 and it also seems to be working. I don't have any newer apple hardware to test with though.
> 
> However maybe I'm wrong and its just apple silicon proving how efficient it is but I have trouble believing that I was getting 1 watt power usage playing 4k video. It fluctuated between .73 and 1.03
> 
> Screenshot 2026-01-19 at 3 43 34 PM
> also I'm not so sure that the power meter is displaying very nicely? Those numbers just don't seem to match well. when the usage drops the meter spikes. and then half the meter is .05 watts but the full meter is 1 watt? If it is supposed to scale like that It would be nice if there was some sort of indicator. (ex. the way net has numbers to indicate the scale)
> 
>  Screen.Recording.2026-01-19.at.3.47.45.PM.mov

This metric in W is only for the GPU module, excluding use for CPU and ANE, which also need to be computed.

Considering that your video is actually running on Apple's decode ASICs and not on the GPU shaders, I see it as a correct value.

src/osx/btop_collect.cpp
Ruaneri-Portela: CF objects are pointer types from OBJ-C, so RAII does not work for them. We need to act in a C-like manner.

If I'm not mistaken

vandabbin:
> This metric in W is only for the GPU module, excluding use for CPU and ANE, which also need to be computed.
> 
> Considering that your video is actually running on Apple's decode ASICs and not on the GPU shaders, I see it as a correct value.

Yeah you are probably right about that.

I do still think the meter shouldn't scale as the value gets smaller but maybe that's just me. It looks super weird to be jumping from half to full for a value change of 0.05w to 0.09w. I would expect both those values to only produce 1 bar on the meter if even that. I don't usually have the GPU box open though on my other systems (relying mainly on the small GPU meter in the CPU box) so maybe that is how they all do it?

edit: just checked on my framework laptop and the gpu power meter does not scale like that on amd gpu. (picture is with super tux kart running in the background) And it idles around 5w or so

Screenshot_20260120_014925


regarding scaling of the power meter. I think it would make sense for it to be scaled based on the total power usage ever expected from a given gpu model

src/osx/btop_collect.cpp
Ruaneri-Portela: I copied this from the way it was used in the Intel GPU implementation. I thought it was necessary. Can the initializer or collector be executed before a first start?

I ended up leaving it the same way.

src/osx/gpu.cpp
Ruaneri-Portela: I kept that from the code I worked on before. I can put everything in std::chrono.

deckstose:
You can use https://github.com/aristocratos/btop/pull/1146 for a .clang-format file so you can clang-format on the new files

src/osx/btop_collect.cpp
deckstose: This depends on https://github.com/aristocratos/btop/pull/1495

src/osx/btop_collect.cpp
aristocratos: Why would it not?
Just create a RAII wrapper class that takes a `CFMutableDictionaryRef` reference (or pointer) in the constructor. And have the destructor check if it's valid and run CFRelease on it.

deckstose:
CI is failing

Ruaneri-Portela:
> CI is failing

Fixed

Issue #1472, opened 2026-01-14
There is an extra mouse_mapping erasure


vandabbin:
When the hide and enter buttons were modified to allow them both to work at the same time a line was also added to erase the `info_enter` mapping.  #1428 

Later when the ~down arrow highlighting~ vim keys mouse mapping for kill fix was added, the loop that erases mouse mappings was updated and fixed with the proper mapping names. #1454

Because of this the earlier added erase for `info_enter` was no longer needed but was forgotten about. 

This doesn't affect anything but it doesn't need to be there.

PR #1473, opened 2026-01-14
Remove redundant mouse_mapping erasure

vandabbin:
Fixes: #1472 

There isn't really anything to say here that hasn't been said in the issue #1472 

There is an extra `mouse_mapping.erase` for `info_enter` and this removes it.


src/btop_draw.cpp
vandabbin, line 1718: `info_enter` is erased here and doesn't need to be erased again

Issue #1475, opened 2026-01-15
A wider process list when net or mem are shown


vandabbin:
**Is your feature request related to a problem? Please describe.**

When the net or mem boxes are shown, the proc box is more narrow then I would like sometimes.

**Describe the solution you'd like**

A menu option toggle to have a wider proc box when mem or net are shown. The would set both net and mem boxes to their minimum widths and the process list fills the remaining space

**Describe alternatives you've considered**

I've considered a multiple choice menu option to have various widths.

**Additional context**

<img width="1552" height="987" alt="Image" src="https://github.com/user-attachments/assets/c59c1517-eeb0-40d4-a8e1-522f2e078f03" />

PR #1476, opened 2026-01-15
Feat: Adjustable width proc box

vandabbin:
Closes: #1475 

This PR allows the user to adjust the width of the proc box with either keybinds, a menu setting, or in the config file.
Additionally the proc width percentage can be added to the presets with proc taking an additional option in the presets string. Old presets strings continue to work. The absence of a width percentage for a proc preset string results in it using the default value.

<hr>
<details>
<summary><b>Details</b></summary>
<br>
<b>Width Percentage:</b>

- New configuration option where the proc box width percentage is defined.
  - This value is saved to the config file.

- Value can be modified in the menu and is clamped between `0` and `100`.
  - screen redraws when value is changed.

- In `Draw::calcSizes()` the blocks that handle net and mem boxes were modified.
  - Instead of calculating width based on a static percentage it is based on a percentage value retrieved from the config.
  - If the width is larger or smaller then allowed the value is clamped into allowable range without modifying the stored value.

- `Proc::width_p` was made external so that it can be used as the default width setting.

<b>Keybinds:</b>

- `Shift+Left`: makes the proc box wider by 1% if it is on the right and narrower by 1% if it is on the left.

- `Shift+Right`: makes the proc box narrower by 1% if it on the right and wider by 1% if is on the left.

- `Alt+Shift+Left`: makes the proc box wider by 10% if it is on the right and narrower by 10% if it is on the left.

- `Alt+Shift+Right`: makes the proc box narrower by 10% if it on the right and wider by 10% if is on the left.

- `Ctrl+Shift+Left`: sets the proc box width to its maximum if on the right and minimum if on the left.

- `Ctrl+Shift+Right`: sets the proc box width to its minimum if on the right and maximum if on the left.

- `Ctrl+Shift+Down`: sets the proc box width to the original default from before this PR (55%)

<b>Modifications:</b>

-  Made a change in `Draw::calcSizes()` to prevent process list cpu graphs from briefly vanishing while adjusting size.
  - This was done with a new `atomic<bool> resized` inside the `Proc` namespace. When this value or `Global::resized` is true. Then the graphs inside the process list are not cleared.

<b>Presets:</b>
- `presetsValid` was modified to properly validate the presets string when proc has a 3rd option set.
-  `presets` menu option description was updated to reflect the new option
-  `apply_preset` was modified to apply the proc width percentage
-  Changing the width with either the arrow keys or the menu sets the `current_preset` value back to `-1`
-  valid formats `proc:P:G` or `proc:P:G:W`
    - `proc:0:default`, `proc:0:default:[0-100]`, and `proc:0:default:default`
</details>
<hr>

<details>
<summary><b>Benefits</b></summary>
<br>

<b>This PR with the included ability to apply it in presets has benefits.</b>
- The user has more choice in the size of the proc, mem and net boxes.
  - Allows for a view with a large proc box while still having mem and net boxes shown.
  - Allows for a very small proc box with extra large mem and net boxes.
  - Or any size in between.
  - Allows the proc box to have more space in a smaller window for potential future data columns
  - Allows more buttons in the proc box to be visible in a smaller window.
- If used in presets then it is easy to switch between various set widths.
  - This means that presets can be customized further.
  - (ex. a preset with default width proc box on the right and a preset with max width proc box on the left) 
- Shift+Arrow keys give fine granularity to the setting
- Alt+Shift+Arrow keys allow for less precise adjustments (faster)
- Ctrl+Shift+Arrow keys allow for quickly setting the maximum or minimum.
- Menu option allows the user to set a precise value without having to guess.

</details>

<hr>
<details>
<summary><b>Example Videos</b></summary>

https://github.com/user-attachments/assets/6c752093-5367-4074-ad75-1579af4ead36

https://github.com/user-attachments/assets/ffc93b47-f810-430e-85f1-55e523bedbc6

</details>

### DIFF
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -313,7 +313,7 @@ namespace Input {
 							new_proc_box_width_percent = Proc::width_p; // set to default width
 
 						if (proc_box_width_percent != new_proc_box_width_percent) {
-							Config::current_preset = -1;
+							Config::current_preset.reset();
 							Config::set("proc_box_width_percent", new_proc_box_width_percent);
 							Draw::calcSizes(!Proc::shown);
 							Draw::update_clock(true);
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -291,11 +291,14 @@ namespace Menu {
 				"P=(0 or 1) for alternate positions.",
 				"G=graph symbol to use for box.",
 				"",
+				"proc box format can also be \"proc:P:G:W\"",
+				"W=(0 - 100) for box width percentage.\n",
+				"",
 				"Use whitespace \" \" as separator between",
 				"different presets.",
 				"",
 				"Example:",
-				"\"mem:0:tty,proc:1:default cpu:0:braille\""},
+				"\"mem:0:tty,proc:1:default:80 cpu:0:braille\""},
 			{"shown_boxes",
 				"Manually set which boxes to show.",
 				"",
@@ -1388,6 +1391,7 @@ static int optionsMenu(const string& key) {
 					if (option == "proc_box_width_percent") {
 						editor.text = fmt::format("{}", std::clamp(stoi(editor.text), 0, 100));
 						screen_redraw = true;
+						Config::current_preset.reset();
 					}
 					Config::set(option, stoi(editor.text));
 				}
@@ -1477,6 +1481,7 @@ static int optionsMenu(const string& key) {
 					if (option == "proc_box_width_percent") {
 						value = std::clamp(static_cast<int>(value), 0, 100);
 						screen_redraw = true;
+						Config::current_preset.reset();
 					}
 					Config::set(option, static_cast<int>(value));
 				}
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -73,8 +73,9 @@ namespace Config {
 
 		{"presets",				"#* Define presets for the layout of the boxes. Preset 0 is always all boxes shown with default settings. Max 9 presets.\n"
 								"#* Format: \"box_name:P:G,box_name:P:G\" P=(0 or 1) for alternate positions, G=graph symbol to use for box.\n"
+								"#* proc box format can also be \"proc:P:G:W\" W=(0-100) for proc width percentage.\n"
 								"#* Use whitespace \" \" as separator between different presets.\n"
-								"#* Example: \"cpu:0:default,mem:0:tty,proc:1:default cpu:0:braille,proc:0:tty\""},
+								"#* Example: \"cpu:0:default,mem:0:tty,proc:1:default:80 cpu:0:braille,proc:0:tty\""},
 
 		{"vim_keys",			"#* Set to True to enable \"h,j,k,l,g,G\" keys for directional control in lists.\n"
 								"#* Conflicting keys for h:\"help\" and k:\"kill\" is accessible while holding shift."},
@@ -468,7 +469,7 @@ namespace Config {
 					return false;
 				}
 				const auto& vals = ssplit(box, ':');
-				if (vals.size() != 3) {
+				if (vals.size() != 3 and not (vals.size() == 4 and vals.at(0) == "proc")) {
 					validError = "Malformatted preset in config value presets!";
 					return false;
 				}
@@ -482,6 +483,10 @@ namespace Config {
 				}
 				if (not v_contains(valid_graph_symbols_def, vals.at(2))) {
 					validError = "Invalid graph name in config value presets!";
+					return false;
+				}
+				if (vals.size() == 4 and not (vals.at(3) == "default" or intValid("", vals.at(3)))) {
+					validError = "Invalid proc width percent in config value presets!";
 					return false;
 				}
 			}
@@ -515,6 +520,7 @@ namespace Config {
 				set("mem_below_net", (vals.at(1) != "0"));
 			} else if (vals.at(0) == "proc") {
 				set("proc_left", (vals.at(1) != "0"));
+				set("proc_box_width_percent", (vals.size() == 4 and vals.at(3) != "default") ? std::clamp(stoi(vals.at(3)), 0, 100) : Proc::width_p);
 			}
 			if (vals.at(0).starts_with("gpu")) {
 				set("graph_symbol_gpu", vals.at(2));


deckstose:
Not sure how complicated but if we go that far we could also think about variable widths and controlling them with the left and right arrow keys.

But in my head this already opens up the wish to refactor the drawing code into a library to allow that for more boxes...

vandabbin:
> Not sure how complicated but if we go that far we could also think about variable widths and controlling them with the left and right arrow keys.

Yeah I was already thinking about this possibility but proposed this option because it was very minimal code change.

> But in my head this already opens up the wish to refactor the drawing code into a library to allow that for more boxes...

Yeah that would be really great, but is probably outside my level of experience so if this interferes with that we can just not merge this I guess

vandabbin:
> Not sure how complicated but if we go that far we could also think about variable widths and controlling them with the left and right arrow keys.

If you would prefer though I can experiment with modifiying this PR to allow for something like that

deckstose:
> Yeah I was already thinking about this possibility but proposed this option because it was very minimal code change.

The problem with these kind of ad-hoc changes (and also with the amount of new configuration options we introduced in the couple of releases) is it gets really unmaintainable really fast as it doesn't scale well.

If you're interested in this, go for it :rocket:. I'd prefer it over this, but not saying I'm against this.

vandabbin:
> The problem with these kind of ad-hoc changes (and also with the amount of new configuration options we introduced in the couple of releases) is it gets really unmaintainable really fast as it doesn't scale well.

Hmm yeah I can see that. I apologize if the PRs I have been making are troublesome. I don't want to be a burden. I just like this project a lot. I can have trouble looking at the bigger picture sometimes.

> If you're interested in this, go for it 🚀. I'd prefer it over this, but not saying I'm against this.

Cool I'll experiment and see what I come up with.

deckstose:
> Hmm yeah I can see that. I apologize if the PRs I have been making are troublesome. I don't want to be a burden.

No, don't think like that, you're contributions are very valuable and drive the project forward.

I didn't mean to say that a single person doing this is the problem, I have found myself in that spot as well where it's easier to just hack together some band aid solution. But at a scale we need to keep track of everything and need to rewrite the portions where a refactor makes sense.

vandabbin:
@deckstose 

Hey, I got an implementation of variable proc box width using `Shift+Left and Shift+Right` it still needs a bit of work.

However there is a strange issue. If the offset value stored in the config is negative then when the program starts it doesn't use it and gets set to 0. it works fine and restores the previous offset if the config value is positive though? Why would this be?

it writes the proper negative value to the config file but doesn't read it if it is negative very odd

Ohhh i see isint() only works with positive numbers. I will need to fix that.

vandabbin:
alrighty negative integers can be read from the config file now.

would you mind seeing what you think of this?

vandabbin:
Main issue I have found so far that I need to address is that resizing the window ( making it wider also messes with the offset) I can fix this by using a percent instead of an offset i think ( not sure why i thought that would make a difference) I think i prefer percent but it wasn't what was causing the issue

oh actually there are a few things to address. standby lol

src/btop_tools.hpp
```diff
@@ -259,9 +259,9 @@ namespace Tools {
 		return is_in(str, "true", "True");
 	}
 
-	//* Check if a string is a valid integer value (only positive)
+	//* Check if a string is a valid integer value
 	constexpr bool isint(const std::string_view str) {
-		return std::ranges::all_of(str, ::isdigit);
+		return str[0] == '-' ? str.size() > 1 && std::ranges::all_of(str.substr(1), ::isdigit) : std::ranges::all_of(str, ::isdigit);
```
aristocratos: Image

vandabbin: Alright thanks for that info!  My current version which hasnt been pushed here yet actually reverts this back to the original since instead of using a positive or negative offset I am planning to use a more static width (instead of basing it off an offset of the default position and sizing)

vandabbin:
Okay I have something going that uses a percentage value and it seems to be working well.

vandabbin:
Oh there is a minor issue I do still need to solve to fix an issue where if the percentage is set lots higher or lower then what can be acheived it potentially takes many presses to get the value into a range you can see movement.

src/btop_input.cpp
```diff
@@ -275,6 +278,24 @@ namespace Input {
...
+						if (key == "shift_left")
+							proc_box_width_percent = Config::getB("proc_left") ? std::min(max_p - 1, std::max(min_p, --proc_box_width_percent)) : std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent));
+						else if (key =="shift_right")
+							proc_box_width_percent = Config::getB("proc_left") ? std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent)) : std::min(max_p - 1, std::max(min_p, --proc_box_width_percent));
+						else
+							proc_box_width_percent = 55;
+
+						Config::set("proc_box_width_percent", proc_box_width_percent);
+						Draw::calcSizes();
+						Runner::run("all", false, true);
```
vandabbin: I don't know if it is better to run this with `no_update` set to `true` or not. 

I've tried it both ways. The only issue with having no_update set is it gets rid of the cpu graphs in the proc box for one update cycle and if the update cycle is short some of them show and some dont while adjusting width

The benefit of having it set though is the graphs don't update in overtime if you hold `Shift+Left`, `Shift+Right`, or `Shift+Down`

vandabbin:
Just added a menu option for the percentage value so that it can be set that way as well

edit: but it appears I need to figure out how to make changing the value in the menu trigger `Draw::calcSizes()`

I've figured this out.

vandabbin:
oops found an issue for when the proc box isn't shown that needs fixing. my bad. one sec

edit: fixed

src/btop_input.cpp
vandabbin: ### With `no_update` set to true

https://github.com/user-attachments/assets/81aecc96-9e41-4b08-ae8e-81e7d09fd4fe

### With ` no_update` set to false

https://github.com/user-attachments/assets/fa8f7ec9-612c-4f90-b7b4-52a04e9e7442

I think I am leaning towards setting it to true now

Maybe I can figure out a fix for the process list cpu graphs

src/btop_input.cpp
```diff
@@ -275,6 +278,26 @@ namespace Input {
...
+				}
+				else if (is_in(key, "shift_left", "shift_right", "shift_down")) {
+					if (Proc::shown and (Mem::shown or Net::shown)) {
+						int proc_box_width_percent = Config::getI("proc_box_width_percent");
+						const int min_p = round(((double)Proc::min_width / (double)Term::width) * 100);
+						const int max_p = round(100 - (((double)(Mem::shown ? Mem::min_width : Net::min_width) / (double)Term::width) * 100));
+						if (key == "shift_left")
+							proc_box_width_percent = Config::getB("proc_left") ? std::min(max_p - 1, std::max(min_p, --proc_box_width_percent)) : std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent));
+						else if (key =="shift_right")
+							proc_box_width_percent = Config::getB("proc_left") ? std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent)) : std::min(max_p - 1, std::max(min_p, --proc_box_width_percent));
```
vandabbin: hmm I don't get this warning on macos where I was writing this. But now that I am testing the compile on linux I do get it. I need to figure out what is bad about this

```
src/btop_input.cpp: In function ‘void Input::process(std::string_view)’:
src/btop_input.cpp:288:80: warning: operation on ‘proc_box_width_percent’ may be undefined [-Wsequence-point]
  288 |                                                         proc_box_width_percent = Config::getB("proc_left") ? std::min(max_p - 1, std::max(min_p, --proc_box_width_percent)) : std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent));
      |                                                         ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/btop_input.cpp:290:80: warning: operation on ‘proc_box_width_percent’ may be undefined [-Wsequence-point]
  290 |                                                         proc_box_width_percent = Config::getB("proc_left") ? std::max(min_p + 1, std::min(max_p, ++proc_box_width_percent)) : std::min(max_p - 1, std::max(min_p, --proc_box_width_percent));
      |                                                         ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

vandabbin: I figured it out. Created a new variable `new_proc_box_width_percent` to deal with it

src/btop_shared.hpp
```diff
@@ -349,7 +349,7 @@ namespace Proc {
 	extern atomic<int> numpids;
 
 	extern string box;
-	extern int x, y, width, height, min_width, min_height;
+	extern int x, y, width, height, min_width, min_height, width_p;
```
deckstose: Declaring multiple values on the same line is not good practise. It's fine to keep existing code this way, but this shouldn't be used for new code.

src/btop_menu.cpp
```diff
@@ -211,6 +211,11 @@ namespace Menu {
 		{"Selected k", "Kill selected process with SIGKILL - 9."},
 		{"Selected s", "Select or enter signal to send to process."},
 		{"Selected N", "Select new nice value for selected process."},
+		{"Shift + Left", "Adjust the proc box width when mem or net shown."},
+		{"Shift + Right", "Adjust the proc box width when mem or net shown."},
+		{"Ctrl+Shift+Left", "Set the proc box to max/min width."},
+		{"Ctrl+Shift+Right", "Set the proc box to max/min width."},
+		{"Ctrl+Shift+Down", fmt::format("Reset the proc box width percentage to {}% (default).", to_string(Proc::width_p))},
```
deckstose:
```suggestion
		{"Ctrl+Shift+Down", fmt::format("Reset the proc box width percentage to {}% (default).", Proc::width_p)},
```

src/btop_menu.cpp
```diff
@@ -211,6 +211,11 @@ namespace Menu {
 		{"Selected k", "Kill selected process with SIGKILL - 9."},
 		{"Selected s", "Select or enter signal to send to process."},
 		{"Selected N", "Select new nice value for selected process."},
+		{"Shift + Left", "Adjust the proc box width when mem or net shown."},
+		{"Shift + Right", "Adjust the proc box width when mem or net shown."},
+		{"Ctrl+Shift+Left", "Set the proc box to max/min width."},
+		{"Ctrl+Shift+Right", "Set the proc box to max/min width."},
```
deckstose: I think these comments should not be generic for all keys.

So 

- Left -> Increase proc box size
- Right -> Decrease proc box size

Wording doesn't matter, but should be specific.

src/btop_draw.cpp
```diff
@@ -2440,7 +2441,8 @@ namespace Draw {
 		//* Calculate and draw net box outlines
 		if (Net::shown) {
 			using namespace Net;
-			width = round((double)Term::width * (Proc::shown ? width_p : 100) / 100);
+			const int new_width = round((double)Term::width * (100 - Config::getI("proc_box_width_percent")) / 100);
```
deckstose: C-Style cast -> static_cast.

src/btop_draw.cpp
```diff
@@ -2376,7 +2376,8 @@ namespace Draw {
 			auto swap_disk = Config::getB("swap_disk");
 			auto mem_graphs = Config::getB("mem_graphs");
 
-			width = round((double)Term::width * (Proc::shown ? width_p : 100) / 100);
+			const int new_width = round((double)Term::width * (100 - Config::getI("proc_box_width_percent")) / 100);
+			width = Proc::shown ? clamp(new_width, min_width, Term::width - Proc::min_width) : (int)Term::width;
```
deckstose: I'd prefer `std::{clamp,round}` here, but this isn't a blocker (This was done it below in some cases).

src/btop_menu.cpp
```diff
@@ -1356,6 +1372,10 @@ static int optionsMenu(const string& key) {
 					}
 				}
 				else if (selPred.test(isInt) and Config::intValid(option, editor.text)) {
+					if (option == "proc_box_width_percent") {
+						editor.text = to_string(std::clamp(stoi(editor.text), 0, 100));
```
deckstose: fmt::format

src/btop_shared.hpp
vandabbin: Ok got it I will move it to its own line.

src/btop_menu.cpp
vandabbin: The issue was that depending on if the proc box is on the left or right they left and right arrows do the opposite and the length was to long to fit in the menu box. 

I will try to think of better wording for these

deckstose: Hmm I see, I guess this can't be dynamic without a lot of effort? Then it's fine

src/btop_menu.cpp
vandabbin: Ohhh right I wasn't thinking about fmt::format here. Ill do that

src/btop_menu.cpp
vandabbin: > Hmm I see, I guess this can't be dynamic without a lot of effort? Then it's fine

Oh I could try and see about making it dynamic! I didn't think of that for some reason.

Which is funny because I had made the set to default one dynamic (kind_of) if width_p was ever changed.

oh although I guess maybe that only works because its a preset variable. It doesn't work with Config....I should have realized that.

deckstose: Don't worry about it, the generic comment makes perfect sense.

src/btop_menu.cpp
```diff
@@ -1440,8 +1460,13 @@ static int optionsMenu(const string& key) {
 				if (key == "right" or (vim_keys and key == "l")) value += mod;
 				else value -= mod;
 
-				if (Config::intValid(option, to_string(value)))
+				if (Config::intValid(option, to_string(value))) {
```
vandabbin, line 1480: @deckstose I only touched this line to add a `{` but do you want me to replace the to_string here with fmt::format as well?

deckstose: Doesn't matter to me

src/btop_draw.cpp
vandabbin: `Mem::width_p` and `Net::width_p` are no longer used after the changes made in the PR. I don't know if they should be left or removed. I left them just in case though. Their use was replaced with being whatever remains based on the proc boxes width percent setting

vandabbin:
An alternative help message I can't decide if this is better or not

Screenshot 2026-01-16 at 10 17 36 AM

I could also potentially do:

left    ->   Move the separator between proc and mem/net left.
right  ->   Move the separator between proc and mem/net right.

vandabbin:
I introduced an issue by adding a keybind to move by 10% that I need to fix one sec. Might just remove that ability all together.

I didn't remove it. I made a fix.

src/btop_input.cpp
```diff
@@ -275,6 +282,33 @@ namespace Input {
...
+							new_proc_box_width_percent = Config::getB("proc_left") ? max_p : min_p;
+						else
+							new_proc_box_width_percent = Proc::width_p; // Default width
+
+						if (proc_box_width_percent != new_proc_box_width_percent) {
+							atomic_wait(Runner::active);
+							Config::set("proc_box_width_percent", new_proc_box_width_percent);
+							Draw::calcSizes();
+							Runner::run("all", true, true);
+						}
```
vandabbin, line 321: I'm not sure if this `atomic_wait(Runner::active)` is needed here or not, but I added it because I saw other input code using it at some point before `Draw::calcSizes()` or `Runner::run()`. I can't notice any difference in the way it runs with it included or not though.

src/btop_draw.cpp
```diff
@@ -2316,6 +2316,7 @@ namespace Draw {
 					b_width - (Config::getB("show_cpu_freq") and hasCpuHz ? (freq_range ? 24 : 14) : 5)
 			);
 			box += createBox(b_x, b_y, b_width, b_height, "", false, cpu_title);
+			update_clock(true);
```
vandabbin: I added `update_clock(true)` to the bottom of the Cpu section of `Draw::calcSizes()` as it prevents the clock from disappearing momentarily while adjusting the box width or switching presets. I can remove this if there is a reason it shouldn't be there. 

edit: I realized I could just call `Draw::update_clock(true);` after `Draw::calcSizes();` in the input block and it would achieve the same thing though it means that when switching presets with `p` or cpu box with `1` that the clock momentarily vanishes. I can fix that in a seperate pr though

src/btop_input.cpp
vandabbin: I realized that `atomic_wait(Runner::active)` is the first thing run inside `Draw::calcSizes()` so I don't think it is needed. I removed it

src/btop_draw.cpp
```diff
@@ -2489,5 +2490,9 @@ namespace Draw {
 			select_max = height - 3;
 			box = createBox(x, y, width, height, Theme::c("proc_box"), true, "proc", "", 4);
 		}
+		else {
+			Proc::p_counters.clear();
+			Proc::p_graphs.clear();
+		}
```
vandabbin: In order to prevent the process graphs from disappearing while resizing the width I moved this to an else that executes only if the proc box is not shown. I didn't see any reason it had to be at the top of calcSizes() and couldn't be run only if Proc wasn't still shown. This also fixes that issue for resizing the terminal window

edit: oh. I discovered why it was where it was... if you turn off cpu graphs in the proc list they still remain there haha. I modified this to account for the config value of `proc_cpu_graphs` 

more edit: oh I see. This strategy isn't going to work because it also is messed up by a change in the graph symbols used...

last edit: this has been solved by using `Global::resized` and a new variable `Proc::resized` as conditions to clearing the proc graphs

vandabbin:
I believe this to be ready for final review now. I have fixed it up a couple issues and it is working great. 

Oh one more edit: I made `Proc::resized` atomic because it wasn't working on linux unless I did so. It was working fine on my mac without being atomic though.

vandabbin:
I guess one question I have if this pr is added is. Should the percent value also be settable in the presets?

If that was to be done the preset string would have to be thought about though. Since it might be weird for just one of the boxes to have 4 elements to the preset string. And it would need to be modified in a way that old preset strings continue to work.

I have a draft that uses a 4th element for proc in preset string and doesn't break existing presets. If the proc preset doesn't have a 4th element it uses the default width value.

valid formats:

`proc:P:G` or `proc:P:G:W`

`proc:0:default`, `proc:0:default:[0-100]`, and `proc:0:default:default`

vandabbin:
I have a fix that I will be submitting in another PR that prevents the rapid updating of the graphs when holding the `p` preset key, or holding any of `1`, `2`, `3`, `4`. Currently every press of those buttons triggers an update and so if you hold them the graphs move very fast. faster then they should based on the `update_ms` setting. I have not submitted it yet however because it depends on the adjustment I made in this PR that prevents the process list cpu graphs from disappearing briefly while adjusting the width. (the same thing happens when showing/hiding boxes or cycling preset when `no_update` is set to false for `Runner:run()`

vandabbin:
If the Preset disable options pr is merged I need to change a few lines in this PR to account for the refactor of `current_preset` as a `std::optional`

PR #1485, opened 2026-01-17
CMake: Organize in modules and use FetchContent for deps, add reflection library

deckstose:
This patch tries to reorganize the CMake file structure by providing a single file per directory and prepare for inclusing of another dependency (reflect-cpp).

- Single units if work for CMake have been exported to own files.
- Split widecharwidth and fmt in the include directory: This enables CMake to use widecharwidth independently of the bundled fmt, and fmt will be used via find_package or FetchContent. (It should be possible to use widecharwidth via FetchContent in the future. However it's not a CMake project so it's a bit weird.)
- Add options to explicitly not fetch dependencies via FetchContent: This is mostly insteresting for packagers with strict guidelines of downloading content during the build phase.
- Add CMake workflow file for easy development.

At the moment this still lacks:

- [x] Support for platforms other than Linux
- [x] The cmake FindPackage modules for BSD
- [ ] Static SMI support via FetchContent
- [x] Compiler feature tests (own module)
- [x] Installation of assets (themes and man page)
- [x] config.h configuration
- [ ] Document changes
- [x] CI integration

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -36,6 +36,42 @@ else()
     FetchContent_MakeAvailable(fmt)
 endif()
 
+if(BTOP_USE_SYSTEM_TOMLPLUSPLUS)
+    find_package(tomlplusplus REQUIRED)
+else()
+    FetchContent_Declare(
+        tomlplusplus
+        GIT_REPOSITORY https://github.com/marzer/tomlplusplus.git
+        # gersemi: off
+        # Includes a fix for running clang-tidy.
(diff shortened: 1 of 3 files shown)


deckstose:
@aristocratos 

In theory, you can checkout the branch and run `cmake --workflow --preset default` and compile btop with `reclect-cpp`...

Do you like this? Any thoughts about this? It's just an idea...

The interesting commit is the second one.

Ah and I've gone for the toml integration, however, we could roll custom de-/serialization.

deckstose:
I never want to write a single line of CMake again 😫 😫 let us switch to meson

PR #1492, opened 2026-01-18
Create orange.theme

neocerambyx:
Add monochromatic orange theme


vandabbin:
You should add these attributes as well. They are for newly added features. So the old themes haven't had them added yet

Process list banner attributes:
```
proc_pause_bg: background color of the banner when the list is paused.
proc_follow_bg: background color of the banner when the process following feature is active.
proc_banner_bg: background color of the banner when the process following feature is active AND the list is paused.
proc_banner_fg: foreground (text) color of the banner
```

Process following attributes:
```
followed_bg: background color of the followed process in the list.
followed_fg: foreground color of the followed process in the list.
```

Issue #1132, opened 2025-05-02
[REQUEST] Be able to remove default preset completely


Crandel:
**Is your feature request related to a problem? Please describe.**
I don't like default preset.

**Describe the solution you'd like**
I want to either configure default preset as I wish
**Describe alternatives you've considered**
Or remove it completely

**Additional context**
I need only CPU and PROC blocks

webdevred:
This feels like a nice enhancement, I wiil look into making a PR for this as soon as possible.

imwints:
What is the issue currently? You can remove the net and mem boxes and it will persist.

Crandel:
> What is the issue currently? You can remove the net and mem boxes and it will persist.

I have two presets. One is cpu+proc and another is pure proc. And in order to switch between them I need to press "p" 3 times and not 2. It is not critical, but rather confusing.

JakovFreund:
Same issue, I have 2 custom presets and want to toggle between them without the default getting in the way

cudonem:
Same issue here. Would like to have btop as a scratchpad in my WM. But the default preset is too big in size for that to be allowed, as it contains all the boxes. Would love to either configure the default preset to contain a single box, or get rid of it all together (as OP suggests).

vandabbin:
I have made a PR that addresses this issue.

PR #1493, opened 2026-01-19
Feat: Preset disabling options

vandabbin:
Closes: #1132

This PR creates a menu option called `disable_presets` that allows the user to disable certain presets.

When the option is set to `Off` none of the presets are disabled. (Default Setting)

When the option is set to `Default` only the default preset is disabled.

When the option is set to `Custom` only the custom presets are disabled.

When the option is set to `All` then all of the presets are disabled.

Launching btop with the `-p [0-9]` flag will still choose the desired preset (ignoring this option)

Additionally this PR refactors `current_preset` using `std::optional` 
- Instead of no preset being `-1` it is now std::nullopt
  - preset can be reset with `Config::current_preset.reset()` instead of `Config::current_preset = -1`

As I see it this PR has several benefits.

- If a user wants to set a specific layout but doesn't want to accidentally mess it up by pressing the `p` key, they can disable all of the presets and pressing `p` will do nothing. This is equivalent to the behaviour of the preset cycling before this PR if the `presets` string was set to `""` and is how you would disable presets if that was your method before.

- If a user wants to set their own default preset then disabling only the default layout makes the first custom preset the defacto default preset. This way they can mess around with the layout (changing box positions or show/hide boxes) and press `p` once (instead of twice) to return to their custom default preset. If they have more presets defined then follow up presses will cycle through them as usual. If it is the only custom preset then further presses of `p` will do nothing and it becomes a custom default preset restore key

- If a user wants to be able to mess around with the layout but then decide that they want to return to the default preset and they also don't want to use the custom presets then disabling only the Custom presets will make a single press of the `p` key change the view to the default layout restoring it from whatever configuration they had put it in and further presses of the `p` key will do nothing. This essentially turns the `p` and `P` keys into a default preset restore key without also having to remove any custom presets that are already defined.

- And of course if the user doesn't care about any of this then having this option set to `Off` keeps the original behavior with the one difference being that if no custom presets are defined (preset string = `""`) then `p` acts as a restore default preset key instead of doing nothing like it did before. This is the default setting.

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -65,6 +65,12 @@ namespace Config {
 		{"force_tty", 			"#* Set to true to force tty mode regardless if a real tty has been detected or not.\n"
 								"#* Will force 16-color mode and TTY theme, set all graph symbols to \"tty\" and swap out other non tty friendly symbols."},
 
+		{"disable_presets",		"#* Option to disable presets. Either the default preset, custom presets, or all presets.\n"
+								"#* \"Off\" All presets are enabled.\n"
+								"#* \"Default\" preset is disabled."
+								"#* \"Custom\" presets are disabled."
+								"#* \"All\" presets are disabled."},
+
 		{"presets",				"#* Define presets for the layout of the boxes. Preset 0 is always all boxes shown with default settings. Max 9 presets.\n"
 								"#* Format: \"box_name:P:G,box_name:P:G\" P=(0 or 1) for alternate positions, G=graph symbol to use for box.\n"
 								"#* Use whitespace \" \" as separator between different presets.\n"
@@ -247,6 +253,7 @@ namespace Config {
 		{"color_theme", "Default"},
 		{"shown_boxes", "cpu mem net proc"},
 		{"graph_symbol", "braille"},
+		{"disable_presets", "Off"},
 		{"presets", "cpu:1:default,proc:0:default cpu:0:default,mem:0:default,net:0:default cpu:0:block,net:0:tty"},
 		{"graph_symbol_cpu", "default"},
 		{"graph_symbol_gpu", "default"},
@@ -442,7 +449,7 @@ namespace Config {
 
 	vector<string> current_boxes;
 	vector<string> preset_list = {"cpu:0:default,mem:0:default,net:0:default,proc:0:default"};
-	int current_preset = -1;
+	std::optional<int> current_preset;
 
 	bool presetsValid(const string& presets) {
 		vector<string> new_presets = {preset_list.at(0)};
--- a/src/btop_input.cpp
+++ b/src/btop_input.cpp
@@ -253,22 +253,27 @@ namespace Input {
 						Menu::show(Menu::Menus::SizeError);
 						return;
 					}
-					Config::current_preset = -1;
+					Config::current_preset.reset();
 					Draw::calcSizes();
 					Draw::update_clock(true);
 					Runner::run("all", false, true);
 					return;
 				}
-				else if (is_in(key, "p", "P") and Config::preset_list.size() > 1) {
+				else if (is_in(key, "p", "P") and Config::getS("disable_presets") != "All") {
+					if (Config::getS("disable_presets") == "Default" and Config::preset_list.size() <= 1) return;
 					const auto old_preset = Config::current_preset;
-					if (key == "p") {
-						if (++Config::current_preset >= (int)Config::preset_list.size()) Config::current_preset = 0;
+					const int first_preset = (Config::getS("disable_presets") == "Default") ? 1 : 0;
+					if (Config::getS("disable_presets") == "Custom") Config::current_preset = 0;
+					else if (Config::current_preset.has_value()) {
+						if (key == "p") {
+							if(++(*Config::current_preset) >= static_cast<int>(Config::preset_list.size())) Config::current_preset = first_preset;
+						}
+						else if (--(*Config::current_preset) < first_preset) Config::current_preset = Config::preset_list.size() - 1;
 					}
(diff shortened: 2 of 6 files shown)


src/btop_input.cpp
```diff
@@ -259,14 +259,19 @@ namespace Input {
...
 					const auto old_preset = Config::current_preset;
+					const int first_preset = (Config::getS("disable_presets") == "Default") ? 1 : 0;
 					if (key == "p") {
-						if (++Config::current_preset >= (int)Config::preset_list.size()) Config::current_preset = 0;
+						if (++Config::current_preset >= static_cast<int>(Config::preset_list.size()) or Config::current_preset < first_preset)
+							Config::current_preset = first_preset;
 					}
 					else {
-						if (--Config::current_preset < 0) Config::current_preset = Config::preset_list.size() - 1;
+						if (--Config::current_preset < first_preset) Config::current_preset = Config::preset_list.size() - 1;
```
deckstose: Collapse else/if? 🕵🏽

src/btop_input.cpp
```diff
@@ -259,14 +259,19 @@ namespace Input {
...
 					return;
 				}
-				else if (is_in(key, "p", "P") and Config::preset_list.size() > 1) {
+				else if (is_in(key, "p", "P") and Config::getS("disable_presets") != "All") {
+					if (Config::getS("disable_presets") == "Default" and Config::preset_list.size() <= 1) return;
 					const auto old_preset = Config::current_preset;
+					const int first_preset = (Config::getS("disable_presets") == "Default") ? 1 : 0;
 					if (key == "p") {
-						if (++Config::current_preset >= (int)Config::preset_list.size()) Config::current_preset = 0;
+						if (++Config::current_preset >= static_cast<int>(Config::preset_list.size()) or Config::current_preset < first_preset)
```
deckstose: Can you explain how incrementing the current preset be less then 1?

Otherwise we can throw an `assert` in there and forget about the `or` block.

src/btop_config.cpp
```diff
@@ -65,6 +65,12 @@ namespace Config {
 		{"force_tty", 			"#* Set to true to force tty mode regardless if a real tty has been detected or not.\n"
 								"#* Will force 16-color mode and TTY theme, set all graph symbols to \"tty\" and swap out other non tty friendly symbols."},
 
+		{"disable_presets",		"#* Option to disable presets. Either just the default preset or all presets.\n"
+								"#* \"Off\" All presets are enabled.\n"
+								"#* \"Default\" Only the default preset is disabled."
+								"#* \"Custom\" Only the custom presets are disabled."
+								"#* \"All\" All of the presets are disabled."},
```
deckstose: Nit: I think this can be trimmed down a bit (same for the menu). Feel free to ignore if you think this looses tone, quality or information.

```suggestion
		{"disable_presets",		"#* Option to disable presets. Either default, custom or or all presets.\n"
								"#* \"Off\" All presets are enabled.\n"
								"#* \"Default\" Default preset is disabled."
								"#* \"Custom\" Custom presets are disabled."
								"#* \"All\" All presets are disabled."},
```

src/btop_menu.cpp
```diff
@@ -1497,6 +1508,8 @@ static int optionsMenu(const string& key) {
 				}
 				else if (is_in(option, "proc_sorting", "cpu_sensor", "show_gpu_info") or option.starts_with("graph_symbol") or option.starts_with("cpu_graph_"))
 					screen_redraw = true;
+				else if (option == "disable_presets" and optList.at(i) != "Off")
+					Config::current_preset = -1;
```
deckstose: Could this be refactored to use an `optional` to encode the off state in type system?

deckstose:
Oh I've forgotten, is there a way to use the old option to "remove" the default preset? I admit I've not used a single preset in my life

src/btop_input.cpp
vandabbin: good point. I missed that.

src/btop_input.cpp
vandabbin: Yes the current_preset variable is set to -1 when it is not on a preset. When that is incremented it is equal to 0 which is below the first_preset if the default preset is disabled

src/btop_config.cpp
vandabbin: No I like your suggestion. I will make this change

vandabbin:
> Oh I've forgotten, is there a way to use the old option to "remove" the default preset? 

I'm not entirely sure what you mean here by old option. However the way to remove the default preset before was to just set the presets string to be empty and that disabled the preset cycling. This disabled all presets though not just the default and unless you made a backup of your config, that also meant throwing away any presets you did have set. The default preset is hard coded as the first entry in the `preset_list` so it isn't shown as part of the `presets` string in the menu or the config. 

So to disable the presets the same way as before this PR you would set the Disable presets option to `All`

When I started working on this initially my approach was a method that relocated the default preset to the preset string. but it felt messy and this is more flexible. (this is why the branch is named `relocate-default-preset` 😆 )

> I admit I've not used a single preset in my life

I don't really use them much either, but once in a while I do

src/btop_menu.cpp
vandabbin: I will find out. tbh I've not used `optional` before so I will have to learn about it first. But if I am understanding this then instead of using -1 as an indicator of not being on a preset (`*`), it would use the `optional` `std::nullopt`.  Is this corrrect?

I'll see what I come up with for this

vandabbin: @deckstose 

I have refactored `current_preset` as a `std::optional` is this what you meant?

deckstose:
That's what I meant

PR #1496, opened 2026-01-20
fix: suppress warning of redefined C++ keyword

deckstose:
Suppresses the diagnostic for clang but adds an explainatory comment why it's needed.


deckstose:
I don't get that warning with GCC 15 and [CI](https://github.com/aristocratos/btop/actions/runs/21168811860/job/60880155968) doesn't have it either. But I'll take a look.

Issue #1497, opened 2026-01-20
collection updates when preset cycling or show/hide box


vandabbin:
**Describe the bug**

When someone presses and hold the `p`, `P`, `1`, `2`, `3`, or `4` keys it triggers an update of the screen
This causes the graphs to draw very quickly while holding those buttons (or pressing them really fast in succession)

Additionally since everytime you push one of those buttons the graph moves, if you have a load going in periodic intervals it may appear to run longer in the graph but it was actually just that you pressed those buttons

**Expected behavior**

When pressing and holding those buttons the presets should cycle or the box should appear and disappear without causing the graphs to draw in overtime.

<details>
<summary><b>Example Video</b></summary>

https://github.com/user-attachments/assets/1833a447-c483-439c-a949-94726bcafe6a

</details>

PR #1498, opened 2026-01-20
fix: set no_update and resized to true when show/hide box or preset cycle

vandabbin:
Fixes: #1497 

This prevents the graphs from updating overly fast if you hold down the preset cycle keybind or the `1`, `2`, `3`, or `4` keys.

Also if you were to just press them very quickly.

Instead the graphs update at the same rate as they are supposed to based on `update_ms`

The reason for `if(Proc::shown) Proc::resized = true;` is to prevent the clearing out of the process list cpu graphs when changing the preset or showing and hiding a box after setting `no_update` to true.

`no_update` being false allowed the process list graphs to stay visible but the update cycle happens sooner then it should based on the `update_ms` value meaning that the timescale of the graphs isn't consistent.

The use of `Global::resized` in `calcSizes()` means that the proc list graphs dont vanish briefly on a terminal window resize

<details>
<summary><b>Example Video</b></summary>

https://github.com/user-attachments/assets/c2b75f58-b905-43d2-b9ad-7e6a3545105a

</details>

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1562,6 +1562,7 @@ namespace Proc {
 	int user_size, thread_size, prog_size, cmd_size, tree_size;
 	int dgraph_x, dgraph_width, d_width, d_x, d_y;
 	bool previous_proc_banner_state = false;
+	atomic<bool> resized (false);
 
 	string box;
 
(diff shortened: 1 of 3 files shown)


deckstose:
Good catch! 🚀

deckstose:
For the `extern` declaration: I'd like to start reducing global variables as much as possible. It makes it hard to reason about the program's state and even harder to debug.

Do you have any ideas for concrete design on how to reduce that? I'm thinking something myself, but it's always good to have another opinion.

deckstose (approved):

vandabbin:
> For the `extern` declaration: I'd like to start reducing global variables as much as possible. It makes it hard to reason about the program's state and even harder to debug.
> 
> Do you have any ideas for concrete design on how to reduce that? I'm thinking something myself, but it's always good to have another opinion.

At the moment I don't have any specific ideas. I will start thinking about it however!

Issue #1500, opened 2026-01-21
Seemingly unnecessary metics updates when menu options toggled


vandabbin:
Similar to PR #1498 

I don't see any reason for the metrics to be updated when menu options are toggled. I have tested them and it seems as if for the ones that affect the screen. screen_redraw is enough and everything does what it should. In fact the only difference I can notice between before this PR and after is that the graphs dont move faster then they should.

If there is a reason it is required and I am missing it. I would love to know what it is.

PR #1501, opened 2026-01-21
fix: metric updates with preset cycle, show/hide box, and menu toggle

vandabbin:
Fixes: #1500

Similar to PR #1498 but also will fix the issue caused by #1498 that I somehow missed. 

I screwed up. I thought my fix for the fast metric update with preset cycle and show/hide box worked. but I forgot to take into account launching btop with the box hidden. I could have sworn I tested this but now testing again I must not have.

Working on fix now however if I can't figure one out I guess this will turn into a revert PR my apologies

toggling options in the menu causes metrics to update. if you toggle options quickly it updates the graphs overly fast. 

There is only one option in the menu that requires a collection update and that appears to be the one for the process list cpu graph symbol. I missed this before so my earlier change needs to be modified.

This rapid update issue also exists for expanding and collapsing the tree in the process list however it only updates the process list and detailed view graph overly fast so its not as huge an issue. Also I haven't figured out a way to solve this one yet as the code that handles the tree collapse and expand is inside the collect code so collect needs to be called for it to happen.

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -1303,7 +1303,6 @@ static int optionsMenu(const string& key) {
 			Theme::updateThemes();
 		}
 		int retval = Changed;
-		bool recollect{};
 		bool screen_redraw{};
 		bool theme_refresh{};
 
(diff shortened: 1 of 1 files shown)


src/btop_menu.cpp
```diff
@@ -1629,11 +1625,6 @@ static int optionsMenu(const string& key) {
 			Draw::calcSizes();
 			Global::overlay = std::move(overlay_bkp);
 			Global::clock = std::move(clock_bkp);
-			recollect = true;
-		}
-		if (recollect) {
-			Runner::run("all", false, true);
```
vandabbin: I tested just setting `no_update` to true here as well but it behaves the same way with that as if I remove all the lines I did. If just setting `no_update` to true is better though I will do that instead of removing all these lines.

PR #1503, opened 2026-01-21
improved fix for metric updates during view change

vandabbin:
Fixes: #1498 #1500

This PR is a better version of the previous fix that had to be reverted for the fast metric updates when show/hiding box and cycling presets.

The previous fix had an issue where if the box had not been viewed since opening btop then it would not populate all the data and the box could take longer then it should to open.

This new fix prevents that by keeping track of the boxes that have been seen since btop started. 

It also removes the global `Proc::resized` variable that was used in `Draw::calcSizes` and replaces it with a boolean argument for `Draw::calcSizes` that default to `true` if it is not included. I think this is cleaner.

<b>What this fixes:</b>
- Rapidly opening and closing a box or cycling the presets will not cause the metrics to update
- If a box has not previously been opened since btop was started then a metric update is triggered for that box when it is shown
  - <b>This prevents the issue caused by the old fix.</b>
  - It does this by running `Runner::run()` for each newly seen box with `no_update` set to false and `redraw` set to false
- If boxes are shown or hidden and the proc box is already visible. The proc cpu graphs do not briefly vanish.
  - likewise when the Terminal window is resized this does the same thing.
- Not clearing proc graphs if the proc box is still shown breaks the graphs when cycling presets if the graph symbol changes. This now clears the proc graphs if cycling presets and the proc graph symbol changes to prevent that.

<b>What this does not fix:</b>
- Toggling options rapidly in the menu continues to update metrics too fast
  - I am still working out the best solution for this.
- If the proc box was previously shown but is not shown now and then is toggled on the proc graphs do not immediately appear.
  - This happens because if the proc box is hidden then the proc graphs clear like before but when it is shown again they need time to update. 
  - I don't have a good solution for this yet.
  - I need to look more into how those graphs work to prevent it.

<details><summary><b>Example Video (Shows fix for newly opened boxes)</b></summary>

https://github.com/user-attachments/assets/18c5dc89-441a-4fbf-a183-a749bab57521

</details>

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -1067,6 +1067,7 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
 		Config::set_boxes("cpu mem net proc");
 		Config::set("shown_boxes", "cpu mem net proc"s);
 	}
+	Config::set_seen_boxes(Config::getS("shown_boxes"));
 
 	//? Update list of available themes and generate the selected theme
 	Theme::updateThemes();
@@ -1163,7 +1164,7 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
 
 			//? Trigger secondary thread to redraw if terminal has been resized
 			if (Global::resized) {
-				Draw::calcSizes();
+				Draw::calcSizes(false);
 				Draw::update_clock(true);
 				Global::resized = false;
 				if (Menu::active) Menu::process();
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -440,7 +440,9 @@ namespace Config {
 
 	vector<string> available_batteries = {"Auto"};
 
+	bool did_proc_graph_symbol_change = false;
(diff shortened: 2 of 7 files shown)


vandabbin:
Another small issue I didn't notice with this is that if you have presets with different graph symbols set on the proc box they don't change properly on a preset cycle. I'm thinking about how to fix this and marking this as a draft until I do.

vandabbin:
I will mark this as ready after I figure out a solution for the proc graphs and the menu

Currently the proc graphs vanishing are the only problem I have to really figure out for the menu. The modifications I currently have for it work for the most part, but do this by doing collection update for any options that change the proc graphs symbols. This isn't great in my opinion so I'd really like to figure out a way to get the proc graphs symbols to change properly without a collection update like all the other  graphs do. If I can't manage that I think just getting the process CPU percentages to not shift to the left when the graphs briefly vanish could be good enough.

Issue #956, opened 2024-11-01
[REQUEST] GPU support for AMD GPUs on arm64 platforms with no ROCm support


geerlingguy:
**Is your feature request related to a problem? Please describe.**

I love `btop`, it's great. But I can't get GPU metrics to display on my Raspberry Pi running an RX 6700 XT (see here: https://pipci.jeffgeerling.com/cards_gpu/amd-radeon-rx6700xt-12gb.html). ROCm [so far doesn't support arm64 at all](https://github.com/ROCm/ROCm/issues/3960), but someday that will hopefully change.

**Describe the solution you'd like**

Btop should be able to display GPU metrics on AMD GPUs in Linux without requiring ROCm support.

I should note the default drivers seem to work fine with `radeontop` and `nvtop`—see for example `nvtop` example output:

<img width="1090" alt="Screenshot 2024-11-01 at 2 57 56 PM" src="https://github.com/user-attachments/assets/fb7dc863-f640-4a61-b3b8-9358e0a18f10">

**Describe alternatives you've considered**

For now I just use `nvtop` with `btop running in a separate window.

**Additional context**

Yeah I know I'm weird, but it'd be nice to get this working so I could also use `btop` to monitor GPU utilization on Ampere Altra / AmpereOne systems.

PR #1512, opened 2026-01-26
Feat 1413 AMD gtt memory usage added with backup for non ROCm

CowboyTim:
This adds GTT memory usage for AMD GPUs that support it. It uses ROCm, as that implementation was already there, with a fallback to using sysfs to also implement #956 possibly (I cant test this).

I've tested both with ROCm, and without. I've also tested on a regular Intel i7-7567U that has no AMD/NVIDIA, but integrated graphics.


TheSovietPancakes:
I feel like these features should be separate (sysfs, gtt), since they feel like solutions to majorly different problems, but otherwise your implementation of GTT with ROCm looks perfect. I will have to test the sysfs later on my RX 9070 XT, but the GTT matches a prototype I made a while ago. 👍

CowboyTim:
Note that, in the mean time I found out that on an AMD Strix HALO with UMA arch, you can also "disable" the GTT memory usage and everything becomes dynamic.  Then the memory usage from llama.cpp is added with the regular process memory, and it's the total memory that just changes Frree/Cached/Available. Total is also the maximum at that point. 

Initially I assumed this was because the MES was disabled, but it's not only in that case aparently, there are more flags/features. It's also annoying to test, as this the whole setup is unstable. It's basically a giant puzzle until things don't crash that often anymore.

Issue #1514, opened 2026-01-26
Force TTY option in config file doesn't apply when btop is started


vandabbin:
**Describe the bug**

When btop is launched the force_tty option in the config file isn't used so if it is set to true. then tty mode is not set like it should. Additionally this means that if it was set to true and then you exit btop (saving the config) and then go back into the options menu, you will see that force_tty is true but you are not in tty mode. toggling that option will then turn tty mode on when force_tty is set to false which is backwards.

**To Reproduce**

1. Launch btop and open option menu. 
2. Set force_tty to `true` (This should enable tty mode)
3. Quit out of btop ensuring that save to config option is enabled.
4. Launch btop again and notice that you are not in tty mode.
5. Go to options menu and see that force_tty mode is set to `true`
6. Toggle force_tty and see tty mode enabled when force_tty is set to false.

**Expected behavior**

If force_tty mode is set to `true` in the config file. Launching btop should use that to know and enable tty mode.

If this behavior happens then simultaneously the toggle option in the menu for force_tty will accurately reflect whether it is forcing tty mode or not.

**Info (please complete the following information):**
 - btop++ version: v1.4.6
 - Binary: self compiled
 - Architecture: arm64
 - Platform: macOS v15.7.3
 - Terminal used: Ghostty

deckstose:
We should remove the option altogether. No one has complained about it not working, yet, and it's really just an ad-hoc options to quickly force tty mode. If you really want it just use the TTY theme and be done with it.

vandabbin:
> We should remove the option altogether. No one has complained about it not working, yet, and it's really just an ad-hoc options to quickly force tty mode. If you really want it just use the TTY theme and be done with it.

Yeah ok I think you are probably right. I will modify my PR to remove the option

vandabbin:
@deckstose 

I have just now made a modification (not pushed yet) that leaves the option available in the menu, but removes it from the config file. This also fixes the issue of it being reversed if you quit btop with it set to true, Do you think that this is sufficient or should I remove it entirely?

rereading your comment I realize you are probably for just complete removal. I will go ahead with that

aristocratos:
> We should remove the option altogether. No one has complained about it not working, yet, and it's really just an ad-hoc options to quickly force tty mode. If you really want it just use the TTY theme and be done with it.

[...]

PR #1515, opened 2026-01-26
fix: set tty mode if force_tty is true in config file

vandabbin:
Fixes: #1514 

The `force_tty` menu option wasn't being applied when btop launched
even though it was being saved in the config file.

This fixes this by applying tty mode if this option is set to true in the config file

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -1455,9 +1455,13 @@ static int optionsMenu(const string& key) {
 					theme_refresh = true;
 					Config::flip("lowcolor");
 				}
+			#if !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__NetBSD__)
+				else if (option == "force_tty" and not Term::current_tty.starts_with("/dev/tty")) {
+			#else
 				else if (option == "force_tty") {
+			#endif
 					theme_refresh = true;
-					Config::flip("tty_mode");
+					Config::set("tty_mode", Config::getB("force_tty"));
 				}
 				else if (is_in(option, "rounded_corners", "theme_background"))
 					theme_refresh = true;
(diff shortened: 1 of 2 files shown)


vandabbin:
> This doesn't account for `--no-tty`+ config file `force_tty = true`

Oh great catch!! I forgot that there was a no-tty cli option as well!

Although this does make it trickier. The no-tty command line option would leave force_tty as true in the config file so the menu option would continue to be backwards if btop is launched with `--no-tty`. I will have to think about this one some more to come up with something good

<hr>

I have compensated for this by changing the `force_tty` menu option to only toggle tty mode on if it is `true` and only toggle it off if it is `false`. I also added some platform specific code that basically disables the effect of toggling the `force_tty` menu option if a TTY was auto-detected (the same way it seems to be done when btop starts). I'm not 100% on the formatting I used for the compiler #if. I may have it indented poorly

- This way the user can launch with --no-tty and if `force_tty` is true in the config then the cli option over rules it (which I think makes sense). 
- The menu option continues to show `force_tty` as true which is technically accurate (as a reflection of the config file) even if not currently in tty mode because it was done by a temporary measure (i.e. cli flag).
- If the menu option is then toggled off tty mode is not enabled as it would have been previously when all that did was `flip` the option.
- If the menu option `force_tty` is then toggled to true again it will reset to tty mode (overriding the cli option).
- If a TTY has been auto-detected toggling `force_tty` on or off will not have any visible effect.

I could possibly figure out some changes to make the `force_tty` menu option reflect the cli flag as well without saving it to the config file (if set by flag) when it exits, but I am not sure it is worth the extra work

src/btop_menu.cpp
```diff
@@ -1455,9 +1455,13 @@ static int optionsMenu(const string& key) {
...
 				}
+			#if !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__NetBSD__)
+				else if (option == "force_tty" and not Term::current_tty.starts_with("/dev/tty")) {
+			#else
 				else if (option == "force_tty") {
+			#endif
 					theme_refresh = true;
-					Config::flip("tty_mode");
+					Config::set("tty_mode", Config::getB("force_tty"));
 				}
```
vandabbin, line 1465: I was not sure about how I should do the indentation here. I could use input on that

PR #1524, opened 2026-02-01
Add rule to editorconfig for parity with CONTRIBUTING.md

TheSovietPancakes:
In [CONTRIBUTING.md](https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md), there is a line that reads:
```md
- Opening curly braces { at the end of the same line as the statement/condition.
```
Along with the other two rules which are defined (4-space-wide tabs, not spaces), the .editorconfig should now match the requirements.

In the future, I think more rules should be added to the editorconfig file, since if I were to format the code it would likely look very different to if somebody else formatted the code. This leads to every source file being a big mix of different styles and since it is not proper etiquette to make a PR for just formatting, the current mess will stick.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,7 @@
 [*.{cpp,h,hpp,sh,md,cfg,sample}]
 indent_style = tab
 indent_size = 4
+brace_style = same_line
 
 [*.{yml,yaml}]
 indent_style = space


deckstose:
Is there any difference or maybe an advantage over using clang-format? I have an ongoing effort at https://github.com/aristocratos/btop/pull/1146 for that. It would include rules on brace placement and much more.

TheSovietPancakes:
Probably not? .Clang-format is only for clang-format, while .editorconfig should be adhered to by all formatters, but I only made this PR because there is already a .editorconfig file in the repo.

deckstose:
I was just wondering, if the other PR is settled we can get rid of this as we then have only one formatter to worry about

deckstose (approved):

PR #1539, opened 2026-02-13
fix: accumulate pwr_usage into pwr_total instead of mem_total

alephpiece:
`gpu.pwr_usage` should be accumulated into `pwr_total` instead of `mem_total`.


aristocratos (changes requested):
Would you mind also on line 1968 changing `pwr_total / gpu_pwr_total_max` to `pwr_total * 100 / gpu_pwr_total_max` since this should be a percentage value.

alephpiece:
I've updated both `gpu-pwr-total` and `gpu-vram-total`. I matched the existing style in the codebase by using `double` and `round()` to avoid integer truncation. Is this what you had in mind, or should we stick to the simpler `* 100` approach?

src/linux/btop_collect.cpp
```diff
@@ -1963,9 +1963,9 @@ namespace Gpu {
 
 		shared_gpu_percent.at("gpu-average").push_back(avg / gpus.size());
 		if (mem_total != 0)
-			shared_gpu_percent.at("gpu-vram-total").push_back(mem_usage_total / mem_total);
+			shared_gpu_percent.at("gpu-vram-total").push_back((long long)round((double)mem_usage_total * 100.0 / (double)mem_total));
```
deckstose: Please use `static_cast` instead of C-style casts

deckstose: Also if one of the operands is a floating point value you don't need to explicitly cast the other one to a floating point number, or the result.

aristocratos (approved):

PR #1546, opened 2026-02-18
proc: add cwd to process detail view

amaanq:
The process detail panel now shows the working directory above the command line. On Linux this reads the `/proc/[pid]/cwd` symlink, on macOS it uses `proc_pidinfo` with `PROC_PIDVNODEPATHINFO`, and on FreeBSD/NetBSD it uses `sysctl` with `KERN_PROC_CWD`. OpenBSD doesn't expose other processes' cwd so it shows "(unavailable)" there.

The CMD display goes from 3 lines to 2 to make room, which means longer commands truncate a line sooner.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1848,16 +1848,24 @@ namespace Proc {
 				if (item_fit >= 8) out += cjust("Nice:", item_width);
 
 
-				//? Command line
-				for (int i = 0; const auto& l : {'C', 'M', 'D'})
-				out += Mv::to(d_y + 5 + i++, d_x + 1) + l;
+				//? CWD and Command line
+				out += Mv::to(d_y + 5, d_x + 1) + Theme::c("title") + Fx::b + "CWD" + Fx::ub;
+				out += Mv::to(d_y + 6, d_x + 1) + Theme::c("title") + Fx::b + "CMD" + Fx::ub;
 
-				out += Theme::c("main_fg") + Fx::ub;
+				out += Theme::c("main_fg");
+				//? CWD line
+				out += Mv::to(d_y + 5, d_x + 5);
+				if (detailed.cwd.empty())
+					out += Theme::c("inactive_fg") + "(unavailable)";
+				else
+					out += uresize(detailed.cwd, d_width - 6, true);
+
+				//? Command lines (2 lines)
 				const auto san_cmd = replace_ascii_control(detailed.entry.cmd);
 				const int cmd_size = ulen(san_cmd, true);
-				for (int num_lines = min(3, (int)ceil((double)cmd_size / (d_width - 5))), i = 0; i < num_lines; i++) {
-					out += Mv::to(d_y + 5 + (num_lines == 1 ? 1 : i), d_x + 3)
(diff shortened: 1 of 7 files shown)


src/linux/btop_collect.cpp
```diff
@@ -2909,6 +2909,14 @@ namespace Proc {
...
 			d_read.close();
 		}
+
+		//? Get current working directory from proc/[pid]/cwd symlink
+		detailed.cwd.clear();
+		try {
+			if (fs::is_symlink(pid_path / "cwd"))
+				detailed.cwd = fs::read_symlink(pid_path / "cwd").string();
+		}
+		catch (...) {}
```
deckstose, line 2919: Please add a comment why an exception can be safely ignored.

src/btop_draw.cpp
```diff
@@ -1848,16 +1848,24 @@ namespace Proc {
 				if (item_fit >= 8) out += cjust("Nice:", item_width);
 
 
-				//? Command line
-				for (int i = 0; const auto& l : {'C', 'M', 'D'})
-				out += Mv::to(d_y + 5 + i++, d_x + 1) + l;
+				//? CWD and Command line
+				out += Mv::to(d_y + 5, d_x + 1) + Theme::c("title") + Fx::b + "CWD" + Fx::ub;
```
deckstose, line 1852: Please use `fmt::format_to` when editing or writing new draw code.

See `CONTRIBUTING.md` and https://github.com/aristocratos/btop/issues/535 for more information.

src/btop_draw.cpp
```diff
@@ -1848,16 +1848,24 @@ namespace Proc {
 				if (item_fit >= 8) out += cjust("Nice:", item_width);
 
 
-				//? Command line
-				for (int i = 0; const auto& l : {'C', 'M', 'D'})
-				out += Mv::to(d_y + 5 + i++, d_x + 1) + l;
+				//? CWD and Command line
```
deckstose, line 1851: NIT: Although it's done like this everywhere else the question mark in comments doesn't add any value IMO and can be left out.

src/btop_shared.hpp
```diff
@@ -422,7 +422,7 @@ namespace Proc {
 		size_t last_pid{};
 		bool skip_smaps{};
 		proc_info entry;
-		string elapsed, parent, status, io_read, io_write, memory;
+		string elapsed, parent, status, io_read, io_write, memory, cwd;
```
deckstose, line 425:
```suggestion
		string elapsed, parent, status, io_read, io_write, memory;
		std::string cwd;
```

```suggestion
		string elapsed, parent, status, io_read, io_write, memory;
		string cwd;
```

Please do not add new variables to multi-line declarations.

deckstose (commented):
Have a few comments on the actual implementation.

Overall I don't like the design. I have several comments on issues in the back of my head where people wanted to see _more_ of the command line in the detailed view, not less, which often doesn't even fit on three lines.

Also the keywords `CWD` and `CMD` are hard to distinguish right beneath each other.

I think we can come up with a better solution than this.

PR #1553, opened 2026-02-24
test: add cpu name trim tests

deckstose:


tests/cpu_names.cpp
```diff
@@ -0,0 +1,34 @@
...
+	EXPECT_EQ(Cpu::trim_name("13th Gen Intel(R) Core(TM) i5-13500"), "13th Gen i5-13500");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Core(TM) i5-14600"), "i5-14600");
+
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz"), "E5-2690");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz"), "E5-2690");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) Silver 4410Y"), "Xeon Silver 4410Y");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz"), "@");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz"), "@");
+	EXPECT_EQ(Cpu::trim_name("INTEL(R) XEON(R) GOLD 6548Y+"), "INTEL XEON GOLD 6548Y+");
+	EXPECT_EQ(Cpu::trim_name("Intel(R) Xeon(R) Platinum 8368Q CPU @ 2.60GHz"), "@");
```
aristocratos, line 34: Why does some of the tests resolved to single letters?

Shouldn't the tests resolve to what we want to see like "i7 Q 840" instead of "Q".

Otherwise the tests would fail if someone actually fixes the CPU name trim function.

aristocratos (approved):

tests/cpu_names.cpp
deckstose: > Shouldn't the tests resolve to what we want to see like "i7 Q 840" instead of "Q".

Yes, this is just the test function. I have forgotten to work on this further, but the first step when fixing `Cpu::trim_name` for these CPUs you would change the single letter to what you expect and work until the test  passes again.

I was just documenting the status quo of what names (that I know exist) resolve to what return value.

deckstose: I'd like to merge this as-is and take a stab at it when I got some more time.

PR #1557, opened 2026-02-25
fix: wrap std::thread to join in destructor in btop_collect.cpp

### LINKED ISSUE #1050: [BUG] SIGABRT thrown on wakeup from suspend
**Read the README.md and search for similar issues before posting a bug report!**

Any bug that can be solved by just reading the [prerequisites](https://github.com/aristocratos/btop#prerequisites) section of the README will likely be ignored.

**Describe the bug**

When waking up from systemctl suspend, a SIGABRT crash screen for btop++ can be seen.

**To Reproduce**

1. Run btop
2. Suspend arch.
3. Resume arch.

**Expected behavior**

btop++ should be able to resume itself securely.

- btop version: 1.4.0+5535f50
 - Compiled with: g++ (14.2.1)
 - Configured with: make STATIC= FORTIFY_SOURCE=true GPU_SUPPORT=true RSMI_STATIC=false
 - btop-git 1.4.0.r1116.5535f50-1 from AUR.
 - Architecture: x86_64
 - Platform: Arch Linux
 - (Linux) Kernel: 6.13.3-zen1-1-zen
 - Terminal used: kitty
 - Font used: JetBrainsMono Nerd Font

**Additional context**

[btop.log](https://github.com/user-attachments/files/18947205/btop.log)

omrash:
These changes prevent the IOT instruction core dump from NVML after waking up from suspend, specifically `nvmlDeviceGetPcieThroughput()`. This resolves #1050.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -110,6 +110,33 @@ long long get_monotonicTimeUSec()
 	return time.tv_sec * 1000000 + time.tv_nsec / 1000;
 }
 
+// Wrapper with join behavior of std::jthread to avoid -fexperimental-library
+// flag usage for Clang 19, ideally replace once flag is no longer needed.
+struct join_thread
+{
+	std::thread t;
+
+	join_thread() noexcept = default;
+
+	template<typename... Args>
+	explicit join_thread(Args&&... args) : t(std::forward<Args>(args)...) {}
+
+	~join_thread() { if (t.joinable()) t.join(); }
+
+	join_thread(join_thread&&) noexcept = default;
+
+	join_thread& operator=(join_thread&& other) noexcept {
+		if (t.joinable()) t.join();
+		t = std::move(other.t);
+		return *this;
+	}
+
+	void join() {
+		t.join();
+	}
+
+};
+
 }
 
 namespace Cpu {
@@ -1293,7 +1320,8 @@ namespace Gpu {
 			if (!initialized) return false;
 
 			nvmlReturn_t result;
-			std::thread pcie_tx_thread, pcie_rx_thread;
+			join_thread pcie_tx_thread;
+			join_thread pcie_rx_thread;
 			// DebugTimer nvTotalTimer("Nvidia Total");
 			for (unsigned int i = 0; i < device_count; ++i) {
 				if constexpr(is_init) {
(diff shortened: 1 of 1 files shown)


omrash:
Apparently clang is lagging behind in their implementation of this C++20 language feature (you may need to enable the `fexperimental-library` flag). Would you prefer I stick with `std::thread` and wrap the thread function block in a try-catch instead?

deckstose:
> Would you prefer I stick with std::thread and wrap the thread function block in a try-catch instead?

Yes.

omrash:
> > Would you prefer I stick with std::thread and wrap the thread function block in a try-catch instead?
> 
> Yes.

Just tested it and a try-catch doesn't work since the thread still terminates and doesn't join. You'd probably need to have a class to wrap std::thread to force it to join in it's destructor, which is basically reimplementing std::jthread. If that's the only way to fix it I'll do it but I think it makes more sense to use an existing implementation in the standard library.

deckstose:
Isn't there another way to get cooperative cancellation of the threads?

omrash:
> Isn't there another way to get cooperative cancellation of the threads?

The only alternatives I can think of are less than ideal, like detaching the threads and using some kind of condition variable to signal the main thread. This is the cleanest solution I could come up with but I'm open to any ideas!

deckstose:
Okay, it seems they have put this behind an experimental flag in LLVM 18 to reserve the possibility to change implementation details. However, they have not done this. So the implementation should be the same regardless of the experimental flag.

I'm fine with using `std::jthread`, but it would be great to limit whatever `-fexperimental-library` enables to `jthread` only.

This is relevant for LLVM 19 only, am I right?

src/linux/btop_collect.cpp
```diff
@@ -1293,7 +1293,7 @@ namespace Gpu {
 			if (!initialized) return false;
 
 			nvmlReturn_t result;
-			std::thread pcie_tx_thread, pcie_rx_thread;
+			std::jthread pcie_tx_thread, pcie_rx_thread;
```
deckstose: Please split that to one declaration per line.

deckstose (commented):
You should add a comment in the Makefile that the flag is only relevant for LLVM 19. I don't expect a full version check there, this would be madness, but this way it's easier to mark this line obsolete once we drop LLVM 19 support.

omrash:
> Okay, it seems they have put this behind an experimental flag in LLVM 18 to reserve the possibility to change implementation details. However, they have not done this. So the implementation should be the same regardless of the experimental flag.
> 
> I'm fine with using `std::jthread`, but it would be great to limit whatever `-fexperimental-library` enables to `jthread` only.
> 
> This is relevant for LLVM 19 only, am I right?

Yes, Clang 20 and up seem to build fine without it. `std::jthread` is implemented in `<thread>` so we're not pulling in any new includes.

omrash:
@deckstose This has been sitting for a while, are there any other changes that need to be made?

omrash:
@aristocratos @deckstose Could this fix please be merged in if nothing is blocking it?

deckstose:
> I'm fine with using std::jthread, but it would be great to limit whatever -fexperimental-library enables to jthread only.

omrash:
> > I'm fine with using std::jthread, but it would be great to limit whatever -fexperimental-library enables to jthread only.

There's no real practical way to achieve this, but I can understand the apprehension in enabling the flag wholesale. I've opted to just wrap the threads in a struct that mimics std::jthread's join behavior instead, and can confirm that it still fixes the issue from local testing.

deckstose (approved):
Thank you, and apologies for the long delay 💚

Issue #1528, opened 2026-02-06
htop features


rokke-git:
**Is your feature request related to a problem? Please describe.**

your interface is much nicer than htop's, and I like how the parent processes inherit all of the usage of their children, but there are four features from htop sorely missing from btop.

the easiest one is a recursive close for the tree view.  
the second is per-process cpu time.  
the third is showing threads as separate items.
the fourth is showing recently dead processes.

**Describe the solution you'd like**

the first one's easy to describe: it would be great if we could close(or open) all of the items at once in tree mode. since the hotkey for tree mode is currently `e`, maybe `E` could open/close everything. the biggest strength of tree mode on btop is that when everything is closed it's a great per-application summary of resource usage, so it's kinda annoying that you have to individually double-click every single plus sign.

the second is a request for a new data metric. on htop, one of the best metrics is the time statistics. for each process, it has both the time since the process was initially started, and also the amount of cpu time the process has actually used. this allows you to easily gauge how impactful each program is actually being over its full life cycle, or how much more intensive one program really is vs another. we currently have the elapsed time available when you open the more details menu, but I would really appreciate the ability to put both elapsed and used time directly in the list view as sortable values

the third is being able to show threads separately. while the last was a metric to see the overall impact of an application, this is more about seeing the specifics. usually, the cpu time of each thread is very different, even when the elapsed time is the same for all of them.

finally, recently dead processes. it would be nice if there was an option to have them stick around as greyed-out in the list for a number of frames. some programs call a whole pile of little scripts that instantly finish, and I would like to be able to make sure that all of them are actually getting run. there is the option of slowing btop down to an update per minute, but then it feels very frozen and it doesn't always act correctly when you proceed to click on those dead processes for more information

**Describe alternatives you've considered**

using htop?

**Additional context**

cpu time in htop:

<img width="176" height="244" alt="Image" src="https://github.com/user-attachments/assets/a82b8bfc-a799-4c6f-ae57-1d023204a15d" />

deckstose:
Please split that into different requests and look for duplicates.

rokke-git:
I did look at all the other issues; you even spotted me flagging a duplicate. I figured one person creating four issues would be loud and annoying, but I can split them if you would like that

deckstose:
No, it's just appropriate.

I just mentioned that cause I'm pretty sure I've seen the recursive tree collapse if I understand the request correctly.

Issue #686, opened 2023-12-14
[REQUEST] Configuration to autocollapse processes by name


halvabner:
### Examples
Processes like `Chromium` (yes, I know I'm using the wrong browser :P). It starts ~40 subprocesses at a time. This causes:

###### Memory used by instances is displayed spread over them but I would like to just know "how much memory the browser is using"

Other examples are `kthreads` and `Discord`, I would like many common processes collapsed by default.

### Visual difference
![231214-1307-51](https://github.com/aristocratos/btop/assets/130675714/c34c7854-0321-45e5-99a2-77811ecc8e2a)
![231214-1308-46](https://github.com/aristocratos/btop/assets/130675714/54f19d5e-e817-4eeb-9edf-3b6791516f95)
### VS _(note how nice memory usage is to look at here)_
![231214-1308-58](https://github.com/aristocratos/btop/assets/130675714/10eced85-98f6-419c-abbe-75eaee36132c)

### Implementation
I don't know what would be easiest or most appropriate but I think a configuration list of names to match for auto-collapsing would be nice.

### Related to [this](https://github.com/aristocratos/btop/issues/520)

Cheers ^_^

imwints:
> Other examples are kthreads and Discord, I would like many common processes collapsed by default.

For `kthreads`, you can disable all kernel processes in the settings, hit `esc`, `enter`, `5` and then scroll down to `Proc filter kernel` and set it to true.

dm17:
Definitely useful to see chromium listed once in btop with RAM & CPU usage totaled! Very useful feature!

johnhamelink:
#673 is also related to this. It would be very nice when trying to have the ability to collapse processes in some way (for me, I'd like to be able to set it by default in the config).

crimsonduelist:
If u have 30 sub processes of chromium having 233 megabytes each it s not as usefull as knowing that chromium takes up ~7 gigs. If u have more programs like that it quickly becomes messy and I have to track down each one to fold them. It would be very usefull to have this. Especially as and option to have it in the conf.

dm17:
> If u have 30 sub processes of chromium having 233 megabytes each in the tree view it s not as usefull as knowing that chromium takes up ~7 gigs. If u have more programs like that it quickly becomes messy and I have to track down each one to fold them. It would be very usefull to have this. Especially as and option to have it in the conf.

Exactly!

postrational:
[...]

Issue #673, opened 2023-11-30
support for `collapse-all` in tree view


Avivbens:
**Is your feature request related to a problem? Please describe.**
When using the `btop` package, it can become cumbersome to navigate through the tree view when there are multiple processes. Currently, there is no option to collapse all nodes in the tree view, making it challenging to get an overview of the processes.

**Describe the solution you'd like**
I would like to propose the addition of a "collapse-all" feature in the tree view of `btop`. This feature would allow users to collapse all nodes in the tree view with a single action, providing a more organized and condensed view of the processes.

**Describe alternatives you've considered**
A configuration option to automatically collapse nodes by default, but having a dedicated "collapse-all" action provides more flexibility to users.

**Additional context**
No additional context at the moment. However, this feature would enhance the user experience by simplifying the navigation and management of processes in the tree view.

NOTE:
Awesome tool! very appreciated 👍

luisnquin:
I wish this issue would get more attention.

Issue #520, opened 2023-04-28
[REQUEST] Add option to autocollapse processes in tree view to a certain level


OmegaRogue:
**Is your feature request related to a problem? Please describe.**
In tree view, i always have to manually collapse all processes if i want to actually see whats going on

**Describe the solution you'd like**
A settings option to set a depth to auto-collapse the process tree to on startup

Otard95:
I'de love something like this. Like `foldlevel` in vim. Could also be nice with key binds like in vim, `zm` folds one more level and `zr` reduces the fold level.

materemias:
I'd be happy with a simple collapse all command as a start as well :)

rtviii:
A collpase button in vim mode would be extremely nice. Kudos on all your work!

luisnquin:
I wish this issue would get more attention.

PR #1564, opened 2026-03-04
Add E hotkey to collapse/expand all processes in tree mode

### LINKED ISSUE #1563: [BUG] Sorting in tree mode with collapsed nodes causes a brief visual glitch
**Describe the bug**

When in tree mode with any collapsed nodes, pressing left/right to change the sort column causes the process list to flicker for one frame before snapping to the correct view.

Image 

Image

**To Reproduce**

1. Enable tree mode (e)
2. Collapse a few nodes with spacebar
3. Press left or right to change the sort column

**Expected behavior**

Tree re-renders cleanly in the new sort order.

**Info:**
- btop++ version: 1.4.6
- Architecture: arm64
- Platform: macOS
- OS release: 26.2
- Terminal: Ghostty
- Font: JetBrainsMono Nerd Font

**Additional context**

Only happens with a few collapsed node. With everything expanded the sort change renders fine.

Writing this bug report because I am working on a PR for adding a new shortcut to auto-collapse everything

sardonyx001:
Closes #1563, addresses #520, #673, #1528 and #686.

Been using btop for years, it's my favorite system monitor by far. 
Tree mode is great but I always found myself tediously spacebar-collapsing every single node just to get the per-app summary view. Finally sat down gave it a try.

Here's the overview:

- `E` collapses all processes in tree mode at once, or expands them all if everything is already collapsed
- Root processes (launchd, systemd, top-level daemons) are left untouched since collapsing them would just hide everything underneath rather than giving you the summary view
- `E` pairs naturally with the existing `e` (toggle tree mode on/off)
- Also fixes the one-frame rendering glitch from #1563: sort key presses were calling `Runner::run` with `no_update=true`, which meant `_tree_gen` skipped setting `p.filtered=true` for collapsed children while [line 211 of `btop_shared.cpp`](https://github.com/aristocratos/btop/blob/1ec8b7379ac0660ce5615ce2d10313f957fe0da9/src/btop_shared.cpp#L211) was unconditionally clearing their stale flag from the previous cycle. Fixed by forcing `no_update=false` on left/right when `proc_tree` is active.

This is my first time contributing to a project so large so please tell me if there's anything I need to update/change.

Before and After:

https://github.com/user-attachments/assets/88068790-fd1f-48ec-adef-446ec4b0f198

### DIFF
--- a/src/freebsd/btop_collect.cpp
+++ b/src/freebsd/btop_collect.cpp
@@ -1303,15 +1303,23 @@ namespace Proc {
 			}
 
 			if (collapse_all != -1) {
-				//? If any parent process is currently expanded, collapse all; otherwise expand all
-				std::unordered_set<size_t> parent_pids;
-				for (const auto& p : current_procs)
+				//? Build sets of all pids and parent pids to identify root processes
+				std::unordered_set<size_t> pid_set, parent_pids;
+				for (const auto& p : current_procs) {
+					pid_set.insert(p.pid);
 					parent_pids.insert(static_cast<size_t>(p.ppid));
-				const bool do_collapse = rng::any_of(current_procs, [&parent_pids](const proc_info& p) {
-					return parent_pids.contains(p.pid) and not p.collapsed;
+				}
+				//? If any non-root parent is expanded, collapse; otherwise expand
+				const bool do_collapse = rng::any_of(current_procs, [&parent_pids, &pid_set](const proc_info& p) {
+					return parent_pids.contains(p.pid)
+						and pid_set.contains(static_cast<size_t>(p.ppid))
+						and not p.collapsed;
 				});
-				for (auto& p : current_procs)
+				//? Root processes (parent not in tracked list) are never touched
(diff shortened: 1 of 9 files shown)


src/freebsd/btop_collect.cpp
```diff
@@ -1301,6 +1301,29 @@ namespace Proc {
...
+						and not p.collapsed;
+				});
+				//? Root processes (parent not in tracked list) are never touched
+				for (auto& p : current_procs) {
+					if (not pid_set.contains(static_cast<size_t>(p.ppid))) continue;
+					p.collapsed = do_collapse;
+				}
+				collapse_all = -1;
+				if (Config::ints.at("proc_selected") > 0) locate_selection = true;
+			}
```
aristocratos, line 1309: This should probably be refactored to a shared function in btop_shared.cpp so we don't duplicate the code for each platform.

aristocratos (approved):

aristocratos:
Thanks for contributing!

PR #1573, opened 2026-03-13
feat: show/hide ip address option

### LINKED ISSUE #1273: Hide ip in the "net" box
It would be amazing if the ip address could be hidden from the "net" box.

Some sort of variable that can be turned to false in order hide the ip would be a nice feature

yuheho7749:
This implements a menu option to show or hide ip addresses. This is useful on systems with public ip addresses.

Closes #1273

Built and tested on Arch Linux 6.19.6

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -1467,6 +1467,7 @@ namespace Net {
 		if (force_redraw) redraw = true;
 		auto net_sync = Config::getB("net_sync");
 		auto net_auto = Config::getB("net_auto");
+		auto net_hide_ip = Config::getB("net_hide_ip");
 		auto tty_mode = Config::getB("tty_mode");
 		auto swap_upload_download = Config::getB("swap_upload_download");
 		auto& graph_symbol = (tty_mode ? "tty" : Config::getS("graph_symbol_net"));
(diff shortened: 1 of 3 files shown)


src/btop_draw.cpp
```diff
@@ -1467,6 +1467,7 @@ namespace Net {
 		if (force_redraw) redraw = true;
 		auto net_sync = Config::getB("net_sync");
 		auto net_auto = Config::getB("net_auto");
+		auto show_ip = Config::getB("show_ip");
```
deckstose: This can be `const`.

src/btop_menu.cpp
```diff
@@ -799,6 +799,12 @@ namespace Menu {
 			    "       setting for bitrates",
 			    "",
 			    "True, False, or Auto",},
+			{"show_ip",
+				"IP address visibility.",
+				"",
+				"Manually set the IP address visibility.",
```
deckstose: This is redundant.

src/btop_menu.cpp
```diff
@@ -799,6 +799,12 @@ namespace Menu {
 			    "       setting for bitrates",
 			    "",
 			    "True, False, or Auto",},
+			{"show_ip",
+				"IP address visibility.",
```
deckstose: This can be more direct on what happens if this is set to true. The false case is self explanatory.

deckstose (commented):
We should rename this to align with the other `net_` options. I also want to flip the option (since it's rather odd or special to _hide_ the IP address). I suggest `net_hide_ip` or `net_hide_ip_addresses` and default it to false obviously.

src/btop_draw.cpp
yuheho7749: I was following the other Config::getB. I can change this if you want.

PR #1580, opened 2026-03-20
Display Total Power consumption on Apple Silicon Devices

georgebastille:
I just got an M5 Macbook and noticed the power consumption element was not working. I found https://github.com/beutton/wattsec.git that works great, and used it as a base to modify btop. Built and tested on this M5 Macbook.


aristocratos:
Read https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md

PR #1581, opened 2026-03-20
Fix: Use correct PID for signals to processes

### LINKED ISSUE #846: [BUG] signal gets send to wrong process
**Describe the bug**

When sending any signal (like terminate or kill) to a process in btop, if its position in the list of programs changes before the confirmation popup is closed, the wrong process receives the signal because its now in the position from the original selected process. btop fails to retain the selection when the confirmation prompt appears, resulting in the signal being sent to the currently selected process upon confirmation, rather than the intended process. (see screenshots)

**To Reproduce**

1. Open btop and select a process you wish to send the signal to
2. Send a signal but refrain from confirming the action yet.
3. Ensure that the selected process's position in the list changes (this could sometimes be induced just by opening/closing other processes, or simply waiting for the order to shift naturally because of memory usage etc..).
4. Upon confirmation, the signal will be sent to the process currently occupying the selection's original position, instead of the intended process.

**Expected behavior**

The signal should be send to the process that was selected before the confirmation prompt appears. Btop should dynamically track the selected process, adjusting as it moves within the list, thus preventing accidental termination of the wrong process.

**Screenshots**

issue

(Screenshot shows that i intended to send a kill signal to pcmanfm, but it got send to firefox instead because the process moved to its position. Happened many times before)

**Info**
btop version: 1.3.2
os: Arch

yinghao-w:
Fixes #846. Process signal popup menus use the "current" selected process, which can drift around when not following processes. Fix by retaining and using the initial PID selected at popup creation. Applies to `t`erminate, `K`ill, `s`ignals and `N`ice.

### DIFF
--- a/src/btop_menu.cpp
+++ b/src/btop_menu.cpp
@@ -981,10 +981,14 @@ namespace Menu {
 	};
 
 	static int signalChoose(const string& key) {
-		auto s_pid = (Config::getB("show_detailed") and Config::getI("selected_pid") == 0 ? Config::getI("detailed_pid") : Config::getI("selected_pid"));
+		static std::optional<int> s_pid;
 		static int x{};
 		static int y{};
 		static int selected_signal = -1;
+
+		if (!s_pid) {
+			s_pid = (Config::getB("show_detailed") and Config::getI("selected_pid") == 0 ? Config::getI("detailed_pid") : Config::getI("selected_pid"));
+		}
 
 		if (bg.empty()) selected_signal = -1;
 		auto& out = Global::overlay;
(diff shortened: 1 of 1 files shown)


src/btop_menu.cpp
```diff
@@ -981,11 +981,15 @@ namespace Menu {
 	};
 
 	static int signalChoose(const string& key) {
-		auto s_pid = (Config::getB("show_detailed") and Config::getI("selected_pid") == 0 ? Config::getI("detailed_pid") : Config::getI("selected_pid"));
+		static int s_pid{};
```
deckstose: I'd rather have this one be an `std::optional` to clearly indicate the absence of a pid.

deckstose (changes requested):

src/btop_menu.cpp
yinghao-w: Good idea. Will do.

deckstose (approved):

PR #1583, opened 2026-03-23
fix: correctly map amd zen cpu core temps

### LINKED ISSUE #1582: Incorrect AMD Zen CPU core temps
_Note: Contains AI generated content, all content has been reviewed and vetted by me._

**Describe the bug**

On AMD multi-CCD CPUs (e.g. Ryzen 5950X), btop++ incorrectly maps per-CCD temperature sensors to CPU cores. `Tccd1` is shown for even-numbered cores (0, 2, 4, 6, ...) and `Tccd2` for odd-numbered cores (1, 3, 5, 7, ...), when the correct mapping is `Tccd1` → cores 0–7 (and their SMT siblings 16–23) and `Tccd2` → cores 8–15 (and their SMT siblings 24–31).

The root cause is in `get_core_mapping()`: the code uses the physical core ID directly as a sensor index when it is smaller than the number of sensors, then cycles `n++` through sensors for the rest — producing an alternating pattern instead of grouping cores by CCD.

**To Reproduce**

1. Run btop++ on a system with an AMD multi-CCD CPU (e.g. Ryzen 5900X, 5950X, Threadripper)
2. Enable core temperature display
3. Observe that the temperature displayed per-core alternates between two values rather than showing the same temperature for all cores within a physical CCD

**Expected behavior**

Cores belonging to the same CCD should all display the temperature from that CCD's sensor. For example, on a Ryzen 5950X:
- Cores 0–7 and 16–23 → `Tccd1`
- Cores 8–15 and 24–31 → `Tccd2`

**Screenshots**

**Current (note that the core with 100% utilization is not reporting the correct temp)**:

Image

**Expected (will provide PR)**:

Image

**Info**
 - btop++ version: 
    ```
    btop version: 1.4.6
    Compiled with: g++ (15.2.1)
    Configured with: make STATIC= GPU_SUPPORT=true RSMI_STATIC=
    ```
 - Architecture: x86_64
 - Platform: Arch Linux
 - (Linux) Kernel: `6.19.8-arch1-1`
 - `sensors` output:
    ```
    k10temp-pci-00c3
    Adapter: PCI adapter
    Tctl:         +68.2°C
    Tccd1:        +50.8°C
    Tccd2:        +45.0°C
    ```
- `lscpu` output:
  ```
  Architecture:                x86_64
    CPU op-mode(s):            32-bit, 64-bit
    Address sizes:             48 bits physical, 48 bits virtual
    Byte Order:                Little Endian
  CPU(s):                      32
    On-line CPU(s) list:       0-31
  Vendor ID:                   AuthenticAMD
    Model name:                AMD Ryzen 9 5950X 16-Core Processor
      CPU family:              25
      Model:                   33
      Thread(s) per core:      2
      Core(s) per socket:      16
      Socket(s):               1
      Stepping:                0
      Frequency boost:         enabled
      CPU(s) scaling MHz:      63%
      CPU max MHz:             5086.1812
      CPU min MHz:             582.1540
      BogoMIPS:                6799.59
      Flags:                   <...>
  ```

unlimitedsola:
Fixes: #1582, more detail in the linked issue.

Note: original code generated by AI, all code has been reviewed, vetted, and tested by me.

The approach is to divide core ids evenly by sensor count, [taken from `htop`](https://github.com/htop-dev/htop/blob/88e00aa53c1cbf727069ace4123072d4fa9ae7c3/linux/LinuxMachine.c#L675-L713).

Before (incorrect temp on the core with load):
image

After:
image

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -710,12 +710,15 @@ namespace Cpu {
 		std::unordered_map<int, int> core_map;
 		if (cpu_temp_only) return core_map;
 
+		const int num_sensors = core_sensors.size();
+
 		//? Try to get core mapping from /proc/cpuinfo
 		ifstream cpuinfo(Shared::procPath / "cpuinfo");
 		if (cpuinfo.good()) {
 			int cpu{};
 			int core{};
-			int n{};
+			int max_core{};
+			std::unordered_map<int, int> cpu_to_core;
 			for (string instr; cpuinfo >> instr;) {
 				if (instr == "processor") {
 					cpuinfo.ignore(SSmax, ':');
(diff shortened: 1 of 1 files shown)


aristocratos:
And how does this fair on non AMD cpus?
Considering this is changing the implementation for all cpus.

unlimitedsola:
> And how does this fair on non AMD cpus? Considering this is changing the implementation for all cpus.

@aristocratos I'm expecting this won't change the behavior of CPUs that reports `core_sensors.size() == core_map.size()` or CPUs that only reports one sensor. Because in those cases `core_map[cpu_id] = core_id * num_sensors / (max_core + 1);` should always equal to the `core_id` or always be zero.

I'm not aware of CPUs that alternates the core temps, would be happy to know if that's an actual case that we need to support.

aristocratos (approved):

PR #1585, opened 2026-03-24
fix: accurate macOS memory usage reporting

scarecrowgit:
- Use host_page_size() instead of sysconf(_SC_PAGE_SIZE) to get the hardware page size for vm_statistics64 calculations, fixing a 4x undercount on Apple Silicon
    - Add other pages to formula for more accurate data

### DIFF
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -1200,10 +1200,19 @@ namespace Mem {
 
 		vm_statistics64 p;
 		mach_msg_type_number_t info_size = HOST_VM_INFO64_COUNT;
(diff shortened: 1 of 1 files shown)


scarecrowgit:
so before data was totally inaccurate like this:
Screenshot 2026-03-25 at 1 11 13 AM

after changes it become pretty much closer:
Screenshot 2026-03-25 at 1 10 57 AM

aristocratos:
@scarecrowgit 
Is this AI generated?

I'm getting exact same value `16384` from `sysconf(_SC_PAGESIZE)` and `host_page_size()` when testing on MacOs arm64.

Issue #1237, opened 2025-08-26
[BUG] Combining every second core on OpenBSD


ehawk:
On OpenBSD/SPARC64 I noticed that a pair of cores in the load graph would have the same value (such that a single threaded workload at 100% would appear on both core0 and core1). 

I assumed this was a SPARC specific issue, but after running it on amd64, the same thing happened.

Looking at the source code, this appears to be on purpose:

https://github.com/aristocratos/btop/blame/fac4a704ecae934cbfa732318249d8cec78b627b/src/openbsd/btop_collect.cpp#L391

I'm not sure why one would want this behavior, can we consider changing it?

anoduck:
@ehawk First off, OpenBSD on Sparc64 sounds incredibly cool. Second, how did you mitigate the `IPO is not supported` error?

deckstose:
`cmake -DBTOP_LTO=OFF`

deckstose:
> I'm not sure why one would want this behavior, can we consider changing it?

@joske

anoduck:
> `cmake -DBTOP_LTO=OFF`

Or I could try this... (You know, that's so simple, it's almost unfair.) 😆 

@imwints Thank you!

joske:
@imwints hmm, I forgot why this was 😅. Probably related to hyperthreading (as I only had x86)

busterb:
I just noticed this as well, and proposed a fix. LMK if it works for you.

joske:
I merged that fix. @ehawk can test again by building from source or waiting for next release.

PR #1587, opened 2026-03-26
Fix OpenBSD display of online CPUs

busterb:
This should fix issue #1237 where stats for only half of the CPU cores are effectively displayed, and their results are doubled in the output on OpenBSD.

OpenBSD by default disables multi-threading on amd64 platforms, though it can be reenabled with the hw.smt sysctl. There appears to have been an attempt earlier to deal with this behavior by dividing the CPU index in 2, which made all of the 8 threads appear to have activity in btop, when only the first 4 were actually active. The knock-on effect is any platform without SMT shows stats for only half of the cores, and the other lines are duplicates.

This fix first corrects the error in CPU stats collection by removing the CPU index divisor, and fixes an incorrect size passed to sysctl for the cpustats struct. The second change adjusts the collect loop to only collect stats for online CPUs, which produces the expected output when SMT is enabled, disabled, or does not exist in the first place.

This is an i7 with SMT enabled:
image

SMT disabled:
image

And a 12 core big/little ARM cpu where I've got the scheduler tuned to deprioritize cores 2-5 (how I first noticed this bug):
image

### DIFF
--- a/src/openbsd/btop_collect.cpp
+++ b/src/openbsd/btop_collect.cpp
@@ -127,7 +127,7 @@ namespace Shared {
 		//? Shared global variables init
 		int mib[2];
 		mib[0] = CTL_HW;
-		mib[1] = HW_NCPU;
+		mib[1] = HW_NCPUONLINE;
 		int ncpu;
 		size_t len = sizeof(ncpu);
 		if (sysctl(mib, 2, &ncpu, &len, nullptr, 0) == -1) {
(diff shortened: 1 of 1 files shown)


joske (approved):
These changes make sense to me. I have less ability to test than you, so I trust you tested it.

busterb:
Thanks! This will help with the next release, where LPE cores will be disabled from the scheduler by default as well.

Issue #190, opened 2021-12-08
[REQUEST] Report CPU frequency per core


frainz-de:
**Is your feature request related to a problem? Please describe.**
I was worried why my processor was at a low clock speed even though a process was using an entire core. Using `grep MHz /proc/cpuinfo` I noticed that the core in use was indeed at a much higher speed. It would be good to show this in btop.

**Describe the solution you'd like**
Instead of on top, the clock frequency is displayed next to the usage of each core.

**Describe alternatives you've considered**
Instead, we could write "avg" or something before the clock speed to make it more obvious that it is an average

pc00per:
This is very much needed. CPU usage percentage is not same as per core frequency. This is needed to check whether frequencies are well within the range of idle frequency.

fery25:
I'd really need it too.

Mobmaker55:
This would be incredibly helpful!

Faione:
This is very useful in heterogeneous systems, especially on ARM platforms. Is there any plan for it?

djhunter67:
I agree, this feature would be greatly appreciated.  I came to the `issues` specifically to request this.

smoluu:
Bump

Shuddown:
I'd really like to see this as well! Would be useful for MacOS.

kristof-dragon:
+1 for this feature. Would be usefule especially with the heterogeneous cores.

AlekseyMelikov:
Necessary and useful feature

FlavioMili:
Anyone ever started developing this? I could give it a try

FlavioMili:
https://github.com/aristocratos/btop/pull/1593

if anyone wants to give a look and help me for other platforms :)

PR #1593, opened 2026-03-27
issue #190:  added frequency per core on linux

FlavioMili:
The PR is AI assisted, for the UI I didn't want to spend too much time on calculating width and different things, I just found it working right away this way and I didn't need to learn the API of fmt :D.

I decided to display only up to 4 characters by representing either xxxM or x.xG for MHz and GHz values, it should be enough and saves 2 characters to keep the core consumption bar long enough.

Hopefully someone can also add the frequency for the other platforms, I don't know how to test those.

Please tell me if you need me to edit or improve something, thank you.

### DIFF
--- a/.github/workflows/cmake-netbsd.yml
+++ b/.github/workflows/cmake-netbsd.yml
@@ -46,7 +46,11 @@ jobs:
             export PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
             export PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/${{ matrix.arch }}/${{ matrix.version }}/All/"
             /usr/sbin/pkg_add pkgin
-            pkgin -y install cmake gcc14 git ninja-build
+            # retry logic for truncation errors
+            for i in 1 2 3; do
+              pkgin -y install cmake gcc14 git ninja-build && break
+              sleep 2
+            done
           run: |
             export CXX="/usr/pkg/gcc14/bin/g++"
             cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -548,7 +548,11 @@ namespace Cpu {
 		if (Runner::stopping) return "";
 		if (force_redraw) redraw = true;
 		bool show_temps = (Config::getB("check_temp") and got_sensors);
-		bool show_freq = Config::getB("show_cpu_freq");
+		#ifdef __linux__
+			bool show_freq = Config::getB("show_cpu_freq");
+		#else
+			bool show_freq = false;
+		#endif
 		bool show_watts = (Config::getB("show_cpu_watts") and supports_watts);
 		auto single_graph = Config::getB("cpu_single_graph");
 		bool hide_cores = show_temps and (cpu_temp_only or not Config::getB("show_coretemp"));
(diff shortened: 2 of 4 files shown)


.github/workflows/cmake-netbsd.yml
```diff
@@ -46,7 +46,11 @@ jobs:
             export PATH="/usr/pkg/sbin:/usr/pkg/bin:$PATH"
             export PKG_PATH="https://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/${{ matrix.arch }}/${{ matrix.version }}/All/"
             /usr/sbin/pkg_add pkgin
-            pkgin -y install cmake gcc14 git ninja-build
```
vincele, line 49: I think that should not be in this PR

src/btop_draw.cpp
```diff
@@ -548,6 +548,11 @@ namespace Cpu {
 		if (Runner::stopping) return "";
 		if (force_redraw) redraw = true;
 		bool show_temps = (Config::getB("check_temp") and got_sensors);
+		#ifdef __linux__
```
vincele, line 551: Just thinking out loud here...

Cannot `Config::getB(...)` be made to return false by default, and then enabling it (true) only for platforms where it is working ?

That would alleviate the need for those `#ifdef`s...

.github/workflows/cmake-netbsd.yml
FlavioMili: I honestly thought so but after adding that and the retry, it stopped failing. Let me see better what I can do

vincele: What I wanted to say is, even if this is needed to fix something in the CI, it should be handled separately from this PR.

But, this is just IMHO, I'll let the maintainer decide if he's OK with it here. Don't be distracted by my nitpicking. ;-)

Just a random passer-by...

src/btop_draw.cpp
FlavioMili: I had written it like this because I had seen (just a couple lines up) that `#ifdef`s are used around the code so I thought it would be ok.

I think it would be easier to do if the frequency per core was not included in the `show_cpu_freq` but at the same time it would make sense to be set only if we are actually showing the cpu freqs. 
In that case I would need to add settings and also some additional settings that appear only in special case.

I think at this point it would be easier to just make sure that everything works and compiles on any platform, thing that at the moment I have no opportunity to try. Do you have any suggestion on how to try this?

deckstose:
> The PR is AI assisted, for the UI I didn't want to spend too much time on calculating width and different things, I just found it working right away this way and I didn't need to learn the API of fmt.

This is the exact type of code _we do not want_. You should always be doing the due diligence of verifying your AI assisted code as if you had written it yourself.

FlavioMili:
Who even said I didn't read and review the code?

Knutowskie:
> > The PR is AI assisted, for the UI I didn't want to spend too much time on calculating width and different things, I just found it working right away this way and I didn't need to learn the API of fmt.
> 
> This is the exact type of code _we do not want_. You should always be doing the due diligence of verifying your AI assisted code as if you had written it yourself.

I wanted to ask if you can add this feature then? We need it to see if the cpu handles right.

FlavioMili:
> > > The PR is AI assisted, for the UI I didn't want to spend too much time on calculating width and different things, I just found it working right away this way and I didn't need to learn the API of fmt.
> > 
> > 
> > This is the exact type of code _we do not want_. You should always be doing the due diligence of verifying your AI assisted code as if you had written it yourself.
> 
> I wanted to ask if you can add this feature then? We need it to see if the cpu handles right.

You could try with my PR branch

Issue #1596, opened 2026-04-03
A way to hide the user column in processes.


Yahddyyp:
I want to hide my user when i post my dotfiles on a place like reddit. There should be a setting to remove or hide the user column or a way for it to not show the username.

ebignumber:
Would something like this fix your issue?

![Image](https://github.com/user-attachments/assets/06e3fd91-6d41-4db8-a665-2aabf5a034d3)

ebignumber:
I've created a fork that fixes this issue. In case my changes don't get merged, you can clone my repo, change the branch, and build it.

Yahddyyp:
thanks man

deckstose:
I don't think this is the right way to solve the problem of "protecting private information in screenshots".

If your intent is to hide private information hiding the user column is not enough.

ebignumber:
There is another pull request that can be helpful for this: #1573 This one allow people to hide their I.P addresses. Are there other changes that you would suggest?

deckstose:
The command column will probably display a path with the user's home directory in it and would leak the user name.

ebignumber:
Okay, I made it so changing the setting would hide the user column as well as the command column. You can view my pull request here: #1600 

![Image](https://github.com/user-attachments/assets/a25a434e-23e6-4d14-8d0a-40558d3cb9e1)

PR #1600, opened 2026-04-06
Add show user and command switch

ebignumber:
Solves issue #1596 

This commit adds a feature to hide the user column in the proc box. This may be beneficial for those who like to showcase their themes or rices without showing their usernames.

example

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -232,6 +232,7 @@ namespace Config {
 		{"log_level", 			"#* Set loglevel for \"~/.local/state/btop.log\" levels are: \"ERROR\" \"WARNING\" \"INFO\" \"DEBUG\".\n"
 								"#* The level set includes all lower levels, i.e. \"DEBUG\" will show all logging info."},
 		{"save_config_on_exit",  "#* Automatically save current settings to config file on exit."},
+		{"hide_user_and_command", "#* Hide user and command columns in the proc box"},
 	#ifdef GPU_SUPPORT
 
 		{"nvml_measure_pcie_speeds",
@@ -357,6 +358,7 @@ namespace Config {
 		{"terminal_sync", true},
 		{"save_config_on_exit", true},
 		{"disable_mouse", false},
+		{"hide_user_and_command", false},
 	};
 	std::unordered_map<std::string_view, bool> boolsTmp;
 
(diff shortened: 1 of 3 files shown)


deckstose:
Show user is misleading then. 

I also don't like the `?`, maybe `***` or `*****`?

ebignumber:
Understood. Would you recommend a different name for the option?

ebignumber:
I have made some changes. I changed the name of the variable `show_user` to `show_user_and_command`. I also changed the `?` in the user column to `***` and the one in the command column to `*****`.

Are there any other changes that you would suggest?

deckstose:
I'd probably revert the boolean and use 'hide_user_and_command' rather than 'show'.

@aristocratos Do you have any opinions on this? I really don't like the incentive behind this, although I'm not entirely opposed to adding something like this generally.

ebignumber:
> I'd probably revert the boolean and use 'hide_user_and_command' rather than 'show'.

I like that idea. I'll add that change.

aristocratos:
Yeah, I'm not sure I can see the intended audience for this feature.

People who record/screenshot their terminal and make it publicly available -> while considering their username to be private information -> but don't know how to edit the recording/screenshot to mask the information they don't want shown...

For 99.99% of people, this will just be bloat in the options.

The way to solve this that would be useful to other people would be to have toggles for which columns is shown.
Which in theory also should make it easier to extend which columns are available to show.

ebignumber:
@aristocratos If I were completely honest, I don't really see the benefit from this either. I just saw an issue that seemed easy to fix, so I did. Should I close this pull request or try to implement the toggles you mentioned?

aristocratos:
> Should I close this pull request or try to implement the toggles you mentioned?

You´re welcome to try, with the expectation that it isn't AI generated.

What I had in mind is a bit more complex though, with a separate "column editor" in the config menu and the columns generated by a function that takes a vector of "column" structs.
Each column would need separate options for min/max width, priority (which column gets hidden first if there isn't space) and so on.

Optimally this would also tie in to the collection cycle, so we don't collect information we are not showing.

ebignumber:
> You´re welcome to try, with the expectation that it isn't AI generated.
> 
> What I had in mind is a bit more complex though, with a separate "column editor" in the config menu and the columns generated by a function that takes a vector of "column" structs. Each column would need separate options for min/max width, priority (which column gets hidden first if there isn't space) and so on.
> 
> Optimally this would also tie in to the collection cycle, so we don't collect information we are not showing.

As someone who only knows the bare minimum for C++, I feel that this would be too advanced for me. I could give it a try but I don't want this task to be assigned to me.

PR #1603, opened 2026-04-08
fix: Clamp GPU graph widths to prevent crash on small terminals

PatWie:
…terminals

When many GPUs are present and the terminal is small, the GPU graph width calculations in Cpu::draw init_graphs and draw_graphs can produce negative values. This causes std::out_of_range from the Draw::Graph constructor, surfacing as "Exception in runner thread -> Cpu:: -> deque".

The root cause is the remainder-distribution formula for the last GPU graph: graph_width + graph_default_width%graph_width - gpus.size() + 1 which underflows when gpus.size() exceeds the available remainder plus graph_width. The per-GPU graph_width itself can also go below 1 when graph_default_width is small relative to gpu_draw_count.

Clamp both graph_width and the last-GPU width to a minimum of 1, and clamp the cursor movement arguments in draw_graphs to prevent negative Mv::l and Mv::r values.

---

I see

```
ERROR: Exception in runner thread -> Cpu:: -> deque
```

otherwise

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -625,7 +625,7 @@ namespace Cpu {
 						gpu_mem_graphs.resize(gpus.size());
 						gpu_meters.resize(gpus.size());
 						const int gpu_draw_count = gpu_always ? Gpu::count : Gpu::count - Gpu::shown;
-						graph_width = gpu_draw_count <= 0 ? graph_default_width : graph_default_width/gpu_draw_count - gpu_draw_count + 1 + graph_default_width%gpu_draw_count;
+						graph_width = gpu_draw_count <= 0 ? graph_default_width : max(1, graph_default_width/gpu_draw_count - gpu_draw_count + 1 + graph_default_width%gpu_draw_count);
 						for (size_t i = 0; i < gpus.size(); i++) {
 							if (gpu_auto and v_contains(Gpu::shown_panels, i))
 								continue;
@@ -638,7 +638,7 @@ namespace Cpu {
 								}
 								else {
 									graph = Draw::Graph{
-										graph_width + graph_default_width%graph_width - (int)gpus.size() + 1,
+										max(1, graph_width + graph_default_width%graph_width - (int)gpus.size() + 1),
 										graph_height, "cpu", safeVal(gpu.gpu_percent, graph_field), graph_symbol, invert, true
 									};
 								}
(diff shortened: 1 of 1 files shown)


FlintyLemming:
Really need this fix, thanks

src/btop_draw.cpp
```diff
@@ -624,7 +624,7 @@ namespace Cpu {
 						gpu_mem_graphs.resize(gpus.size());
 						gpu_meters.resize(gpus.size());
 						const int gpu_draw_count = gpu_always ? Gpu::count : Gpu::count - Gpu::shown;
-						graph_width = gpu_draw_count <= 0 ? graph_default_width : graph_default_width/gpu_draw_count - gpu_draw_count + 1 + graph_default_width%gpu_draw_count;
+						graph_width = gpu_draw_count <= 0 ? graph_default_width : max(1, graph_default_width/gpu_draw_count - gpu_draw_count + 1 + graph_default_width%gpu_draw_count);
```
dieter-apptronik:
```suggestion
                                                // Fairly distribute graph_default_width across gpu_draw_count GPUs, leaving 1 col per separator.
                                                // Clamp to >=1 to avoid degenerate/negative widths reaching Draw::Graph::_create (#1118, #1017).
                                                const int gpu_drawable_width = graph_default_width - max(0, gpu_draw_count - 1);
                                                graph_width = gpu_draw_count <= 0 ? graph_default_width : max(1, gpu_drawable_width / gpu_draw_count);
  ```

File: src/btop_draw.cpp [hunk 2]
```diff
@@ -637,7 +637,7 @@ namespace Cpu {
 								}
 								else {
 									graph = Draw::Graph{
-										graph_width + graph_default_width%graph_width - (int)gpus.size() + 1,
+										max(1, graph_width + graph_default_width%graph_width - (int)gpus.size() + 1),
```
dieter-apptronik:
```suggestion
                                                                                max(1, graph_width + (gpu_draw_count > 0 ? gpu_drawable_width % gpu_draw_count : 0)),
  ```

dieter-apptronik (commented):
Hey @PatWie — great catch on the clamps and the `Mv::l/r` cursor guards. Your `max(1, ...)` prevents the crash on multi-GPU hosts.

  While testing this on a dual-socket EPYC 9575F with 8× NVIDIA GPUs I noticed the crash is gone but the GPU time-series panel still renders squeezed into ~60% of its allotted width. The underlying distribution formula is the other half of the
  bug:

  ```cpp
  graph_default_width/gpu_draw_count - gpu_draw_count + 1 + graph_default_width%gpu_draw_count
  ```

  The `- gpu_draw_count + 1` looks like it was meant to reserve 1 col per separator, but it subtracts once *per GPU* instead of N-1 total. Then the "last GPU" adjustment `graph_width + gdw%graph_width - gpus.size() + 1` subtracts `(N-1)` again.
  For 8 GPUs at a typical CPU-box width the combined under-allocation leaves ~40% of the panel empty.

  The two suggestions above replace the formula with a straight `drawable / count` + remainder split, where `drawable = graph_default_width - (N-1)`. They compose with the rest of your PR — your cursor-move guards stay relevant for the
  very-narrow-width case.

  Before (this PR as-is) vs after (both suggestions applied) on the same 8-GPU host:

Image

<br/>

Image

dieter-apptronik:
If you agree would you mind applying my suggestions @PatWie ? I'm not able to commit to your PR. Thanks 😄

aristocratos (approved):

PR #1611, opened 2026-04-15
Fix NVML memory reporting regression on coherent UMA platforms

parallelArchitect:
On GB10 / DGX Spark, nvmlDeviceGetMemoryInfo returns NVML_SUCCESS with total == system MemTotal (~121GB). This prevents correct UMA detection, causing incorrect VRAM reporting.

Fix: detect UMA by comparing NVML total against /proc/meminfo MemTotal. If total >= 90% of system RAM, classify as unified memory and use MemAvailable instead of MemTotal for display.

Also fix GCC 13 compatibility by replacing std::ranges::to with equivalent C++20 constructs.

Note: requires validation on GB10 / DGX Spark hardware. Author does not have access to a coherent UMA system.

References
Same root cause fixed in nvtop: https://github.com/Syllo/nvtop/pull/463
NVML unified shim for GB10: https://github.com/parallelArchitect/nvml-unified-shim
btop discussion: https://github.com/aristocratos/btop/discussions/1586

Note
Requires validation on GB10 / DGX Spark hardware. Author does not have access to a coherent UMA system.


parallelArchitect:
For anyone needing the NVML memory fix now while this PR is under review — the fix is available in this fork: https://github.com/parallelArchitect/nvml-unified-shim

PR #1615, opened 2026-04-19
Fixes two places where cpu.temp_max = 0 would result in a crach

jpyllman:
This is a simple but not best solution to cpu.temp_max being 0 and getting division by zero exceptions.


aristocratos:
Fixed in 7657384

jpyllman:
Yeah, I did see the other PR, just thought something had to happen about these crashes. Thanks for all the work.

PR #1624, opened 2026-04-24
fix: avoid SIGFPE when temp_max is zero in temperature scaling

prateek9623:
## Summary

On Linux x86-64, **integer division by zero** while scaling temperatures for the UI can raise **SIGFPE** (often reported as a “floating point exception”). This change ensures `temp_max` is never used as a zero divisor, and that hwmon critical temperature values are never stored as 0 when they would break scaling.

## Root cause

1. **`Cpu::cpu_info::temp_max` defaulted to 0** until sensors were updated, so any draw that computed `temp * 100 / cpu.temp_max` could divide by zero.
2. **Linux hwmon `crit` values** are read in millidegrees and divided by 1000. Values in the range 0–999 m°C become **0** as an `int64_t` critical temperature, unlike the `/sys/class/thermal` path which already clamps sub-1°C values to a safe default.
3. **UI code** used integer division for the 0…100 color scale. Defensive `max(1ll, temp_max)` in the draw path prevents a regression if a zero slips through from another source (e.g. drivers).

## Changes

| Area | Change |
|------|--------|
| `src/btop_shared.hpp` | Default `cpu_info::temp_max` to **95** instead of **0**. |
| `src/linux/btop_collect.cpp` | After reading hwmon `crit`, if **`crit < 1`**, set **`crit = 95`** (same idea as the thermal_zone trip-point loop). |
| `src/btop_draw.cpp` | Use **`max(1ll, cpu.temp_max)`** and **`max(1ll, gpu.temp_max)`** (where applicable) in temperature percentage clamps for the CPU box, per-core display, and GPU brief/detail views. |

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -843,7 +843,7 @@ namespace Cpu {
 			+ Theme::g("cpu").at(clamp(safeVal(cpu.cpu_percent, "total"s).back(), 0ll, 100ll)) + rjust(to_string(safeVal(cpu.cpu_percent, "total"s).back()), 4) + Theme::c("main_fg") + '%';
 		if (show_temps) {
 			const auto [temp, unit] = celsius_to(safeVal(cpu.temp, 0).back(), temp_scale);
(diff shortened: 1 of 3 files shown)


aristocratos:
Fixed in 7657384

Issue #1049, opened 2025-02-24
[BUG]: For CPUs with more that 8 cores, the temperature returned is "-1ºC"


AristeidisK:
**Read the README.md and search for similar issues before posting a bug report!**

Any bug that can be solved by just reading the [prerequisites](https://github.com/aristocratos/btop#prerequisites) section of the README will likely be ignored.

**Describe the bug**

I have a MacPro 6.1 with a Xeon Chip with 12 core, 24 threads.
The first 8 CPUs are listed with their temperatures, but the rest (C8-C22) are listed with "-1ºC".

**To Reproduce**

If you run btop on a system with more than 8 cores?

**Expected behavior**

To list temperatures on all cores

**Screenshots**

![Image](https://github.com/user-attachments/assets/8d1d6d83-2143-43f8-80d2-5092e0bb4243)

**Info (please complete the following information):**
version: 1.4.0
   - If using snap: `snap info btop`
 - Architecture: x86_64
 - Platform: OSX
 - (Linux) Kernel: 24.3.0
 - (OSX/FreeBSD) Os release version: 15.3.1 (24D70)
 - Terminal used: iterm2, warp, (mac os) terminal
 - Font used:

**Additional context**

Contents of `~/.config/btop/btop.log`

_cat ~/.config/btop/btop.log

2025/02/24 (09:21:53) | ===> btop++ v.1.4.0
2025/02/24 (09:21:53) | ERROR: Stall in Runner thread, restarting!_

imwints:
This has nothing to do with the amount of CPU cores. I have a range of machines with more cores where this is not an issue. Maybe the reported temperature by the kernel is wrong/invalid for this CPU model.

130e:
Just want to report the same bug. I think it could be a (intel?) Mac issue. For me all C0-7 cores show "-1ºC".
Device: Macbook Pro 2019, 2.4 GHz Quad-Core Intel Core i5, x86_64, Darwin Kernel 24.4.0, OSX 15.4. Here is the log:
```
2025/04/14 (14:53:51) | ===> btop++ v.1.4.0
2025/04/14 (14:53:51) | DEBUG: Running in DEBUG mode!
2025/04/14 (14:53:51) | INFO: Logger set to DEBUG
2025/04/14 (14:53:51) | DEBUG: Using locale en_US.UTF-8
2025/04/14 (14:53:51) | INFO: Running on /dev/ttys000
[...]

Issue #750, opened 2024-02-06
[BUG] macOS: no longer displays all core temps in 1.3.0


treellama:
**Describe the bug**

After upgrading to 1.3.0, I only see a couple coretemp readings on macOS. I have tried toggling coretemp off/on, read temp off/on, and configuring the CPU core map.

**To Reproduce**

Update to 1.3.0

**Expected behavior**

I expect to see temperature readings on all 8 cores.

**Screenshots**

1.2.13:

<img width="777" alt="image" src="https://github.com/aristocratos/btop/assets/3276879/7512830d-795a-4bfd-a0e2-2e72470aeb41">

1.3.0:

<img width="777" alt="image" src="https://github.com/aristocratos/btop/assets/3276879/6f1f4519-59d9-4a05-b434-75fe3a7ae719">

**Info (please complete the following information):**
 - btop++ version: 1.3.0
 - Binary: from macports
 - (If compiled) Compiler and version:
 - Architecture: [x86_64, aarch64, etc.] x86_64
 - Platform: macOS
 - Os release version: 12
 - Terminal used: Terminal.app
 - Font used: Menlo

**Additional context**

Contents of `~/.config/btop/btop.log`

```
2024/02/05 (11:09:18) | ===> btop++ v.1.3.0
2024/02/05 (11:09:18) | DEBUG: Starting in DEBUG mode!
2024/02/05 (11:09:18) | INFO: Logger set to DEBUG
2024/02/05 (11:09:18) | DEBUG: Using locale en_US.UTF-8
2024/02/05 (11:09:18) | INFO: Running on /dev/ttys002
2024/02/05 (11:09:18) | DEBUG: get_sensors(): show_coretemp=1 check_temp=1
2024/02/05 (11:09:18) | DEBUG: checking intel
2024/02/05 (11:09:18) | DEBUG: intel sensors found
2024/02/05 (11:10:34) | DEBUG: Writing new config file
2024/02/05 (11:10:34) | INFO: Quitting! Runtime: 00:01:16
```

IncPlusPlus:
Seeing similar behavior on my 2020 Intel MacBook Pro just after upgrading to Sonoma 14.2

<img width="643" alt="image" src="https://github.com/aristocratos/btop/assets/6992149/2de30765-961e-4590-ac95-52fb88ca73ce">

xiaohuirong:
[...]

PR #1628, opened 2026-04-29
[AI generated] fix: restore Intel Mac CPU temperature fallback

seanniu93:
## Summary
- Fix macOS Intel SMC fallback key formatting for per-core CPU temperatures.
- Use `%c` when formatting `KeyIndexes[core]`, producing keys like `TC0C` instead of invalid truncated ASCII-number keys like `TC48`.

## Why
- Commit `dcbdb736` changed SMC key indexing for many-core systems but left the uppercase fallback formatter as `%d`.
- On affected Intel Macs, the lowercase key path fails and the uppercase fallback is required, causing btop to display `-1C`.
- Related: #1049, #750

## Testing
- `make -B obj/osx/smc.o CXX="/usr/local/opt/llvm/bin/clang++"`

## AI Disclosure
- This PR was prepared with AI assistance.

### DIFF
--- a/src/osx/smc.cpp
+++ b/src/osx/smc.cpp
@@ -101,7 +101,7 @@ namespace Cpu {
 		long long result = getSMCTemp(key);
 		if (result == -1) {
 			// try again with C
-			snprintf(key, 5, "TC%1dC", KeyIndexes[core]);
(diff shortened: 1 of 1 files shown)


seanniu93:
Before:
image

After:
image

aristocratos (approved):

Issue #928, opened 2024-09-17
[REQUEST]Automatic switch theme from dark to light based on Gnome theme


lamyergeier:
**Is your feature request related to a problem? Please describe.**
I would like to switch from solarized light to solarized dark depending on if on Gnome it is dark theme enabled or not. 

**Describe the solution you'd like**
Automatic switch theme from dark to light based on Gnome theme, or atleast when it is starting it should be mindful of the Gnome dark theme settings

# Additional

Please give CLI option to set theme, so that I can atleast automate through a bash function

imwints:
The first problem is that there is no mapping of [Light <-> Dark] themes. There are a couple of themes that only have one variant. So btop doesn't know (and probably shouldn't care) if you're theme is dark or light. The other thing is that this would tie a CLI application to a desktop, and I'm not sure if this is good idea.

> Please give CLI option to set theme, so that I can atleast automate through a bash function

This seems reasonable.

As a workaround, this wrapper script edits the config based on the theme type you provide it.
You still need to talk to gnome and parse it's mode and pass either `dark` or `light` to the script.

```python
#!/usr/bin/env python3

import os
import subprocess
import sys
from pathlib import Path

# Mode is the first command line argument
# As an alternative, set mode based on gnome's theme type here
mode = sys.argv[1]

# Path to btop's config file in the user's home
config = (
    Path(os.path.expanduser("~"))
    .joinpath(".config")
    .joinpath("btop")
    .joinpath("btop.conf")
)

# Read the config, replacing the `color_theme` statement based on the `mode`
with config.open() as file:
    lines = [
        line if "color_theme" not in line else f'color_theme = "solarized_{mode}"\n'
        for line in file
    ]

# Write the new config
with config.open("w") as file:
    for line in lines:
        file.write(line)

# Launch btop
subprocess.run(["btop"], check=False)
```

lamyergeier:
@imwints Could you provide a CLI Option --theme?

So users can start btop as follows

btop --theme "Solarized dark"

TheurgicDuke771:
+1 for the [workaround](https://github.com/aristocratos/btop/issues/928#issuecomment-2361039253). 
Instead of passing the `mode` value manually every time you can use a third party library [darkdetect](https://github.com/albertosottile/darkdetect), and replace this below code: 
```
mode = sys.argv[1]
``` 
with this if block: 
```
if darkdetect.isDark():
    mode = 'dark'
else:
    mode = 'light'
```

[...]

PR #1629, opened 2026-04-29
[AI generated] feat: add --theme CLI option to set initial theme

Pelanglene:
## Summary

Adds a `--theme <name>` CLI flag that lets users pick the initial color theme at startup, overriding the `color_theme` value from the config file. This addresses the CLI half of [#928](https://github.com/aristocratos/btop/issues/928); the GNOME desktop auto-switch part of that issue is intentionally out of scope (would require D-Bus integration and is a separate feature).

### What it does

- Adds `--theme <name>` to `Cli` parsing (`src/btop_cli.{cpp,hpp}`), following the existing pattern used by `--filter`, `--update`, and `--themes-dir`. No short form is used, since `-t` is already taken by `--tty`.
- New `Theme::has_theme()` helper in `src/btop_theme.{cpp,hpp}` validates the requested name against the available themes list using the same matching rules (full path, stem, or filename) as `Theme::setTheme()`.
- In `src/btop.cpp`, after `Theme::updateThemes()` and before `Theme::setTheme()`:
  - If the requested name matches an available theme → write it to `Config["color_theme"]`.
  - If not → log a `Logger::warning(...)` and write `"Default"` instead, so the Options menu and the persisted config stay consistent with what is actually rendered.

Total diff: +46 / -2 across 5 files, 2 commits.

## Test plan

All checks below were run against the built binary in a headless pty (`python3 pty.fork()` with explicit window size + `LC_ALL=C.utf8`).

- [x] Build clean with the project's required toolchain (clang++-18, libc++, `-std=c++23`)
- [x] `./bin/btop --help` shows new `--theme <name>` line right after `--themes-dir`
- [x] `./bin/btop --theme` (no arg) → `error: Theme requires an argument`, exit 1
- [x] `./bin/btop --theme ""` → `error: Theme name cannot be empty`, exit 1
- [x] `./bin/btop --themes-dir ./themes --theme dracula` → log: `DEBUG: Loading theme file: ./themes/dracula.theme` (no warning)
- [x] `./bin/btop --themes-dir ./themes --theme totally_fake_theme` → log: `WARNING: Theme 'totally_fake_theme' from --theme flag not found, falling back to Default`, btop launches with Default theme and the Options menu shows `Color theme N/N → Default` (regression check — earlier draft stored the bad name in config and rendered `Color theme 43/42 → totally_fake_theme`)
- [x] CLI value beats config file: with `color_theme = "elementarish"` in config + `--theme dracula --themes-dir ./themes` on CLI → only `Loading theme file: ./themes/dracula.theme` appears in log; `elementarish` is never loaded

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -1061,13 +1061,14 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
 	//? Update list of available themes
 	Theme::updateThemes();
 
-	//? Apply theme override from --theme CLI flag if provided
+	//? Apply theme override from --theme CLI flag if provided.
+	//? On miss we deliberately do NOT touch color_theme so the value from the
+	//? config file is preserved (and not later persisted to disk as "Default").
 	if (cli.theme.has_value()) {
-		if (auto match = Theme::find_theme(cli.theme.value()); match) {
-			Config::set("color_theme", match.value());
+		if (Theme::find_theme(cli.theme.value())) {
+			Config::set("color_theme", cli.theme.value());
 		} else {
-			Logger::warning("Theme '{}' from --theme flag not found, falling back to Default", cli.theme.value());
-			Config::set("color_theme", "Default"s);
+			Logger::warning("Theme '{}' from --theme flag not found, keeping color_theme from config", cli.theme.value());
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


Pelanglene:
Just to be clear about the `[AI generated]` tag: the code was AI-assisted, but I reviewed every change before opening this PR and ran each item in the test plan against the built binary myself.

Issue #1196, opened 2025-07-20
AMD-SMI Support


chillguycode:
**Is your feature request related to a problem? Please describe.**

Yes, as the rocm-smi is depricated, some of the GPUs are not supported in rocm-smi and only supported in newer amd-smi

**Describe the solution you'd like**

Integrate or shift towards amd-smi

**Describe alternatives you've considered**

None.

**Additional context**

None

SuitDeer:
The lib rocm-smi is depricated: https://github.com/ROCm/rocm_smi_lib/blob/0090adeb26637547782a7a1d1acad5fb40dde2c6/README.md?plain=1#L2

alerque:
The link above about ROCM being deprecated is obsolete since they switched to a mono-repo. The new link is here: https://github.com/ROCm/rocm-systems/tree/develop/projects/rocm-smi-lib#%EF%B8%8F-maintenance-mode-notice-%EF%B8%8F

bcmyguest:
https://github.com/aristocratos/btop/issues/1009 somewhat related, crashes when rocm exists but doesn't fully match what is expected

bcmyguest1:
Going to take a look at this but no promises (and if someone beats me to it, no worries)

xanni:
#1540 is not quite a duplicate - this issue is about unsupported new hardware, while #1540 is about existing supported hardware which is why it is a bug and not a feature.

rnk:
My LLM got sidetracked and took a stab at this. I'm not sure it's PR-worthy, but if anyone wants to start from not-zero I uploaded it: https://github.com/aristocratos/btop/compare/main...rnk:btop:amd-smi-sysfs

mikeroySoft:
I see we have multiple forks on this... 

FWIW I work at AMD, work on the ROCm team, and I love btop so I was happy to build this out with Fable to support my R9700.

alerque:
> I work at AMD, work on the ROCm team [...]

And you rock an Intel CPU in your own machine‽

mikeroySoft:
Ouch, nice catch... Yah still waiting for my threadripper, just had to use what I had lying around!

Issue #1045, opened 2025-02-22
[REQUEST] Add support for older AMD GPUs not supported by ROCm


jroovy:
My GPU (Radeon R5) shows up in `radeontop`, but not in ROCm, since it's not officially supported by ROCm. Please add support if possible.

![Image](https://github.com/user-attachments/assets/49377740-5a46-4bec-b619-9edb5e6bfe8f)
![Image](https://github.com/user-attachments/assets/b2fb635e-2b8e-4616-af34-0838ce799010)

**Describe the solution you'd like**
See title above

**Additional context**
Operating System: Arch Linux 
Kernel Version: 6.12.15-1-lts (64-bit)
Processors: 2 × AMD A9-9420e RADEON R5, 5 COMPUTE CORES 2C+3G
Memory: 7.2 GiB of RAM
Graphics Processor: AMD Radeon R5 Graphics
Manufacturer: Dell Inc.
Product Name: Inspiron 3180

Somewhat related to this:
#956

Issue #774, opened 2024-02-15
[BUG] failing to load ROCm-SMI shared library


Calandracas606:
**Read the README.md and search for similar issues before posting a bug report!**

Any bug that can be solved by just reading the [prerequisites](https://github.com/aristocratos/btop#prerequisites) section of the README will likely be ignored.

**Describe the bug**

btop is not showing gpu info, despite librocm_smi64.so.1.0 being available

**To Reproduce**

Build btop without static linking ROCm-SMI
Have librocm_smi64.so.1.0 in a standard link path `/usr/lib`
Run btop

**Expected behavior**

btop shows gpu information

**Screenshots**

![image](https://github.com/aristocratos/btop/assets/40183138/79e3fa39-7b76-4528-829c-c741b3dad86d)

**Info (please complete the following information):**
 - btop++ version: `btop version: 1.3.2`
 - Binary: [self compiled](https://0x0.st/Hn8T.bin)
 - (If compiled) Compiler and version: `gcc (GCC) 13.2.0`
 - Architecture: [x86_64, aarch64, etc.] `x86_64-musl`
 - Platform: `Linux (void musl)`
 - (Linux) Kernel: `6.6.16_1`
 - Terminal used: `foot version: 1.16.2 -pgo +ime +graphemes -assertions`
 - Font used: `JetBrainsMonoNerdFont-Regular`

**Additional context**

```
2024/02/15 (17:19:10) | ===> btop++ v.1.3.2
2024/02/15 (17:19:10) | INFO: Logger set to DEBUG
2024/02/15 (17:19:10) | DEBUG: Setting LC_ALL=en_US.UTF-8
2024/02/15 (17:19:10) | INFO: Running on /dev/pts/5
2024/02/15 (17:19:10) | INFO: Failed to load libnvidia-ml.so, NVIDIA GPUs will not be detected: Error loading shared library libnvidia-ml.so.1: No such file or directory
2024/02/15 (17:19:10) | ERROR: ROCm SMI: Couldn't find function rsmi_init: Error loading shared library librocm_smi64.so.5: No such file or directory
2024/02/15 (17:19:10) | DEBUG: Shared::init() : Initialized.
2024/02/15 (17:19:10) | DEBUG: Loading theme file: /usr/share/btop/themes/solarized_dark.theme
2024/02/15 (17:19:14) | INFO: Quitting! Runtime: 00:00:04
```

[...]

PR #1644, opened 2026-05-03
AMD APU/iGPU sysfs driver

### LINKED ISSUE #1169: Feature Request: Use sysfs/hwmon as fallback for AMD GPU stats instead of relying solely on rocm-smi
Hi,

Currently, btop uses rocm-smi to display AMD GPU stats. This leads to issues on many systems because:

    rocm-smi is not installed by default on most AMD desktop setups.

    The full ROCm stack is large, primarily intended for HPC/GPGPU workloads, not desktop usage.

    Most users running AMD GPUs on Linux rely solely on the kernel's amdgpu driver, which already exposes all relevant metrics via sysfs/hwmon.

    Flatpak and Wayland environments often make accessing rocm-smi difficult or impossible due to sandboxing and missing paths.

Other tools like radeontop, GNOME’s “Resources”, ksysguard, or glances successfully read GPU data directly from:

    /sys/class/drm/card*/device/gpu_busy_percent

    /sys/class/drm/card*/device/hwmon/hwmon*/temp*_input

    /sys/class/drm/card*/device/pp_dpm_*

    etc.

Suggestion:
Please consider implementing a fallback (or primary path) for AMD GPU monitoring that parses values directly from sysfs and hwmon, without requiring rocm-smi.

This would make btop much more usable out of the box for AMD users and better aligned with the goal of being a lightweight, cross-platform system monitor.

Thanks for your great work on btop, and for considering this improvement!

Best regards,
Gr33n93

### LINKED ISSUE #1643: Feature request: AMD consumer APU / iGPU monitoring via amdgpu sysfs
### Problem

btop's `Rsmi` backend (rocm-smi) depends on `/dev/kfd`. On consumer
APUs (Phoenix, Strix, Strix Halo, Krackan…) and minimal installs
without the ROCm runtime, the library cannot enumerate the GPU, so
btop has no AMD GPU widget — even though the kernel exposes full
data through `amdgpu` sysfs. (amd-smi has the same `/dev/kfd`
dependency, so a future amd-smi backend would not help these systems.)

### Proposal

Add a third AMD backend that reads `/sys/class/drm/card*/device` (and
the matching `hwmon`). Pure file I/O, no library, no `dlopen`.
Activates only when rocm-smi enumerates 0 devices, so it doesn't
collide with the existing path on supported hardware.

Reads: `gpu_busy_percent`, `mem_info_vram_total/used`,
`hwmon*/temp1_input`, `hwmon*/freq1_input`,
`hwmon*/power1_average` (or `power1_input`). Skips cards that expose
none of these.

### Scope

Linux only. Vendor 0x1002 + driver `amdgpu` only. No new dependencies.

mgajda:
Pull GPU information from Linux SysFS for APUs without /dev/kfd, or systems without ROCm libs installed.

Closes #1643
Closes #1169
Related: #1045, #956

## Scope

This is an **additive** backend that activates only when rocm-smi enumerates 0 devices, so it does not change behavior on hardware where rocm-smi already works. It is deliberately *not* a migration to amd-smi (#1196 remains open and warranted for discrete GPUs that the sysfs path may not fully cover).

## Alternative approach (discrete GPUs)

@rnk drafted a parallel approach at https://github.com/rnk/btop/compare/main...amd-smi-sysfs that tries `libamd_smi.so` via `dlopen` first and falls back to sysfs. That direction may be a better fit for truly discrete AMD GPUs, but per @rnk's own testing it did not help his APU iGPU — which is why this PR uses pure sysfs for the iGPU/consumer-APU case. The two approaches are complementary; the dlopen branch is parked publicly for whoever picks up the discrete-GPU side.

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -231,6 +231,7 @@ void clean_quit(int sig) {
 #ifdef GPU_SUPPORT
 	Gpu::Nvml::shutdown();
 	Gpu::Rsmi::shutdown();
+	Gpu::Asysfs::shutdown();
 	#ifdef __APPLE__
 	Gpu::AppleSilicon::shutdown();
 	#endif
--- a/src/btop_shared.hpp
+++ b/src/btop_shared.hpp
@@ -183,6 +183,9 @@ namespace Gpu {
 	namespace Rsmi {
 		extern bool shutdown();
 	}
+	namespace Asysfs {
+		extern bool shutdown();
+	}
 	#ifdef __APPLE__
 	namespace AppleSilicon {
 		extern bool shutdown();
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -302,14 +302,12 @@ namespace Gpu {
 		struct device_paths {
 			std::filesystem::path device;
 			std::filesystem::path hwmon;
+			std::filesystem::path power;            //? empty if no power sensor
 			uint32_t pci_device_id;
 			bool has_temp;
-			bool has_power;
 			bool has_freq;
 			bool has_busy;
 			bool has_vram;
-			bool has_mem_clock;
-			long long mem_clock_max_mhz;
 		};
 
 		bool initialized = false;
@@ -400,10 +398,7 @@ namespace Shared {
 
 		if (shown_gpus.contains("amd")) {
 			Gpu::Rsmi::init();
-			//? Fall back to amdgpu sysfs (consumer iGPU / dGPU) if rocm-smi found nothing.
-			if (Gpu::Rsmi::device_count == 0) {
-				Gpu::Asysfs::init();
-			}
+			Gpu::Asysfs::init(); //? self-skips when rocm-smi already enumerated devices
 		}
 
 		if (shown_gpus.contains("intel")) {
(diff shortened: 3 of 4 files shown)


rnk:
I'm also affected by this issue (AMD APU hardware on a Framework), and I confirmed this PR fixed my issue and reported accurate results:
image

I will admit, I created a 444 line patch to fix this as well, but my LLM got sidetracked and started by dlopening libamd_smi.so, which ended up not mattering for me since I have an APU. I'll park the patch on github shortly in case it's useful as a future direction for truly discrete GPUs, but I understand it's LLM-generated vibeslop, not gold.

I asked the LLM to compare our two PRs, and it likes this one better because it's shorter and more targeted, and I agree, although I have no firsthand experience of /sys filesystem hardware monitoring.

This PR *does* seem to be LLM-assisted, based on the commit messages, if not the PR description, but it does work for me.

mgajda:
It is LLM assisted.
However it was tested by at least two humans now.

The patch was preceded by research of three possible implementation options: ROCm, amd-smi library, sysfs. The sysfs worked the best because it requires no dlopens, and does not mislead with 0 instead of missing result.
Amd-smi fails to report for my APU iGPU. ROCm seems to support older dGPUs.

NetBSD may not be supported, but fails due to unrelated build issue:
    
     Error: The process '/usr/bin/ssh' failed with exit code 1

mgajda:
# Testing requested

Tested on AMD APU iGPU (the author's hardware) and independently confirmed by @rnk on an AMD APU Framework laptop. Hardware coverage beyond APUs would be very welcome — if you have any of the following, please give this branch a try and report back:

- @geerlingguy — Raspberry Pi / arm64 + Radeon RX 6700 XT (re: #956, ROCm has no arm64 support, so this sysfs path should be the only thing that works for you)
- @jroovy — Radeon R5 / older AMD GPUs not supported by ROCm (re: #1045)
- @Gr33n93 — original sysfs/hwmon fallback request (#1169)
- @Calandracas606 — void musl / rocm-smi shared-library load failures (#774); root cause is unrelated, but this fallback should restore a useful GPU panel

No pressure if your hardware setup has moved on — flagging in case any of these are still relevant.

aristocratos:
> It is LLM assisted.

@mgajda
Read https://github.com/aristocratos/btop/blob/main/CONTRIBUTING.md

Issue #1349, opened 2025-11-05
btop still freezing on macOS


Kreeblah:
**Read the README.md and search for similar issues before posting a bug report!**

**Describe the bug**

I saw https://github.com/aristocratos/btop/issues/1292 and https://github.com/aristocratos/btop/pull/1291, so I figured I'd check whether I still experience freezes or crashes with the current source tree.  So far, I haven't seen any crashes, but I do still consistently get freezes.

**To Reproduce**

Start `btop` and let it run for a while.  It freezes eventually.

**Screenshots**

**Info (please complete the following information):**
 - btop++ version: Self-built from https://github.com/aristocratos/btop/commit/5b32da5a3a950ac866fb591ddc0f97a8b625d717
 - Binary: Self-compiled with `STRIP=false` and `DEBUG=true`
 - Architecture: arm64
 - Platform: macOS
 - (OSX/FreeBSD) Os release version: 26.1
 - Terminal used: Ghostty 1.2.3 build 12214 (the latest release build)
 - Font used: MonoLisa Variable 2.017 with Symbols Nerd Font Mono as a fallback

**Additional context**

`btop.log` contents when running in `lldb`:

```
2025/11/05 (03:35:14) | ===> btop++ v.1.4.5
2025/11/05 (03:35:14) | ERROR: Couldn't determine terminal size of "STDOUT_FILENO"!
(repeat the STDOUT_FILENO error once a second until the next line)
2025/11/05 (04:18:58) | ERROR: Stall in Runner thread, restarting!
```

`btop.log` contents when running not in `lldb` (and with the `--debug` flag):

```
2025/11/05 (13:17:06) | ===> btop++ v.1.4.5
2025/11/05 (13:17:06) | DEBUG: Running in DEBUG mode!
2025/11/05 (13:17:06) | INFO: Logger set to DEBUG
2025/11/05 (13:17:06) | DEBUG: Using locale C
2025/11/05 (13:17:06) | INFO: Running on /dev/ttys001
2025/11/05 (13:17:06) | DEBUG: TTY mode enabled: false
2025/11/05 (13:17:06) | DEBUG: get_sensors(): show_coretemp=1 check_temp=1
2025/11/05 (13:17:06) | DEBUG: checking intel
2025/11/05 (13:17:06) | DEBUG: SMC connection established
2025/11/05 (13:17:06) | DEBUG: no intel sensors found
2025/11/05 (13:43:45) | ERROR: Stall in Runner thread, restarting!
```

[...]

PR #1647, opened 2026-05-04
fix: prevent runner thread deadlock on macOS/OpenBSD/NetBSD [AI generated]

jsquyres:
This introduction was written by a human (and is slightly repetitive of what is below, but I wanted to assert some context and hopefully convince you that this is not just an AI slop PR).

Many thanks for btop!  I use it every day.  I periodically run into the MacOS btop hang cited in https://github.com/aristocratos/btop/issues/1349.  I used Claude to help me investigate how this happened, and how to fix it.

I apologize for the length of this description.  But the macOS hang is a complicated thread-related issue, and it is worth explaining both the problems and the proposed solutions.

Below is a more detailed writeup of the issues discovered.  It turns out that btop freezes in macOS in both the v4.1.6 release (i.e., from the `v1.4.6` git tag) as well the tip of the `main` branch, but for different reasons.

* The `v4.1.6` tag has 1 deadlock
* The tip of `main` has 2 deadlocks

During my investigation with Claude, it made a claim to me that I found surprising: `pthread_cancel()` does *not* invoke C++ destructors of objects on the thread-local stacks that it deletes.  After a bunch of research, I could not find any documentation definitively stating whether `pthread_cancel()` is *supposed* to invoke destructors on the thread it cancels or not.  The issue seems to be that `pthread_cancel()` is a C library and is not well defined with the C++ object model (i.e., that it should -- or should not -- unwind a thread's stack and call the destructor all corresponding objects).  It appears to be an implementation issue.  Claude tells me this:

* The POSIX standard specifies that "cancellation cleanup handlers" (registered via `pthread_cleanup_push`) are called, but says nothing about C++ destructors or stack unwinding. The interaction is implementation-defined.
* Linux (glibc): chose to implement cancellation by throwing a special exception (`__cxxabiv1::__forced_unwind`) on the target thread. This does trigger C++ stack unwinding and does call destructors — *but it can be accidentally swallowed by `catch(...)` blocks.*  See the [GCC libstdc++ exception documentation](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_exceptions.html).
* macOS (libc++): uses a different mechanism that does not trigger C++ stack unwinding.  Destructors are simply not called.

Apparently (according to Claude) this is solved in C++20:

* C++20 acknowledged this gap by introducing `std::jthread` and `std::stop_token` — cooperative cancellation where the thread checks a flag at safe points and unwinds itself via a normal return or exception. That sidesteps `pthread_cancel()` entirely.

Claude and I therefore wrote up [a proof-of-concept C++ program](https://gist.github.com/jsquyres/4f752835171e72c023e0adca90d01762) that -- at least with libc++ on macOS Tahoe 26.4.1 -- reliably seems to confirm that this is true: `pthread_cancel()` **does not** invoke C++ destructors.

This turned out to be an important point in the btop code base, and is discussed below.

-----

The majority of the rest of this document is AI written, but edited and amended a bit by a human for clarity, additional context, etc.

## Summary

btop freezes after running for a period of time on macOS (and likely OpenBSD/NetBSD). The freeze is caused by deadlocks in the stall-recovery path of `Runner::run()`, where a stalled runner thread is cancelled via `pthread_cancel` and replaced with a new one. Two distinct bugs exist, one present since v1.4.6 and a second introduced by a later refactor on `main`.

---

## Testing methodology

I was not able to reliably reproduce the issue other than just letting btop run for a (long) while.  Inducing high load on my Mac did not reliably cause the issue, nor did artificially creating a (very) large number of processes.  The root cause of the problem appears when `Runner::_runner()` takes longer than 5 seconds to collect data; macOS is apparently *usually* very good at making all the data available in far less than 5 seconds.

Hence, I typically just had let two copies of btop run for long periods of time:

1. Unmodified 1.4.6 release, installed by Homebrew
1. Hand-compiled git clone at the tip of `main` with the fixes (described below)

Eventually, the problem would occur in both copies of btop at approximately the same time.  Here's a screenshot showing a Homebrew-installed 1.4.6 btop in the bottom, and a patched / locally-compiled btop running on the top.  You can see that the Homebrew-installed btop (bottom) froze about 2.5 hours earlier; the locally-compiled btop (top) is still running / is not frozen.

Screenshot 2026-05-04 at 6 48 51 PM

Ultimately, I ended up manually injected stalls into `Runner::_runner()` to shorten the debugging cycle; I can provide that patch if it's useful.

## Pull request

This pull request contains just the fix for the tip of `main`.

Looking through the git history of this repo, it looks like most (all?) of the release tags are on `main`.  I'm therefore guessing that a `v4.1.6`-specific fix is therefore not needed.

### Root cause

The runner thread uses a `std::binary_semaphore` (`do_work`) to sleep between collection cycles:

```cpp
inline void thread_wait() { do_work.acquire(); }
```

On macOS, `binary_semaphore::acquire()` is implemented via `__libcpp_atomic_wait` / `__ulock_wait`, which are **not** POSIX thread cancellation points. When `Runner::run()` detects a stall (the runner thread's collection cycle exceeds 5 seconds) and calls `pthread_cancel()`, the cancellation is set to pending but can never be delivered if the runner thread has already finished its work and parked in `thread_wait()`. The subsequent `pthread_join()` then hangs forever.

### How the race occurs

1. Runner thread starts a collection cycle (`active` = true).
2. Collection takes long enough to trigger stall detection (> 5 seconds).
3. Runner thread finishes its work, `active` becomes false, and the thread loops back to `thread_wait()` → `binary_semaphore::acquire()`.  Main thread's `atomic_wait_for(active, true, 5000)` times out, re-checks `active` — it was true at the timeout instant.
5. Main thread calls `pthread_cancel()` — but the runner is now in `acquire()`, not at a cancellation point.
6. Main thread calls `pthread_join()` — hangs forever.

### Observed behavior

Both the main btop display **and** the clock at the top of the btop display freeze. The main thread is stuck in `pthread_join` and can no longer poll for input or update the clock.

Stack trace (from a frozen homebrew-installed btop 1.4.6):

```
thread #1 (main):  __ulock_wait → _pthread_join → Runner::run()
thread #3 (runner): __ulock_wait → __libcpp_atomic_wait  (= semaphore::acquire)
```

---

### Introducing commit

Commit [`651b50c`](https://github.com/aristocratos/btop/commit/651b50c) ("refactor: replace custom RAII lock with stdlib") (after the `v4.1.6` tag) replaced:

- `pthread_mutex_t mtx` → `std::mutex mtx`
- `thread_lock pt_lck(mtx)` → `std::lock_guard lock{mtx}`

This removed the `pthread_mutex_init()` call that was resetting the mutex after stall recovery.

### Root cause

As discussed above, `pthread_cancel()` does not reliably invoke C++ destructors on macOS.  When the runner thread is cancelled, its `lock_guard` destructor never fires, leaving `std::mutex mtx` permanently locked. The replacement thread calls `mtx.lock()` at startup and blocks forever.

### Proof

A standalone proof-of-concept program demonstrates the "`pthread_cancel()` doesn't invoke destructors on macOS" issue on macOS(): [**test_pthread_cancel.cpp**](https://gist.github.com/jsquyres/4f752835171e72c023e0adca90d01762)

```bash
c++ -std=c++17 -pthread -o test_pthread_cancel test_pthread_cancel.cpp
./test_pthread_cancel
```

Output on macOS (Apple Silicon, Apple Clang 21.0.0):

```
--- Test 1: std::mutex + lock_guard (current btop main) ---
  Destructor ran:       NO
  Mutex lockable after: NO (orphaned!)

--- Test 2: pthread_mutex_t + RAII wrapper (btop v1.4.6) ---
  Destructor ran:       NO
  Mutex lockable after: NO (orphaned!)
```

Neither case invokes the C++ destructor. 

### Why v1.4.6 doesn't have the 2nd deadlock that is on main

The v1.4.6 code did not have this problem only because `pthread_mutex_init()` reset the mutex — not because destructors ran. Specifically, the v1.4.6 code uses a custom `thread_lock` class wrapping `pthread_mutex_t`:

```cpp
explicit thread_lock(pthread_mutex_t& mtx) : pt_mutex(mtx) {
    pthread_mutex_init(&pt_mutex, nullptr);  // reinitializes the mutex!
    status = pthread_mutex_lock(&pt_mutex);
}
```

The `pthread_mutex_init()` call in the constructor **reinitializes** the mutex every time a new runner thread starts. Even though `pthread_cancel()` does not invoke the old thread's C++ destructors (leaving the mutex locked), the replacement thread's `thread_lock` constructor resets the mutex state before locking it. This masked the mutex orphan problem.

### Observed behavior

The display freezes but **the visible clock keeps updating**. The main thread is not deadlocked — it continues polling for input and updating the clock via `Runner::run("clock")`. But the runner thread (which does all data collection and drawing) is permanently blocked on `std::mutex::lock()`.

Stack trace (from a frozen hand-compiled btop on `main`):

```
thread #1 (main):  __pselect → Input::poll()  (normal — clock updates work)
thread #3 (runner): __psynch_mutexwait → std::mutex::lock() → Runner::_runner()
```

---

## Fixes

All fixes are in `src/btop.cpp`. The changes address both bugs on the current `main` branch.

### Fix 1: Wake the thread from `semaphore::acquire()` during stall recovery

In the stall-recovery path in `Runner::run()`, call `pthread_cancel()` **before** `thread_trigger()`. This sets cancellation to pending first. `thread_trigger()` then wakes the thread from `semaphore::acquire()`. Combined with setting `stopping = true`, the thread reaches `sleep_ms(1)` — a real POSIX cancellation point — where the pending cancellation is delivered.

```cpp
stopping = true;
pthread_cancel(Runner::runner_id);
thread_trigger();
```

### Fix 2: Unlock the orphaned mutex after `pthread_join()`

After `pthread_join()` succeeds, force-unlock the mutex. The cancelled thread's `lock_guard` destructor did not run, so `mtx` is permanently locked. Calling `unlock()` from a non-owning thread is technically undefined behavior per \[thread.mutex.requirements.mutex\], but `PTHREAD_MUTEX_NORMAL` (the default underlying `std::mutex` on all btop target platforms) does not check ownership, and the alternative is a permanent deadlock.

Bottom line: this is a "good enough" solution, even though it is technically undefined behavior.

```cpp
if (join_result == 0) {
    mtx.unlock();
}
```

### Fix 3: Drain unconsumed semaphore token after stall recovery

A thread cancelled mid-work never loops back to `thread_wait()` (`do_work.acquire()`), leaving the semaphore counter at 1. Drain it after `pthread_join()` so the replacement thread blocks properly on its first `thread_wait()`.

```cpp
do_work.try_acquire();
```

### Fix 4: Wake the thread in `clean_quit()` on macOS/OpenBSD/NetBSD

In `clean_quit()`, call `safe_thread_trigger()` (see below) before `pthread_join()` so a thread parked in `semaphore::acquire()` wakes, sees `Global::quitting == true`, and exits the loop cleanly. 

### Fix 5: Introduce `safe_thread_trigger()`

`binary_semaphore` has a max count of 1. Calling `release()` when the counter is already 1 is undefined behavior per \[thread.sema.cnt\]. `stop()` can post a token that the thread never consumes when it exits the `while(!quitting)` loop. `safe_thread_trigger()` drains any pre-existing token before releasing:

```cpp
inline void safe_thread_trigger() { do_work.try_acquire(); do_work.release(); }
```

## Thanks!

You made it to the bottom -- thanks for reading all of this! 🎉

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -127,6 +127,7 @@ namespace Global {
 
 namespace Runner {
 	static pthread_t runner_id;
+	void safe_thread_trigger();
 } // namespace Runner
 
 //* Handler for SIGWINCH and general resizing events, does nothing if terminal hasn't been resized unless force=true
@@ -215,6 +216,12 @@ void clean_quit(int sig) {
 	Runner::stop();
 	if (Global::_runner_started) {
 	#if defined __APPLE__ || defined __OpenBSD__ || defined __NetBSD__
+		// binary_semaphore::acquire() is not a POSIX cancellation point on macOS;
+		// pthread_join() would hang if the thread is parked in thread_wait(). Post
+		// a token so it wakes, sees Global::quitting==true, and exits the loop.
+		// safe_thread_trigger() drains any pre-existing token to avoid undefined behavior on the
+		// max-1 semaphore.
+		Runner::safe_thread_trigger();
 		if (pthread_join(Runner::runner_id, nullptr) != 0) {
 			Logger::warning("Failed to join _runner thread on exit!");
 			pthread_cancel(Runner::runner_id);
@@ -374,6 +381,9 @@ namespace Runner {
 	std::binary_semaphore do_work { 0 };
 	inline void thread_wait() { do_work.acquire(); }
 	inline void thread_trigger() { do_work.release(); }
+	// Guarantees exactly one token is posted: drains any pre-existing token
+	// before release() to avoid undefined behavior on the max-1 binary_semaphore.
+	inline void safe_thread_trigger() { do_work.try_acquire(); do_work.release(); }
 
 	//* Wrapper for raising privileges when using SUID bit
 	class gain_priv {
@@ -742,14 +752,35 @@ namespace Runner {
 			Logger::error("Stall in Runner thread, restarting!");
 			set_active(false);
 			// exit(1);
+			// pthread_cancel() marks cancellation pending before thread_trigger() wakes
+			// the thread. binary_semaphore::acquire() is not a POSIX cancellation point
+			// on macOS; sleep_ms(1) is. With cancellation already pending when the
+			// thread reaches sleep_ms(), delivery is guaranteed regardless of scheduling.
+			stopping = true;
 			pthread_cancel(Runner::runner_id);
+			thread_trigger();
 
 			// Wait for the thread to actually terminate before creating a new one
 			void* thread_result;
 			int join_result = pthread_join(Runner::runner_id, &thread_result);
 			if (join_result != 0) {
-				Logger::warning("Failed to join cancelled thread: {}", strerror(join_result));
+				Logger::error("Failed to join cancelled thread: {} — aborting to avoid deadlock", strerror(join_result));
+				Global::exit_error_msg = "Failed to join cancelled Runner thread!";
+				clean_quit(1);
 			}
+			// pthread_cancel does not reliably invoke C++ destructors on macOS,
+			// so the cancelled thread's lock_guard on mtx may not have fired.
+			// The thread is gone (joined above); unlock the orphaned mutex so
+			// the replacement thread can acquire it. Technically undefined
+			// behavior per [thread.mutex.requirements.mutex] (calling unlock
+			// from a non-owning thread), but the mutex would otherwise remain
+			// permanently locked.
+			mtx.unlock();
+			// A thread cancelled mid-work never calls do_work.acquire(), leaving the
+			// counter at 1. Drain it so the new thread blocks on its first thread_wait()
+			// rather than racing on current_conf.
+			do_work.try_acquire();
+			stopping = false;
 
 			if (pthread_create(&Runner::runner_id, nullptr, &Runner::_runner, nullptr) != 0) {
 				Global::exit_error_msg = "Failed to re-create _runner thread!";
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -735,10 +735,17 @@ namespace Runner {
 
 			//? If overlay isn't empty, print output without color and then print overlay on top
 			const bool term_sync = Config::getB("terminal_sync");
+			// Disable cancellation so pthread_cancel() cannot interrupt between
+			// sync_start and sync_end — an incomplete synchronized frame leaves
+			// the terminal holding a stale
+			// image indefinitely (observed on macOS Terminal.app).
+			int old_cancel_state = PTHREAD_CANCEL_ENABLE;
+			pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &old_cancel_state);
 			cout << (term_sync ? Term::sync_start : "") << (conf.overlay.empty()
 					? output
 					: (output.empty() ? "" : Fx::ub + Theme::c("inactive_fg") + Fx::uncolor(output)) + conf.overlay)
 				<< (term_sync ? Term::sync_end : "") << flush;
+			pthread_setcancelstate(old_cancel_state, nullptr);
 		}
 		//* ----------------------------------------------- THREAD LOOP -----------------------------------------------
 		return {};


Kreeblah:
For what it's worth, I've been running a build from this PR for the last four hours without any sign of freezing, whereas the latest release version typically freezes on me in 1-2 hours.

jsquyres:
Rebased to the tip of `main` (which is currently the `v4.1.7` tag).

fraggerfox:
@jsquyres : I applied this patch in NetBSD and ran it for a while and it would still end up in a frozen state after a while. (Usually when I have other heavy tasks running in the system).

jsquyres:
I just hit another freeze yesterday morning, as well.  Drat. 😢 

I did some more investigation, but couldn't find a definitive cause for my hang yesterday.

However, I found that if `pthread_cancel()` actually failed, we'd still end up in a deadlock.  So I've updated this PR with a fix for that.  But it seems unlikely that `pthread_cancel()` is failing.

It unfortunately can take a long time for the problem to reproduce on my laptop (it's totally random -- sometimes it happens multiple times a day, sometimes it doesn't happen for multiple days).  So if others could try out this PR -- or better yet, my debugging branch (see below) -- to gather more information, that would be great.

## Debugging branch

In order to not pollute this PR with debugging stuff, I made a 2nd branch in my fork, `fix/macos-runner-thread-deadlock-moar-logging` (https://github.com/jsquyres/btop/tree/fix/macos-runner-thread-deadlock-moar-logging) that contains two additional debugging commits:

1. Add more logging to the btop logging output that should hopefully provide more information if the hang happens again.  E.g., it emits a "proof of life" log message every 10 minutes.  In macOS, this log file is `~/.local/state/btop.log`.
2. Add a diagnostics collection script (in `scripts/collect-btop-diagnostics.sh`) that should work on macOS and NetBSD (so says Claude, at least -- I can't test NetBSD myself).  
 
To get a fresh, debugging-enabled build:

```bash
# Remove any prior build
rm -rf build

# Build it
cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug
cmake --build build

# Run it
./build/btop
```

This is as opposed to the `make`-powered build, which will do an optimized build and may not contain all the symbols we want for the stack traces, etc.

If you run with a build from this debugging branch and the resulting  btop hangs, run `scripts/collect-btop-diagnostics.sh` (in my fork's `-moar-logging` branch) to gather thread stack traces, etc.  It'll create a tarball that you can post somewhere so that I can look at it.

If you're unable to reproduce the hang with the debugging build, try with the `make`-powered build and see if an optimized build recreates the problem (i.e., just run `make` and then run `./bin/btop`, not `./build/btop`).  If that hangs, run the collection script and let's see what info we get from the optimized build.

@fraggerfox Could you try what I listed above?  Thank you!

fraggerfox:
> @fraggerfox Could you try what I listed above? Thank you!

I shall apply the patch and give it a try (takes a while before it gets hung). But the last time it was hung I did take a stacktrace.

```
Thread 2 (LWP 14598 of process 14598 ""):
#0  0x00007b4adc44741a in _sys___pselect50 () from /usr/lib/libc.so.12
#1  0x00007b4adca084e3 in __pselect50 () from /usr/lib/libpthread.so.1
#2  0x0000000153a60853 in Input::poll (timeout=1000) at src/btop_input.cpp:109
#3  0x0000000153a1e8f2 in btop_main (args=...) at src/btop.cpp:1219
#4  0x0000000153aa1770 in main (argc=2, argv=0x7f7fff790778) at src/main.cpp:11

Thread 1 (LWP 14764 of process 14598 ""):
#0  0x00007b4adc4a8e7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x00007b4adca09f4b in pthread.mutex_lock_slow () from /usr/lib/libpthread.so.1
#2  0x0000000153ab14ec in __gthread_mutex_lock (__mutex=0x153e3e800 <Runner::mtx>)
#3  0x0000000153ab15da in std::mutex::lock (this=0x153e3e800 <Runner::mtx>)
#4  0x0000000153abe27c in std::lock_guard<std::mutex>::lock_guard (this=..., __m=...)
#5  0x0000000153a18dde in Runner::_runner () at src/btop.cpp:482
#6  0x00007b4adca0c89f in pthread.create_tramp () from /usr/lib/libpthread.so.1
```

Thread 1 (bottom) seems to be the locked one, and it points to the following line

```C
		// TODO: On first glance it looks redudant with `Runner::active`.
		std::lock_guard lock {mtx};
```

https://github.com/jsquyres/btop/blob/bd20f353911614acefe0608eb81441516d501599/src/btop.cpp#L482 (lines are relative to the current PRs patch).

The trace seems to be similar to the analysis you posted above.

jsquyres:
@fraggerfox Excellent information; thank you.  It looks like there's 2 possibilities:

1. pthread_join() failed.  If that's the case, the new patch on this PR (and on the debugger thread) should fix it.
2. Or NetBSD’s std::mutex implementation did not tolerate cross-thread unlock().

Can't quite tell from that stack trace.  Let's see how the new code works for you.

jsquyres:
Fortunately (?) my the display output on my hand-compiled btop just hung again.  The logging shows a few Runner thread stalls and restarts (possibly when my laptop was really busy?):

```
2026-05-07 (16:28:58) | ERROR: Stall in Runner thread (runner_phase=drawing pause_output=false cycle=3869), restarting!
2026-05-07 (16:29:01) | WARNING: Runner thread recreated (pause_output=false Menu::active=false)
2026-05-07 (16:29:13) | ERROR: Stall in Runner thread (runner_phase=drawing pause_output=false cycle=3873), restarting!
2026-05-07 (16:29:26) | WARNING: Runner thread recreated (pause_output=false Menu::active=false)
2026-05-07 (16:29:58) | ERROR: Stall in Runner thread (runner_phase=drawing pause_output=false cycle=3887), restarting!
2026-05-07 (16:30:00) | WARNING: Runner thread recreated (pause_output=false Menu::active=false)
```

But 90+ minute later, the diagnostics show that the main thread and the Runner thread *are* running, however -- btop's logging continues.  

Claude is now theorizing that this is a actually terminal / hung output state, not a btop deadlock.  I.e., btop is running just fine, but the terminal is waiting for something (and not updating the terminal while it is waiting) that will never occur.  Specifically:

> With terminal sync enabled, output is wrapped like this:
>
>  cout << Term::sync_start
>      << output
>      << Term::sync_end
>      << flush;
>
> If the runner thread is cancelled while inside the drawing/output path, it can die after emitting sync_start but before emitting sync_end. Many terminals treat synchronized output as “hold this update until the sync end marker arrives.” If the end marker never arrives, Terminal.app can keep showing the old frame indefinitely even though btop continues to write later frames.

If you can reliably make this issue occur, Claude suggests setting "terminal_sync = false" in your btop config.

I'll look at a code workaround.

jsquyres:
I just updated both branches with a possible fix for the "getting canceled while terminal is waiting for end sync state" issue.

@fraggerfox Give that a whirl.  Thanks for your patience!

fraggerfox:
> I just updated both branches with a possible fix for the "getting canceled while terminal is waiting for end sync state" issue.
> 
> @fraggerfox Give that a whirl. Thanks for your patience!

Might take a bit longer for a reply here, but will also try to do some runs over the weekend.

deckstose:
> Claude and I therefore wrote up [a proof-of-concept C++ program](https://gist.github.com/jsquyres/4f752835171e72c023e0adca90d01762) that -- at least with libc++ on macOS Tahoe 26.4.1 -- reliably seems to confirm that this is true: pthread_cancel() does not invoke C++ destructors.

Are you using Apple Clang or LLVM Clang? And is macos a pthread platform?

Upstream LLVM is using pthread for `std::thread` and `std::mutex` under the hood, at least on Linux, so I'm curious why the switch to the standard library wrapper is surfacing such a bug.

https://github.com/llvm/llvm-project/blob/b14eb2fc0db35080e6da8df26e329e70ebae4552/libcxx/include/__thread/support/pthread.h#L106

jsquyres:
> Are you using Apple Clang or LLVM Clang? And is macos a pthread platform?

This is Apple Clang / Apple libc++ on macOS. 

macOS is pthread-based, and Claude tells me that libc++ does use pthreads underneath for `std::thread` / `std::mutex`. 

But one of the bugs fixed by this PR is not that `std::mutex` is implemented differently on macOS; it’s that `pthread_cancel()` does not portably run C++ destructors. On macOS, cancellation of a thread holding a `std::lock_guard` leaves the underlying mutex locked because the `lock_guard` destructor is skipped. So switching to the standard-library wrapper exposed the problem by relying on RAII cleanup across pthread cancellation, which POSIX does not guarantee.

Does that make sense?

fraggerfox:
> > I just updated both branches with a possible fix for the "getting canceled while terminal is waiting for end sync state" issue.
> > @fraggerfox Give that a whirl. Thanks for your patience!
> 
> Might take a bit longer for a reply here, but will also try to do some runs over the weekend.

```gdb
(gdb) thread apply all bt

Thread 2 (LWP 10527 of process 10527 ""):
#0  0x0000704cb9ea8e7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x0000704cba409f4b in pthread.mutex_lock_slow () from /usr/lib/libpthread.so.1
#2  0x00000000068b151c in __gthread_mutex_lock (__mutex=0x6c43dc8 <std::__detail::__waiter_pool_base::_S_for(void const*)::__w+264>) at /usr/pkg/gcc14/include/c++/x86_64--netbsd/bits/gthr-default.h:762
#3  0x00000000068b160a in std::mutex::lock (this=0x6c43dc8 <std::__detail::__waiter_pool_base::_S_for(void const*)::__w+264>) at /usr/pkg/gcc14/include/c++/bits/std_mutex.h:113
#4  0x00000000068be0ee in std::lock_guard<std::mutex>::lock_guard (this=0x7f7fffdb7458, __m=...) at /usr/pkg/gcc14/include/c++/bits/std_mutex.h:250
#5  0x00000000068b1883 in std::__detail::__waiter_pool_base::_M_notify (this=0x6c43dc0 <std::__detail::__waiter_pool_base::_S_for(void const*)::__w+256>, __addr=0x6c43e00 <std::__detail::__waiter_pool_base::_S_for(void const*)::__w+320>, __all=true, __bare=true) at /usr/pkg/gcc14/include/c++/bits/atomic_wait.h:242
#6  0x00000000068be196 in std::__detail::__waiter_base<std::__detail::__waiter_pool>::_M_notify (this=0x7f7fffdb74a0, __all=true, __bare=true) at /usr/pkg/gcc14/include/c++/bits/atomic_wait.h:315
#7  0x00000000068b1b99 in std::__atomic_notify_address_bare (__addr=0x6c439c8 <Runner::do_work>, __all=true) at /usr/pkg/gcc14/include/c++/bits/atomic_wait.h:474
#8  0x00000000068bf13c in std::__atomic_semaphore::_M_release (this=0x6c439c8 <Runner::do_work>, __update=1) at /usr/pkg/gcc14/include/c++/bits/semaphore_base.h:256
#9  std::counting_semaphore<1l>::release (this=0x6c439c8 <Runner::do_work>, __update=1) at /usr/pkg/gcc14/include/c++/semaphore:71
#10 0x00000000068bc48d in Runner::thread_trigger () at src/btop.cpp:376
#11 0x000000000681c3ee in Runner::run (box=..., no_update=false, force_redraw=false) at src/btop.cpp:790
#12 0x000000000681e7a6 in btop_main (args=...) at src/btop.cpp:1178
#13 0x00000000068a17a6 in main (argc=2, argv=0x7f7fffdb7f48) at src/main.cpp:11

Thread 1 (LWP 18422 of process 10527 ""):
#0  0x0000704cb9ea8e7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x0000704cba409f4b in pthread.mutex_lock_slow () from /usr/lib/libpthread.so.1
#2  0x00000000068b151c in __gthread_mutex_lock (__mutex=0x6c3e800 <Runner::mtx>) at /usr/pkg/gcc14/include/c++/x86_64--netbsd/bits/gthr-default.h:762
#3  0x00000000068b160a in std::mutex::lock (this=0x6c3e800 <Runner::mtx>) at /usr/pkg/gcc14/include/c++/bits/std_mutex.h:113
#4  0x00000000068be0ee in std::lock_guard<std::mutex>::lock_guard (this=0x704cb93ff8b0, __m=...) at /usr/pkg/gcc14/include/c++/bits/std_mutex.h:250
#5  0x0000000006818e39 in Runner::_runner () at src/btop.cpp:472
#6  0x0000704cba40c89f in pthread.create_tramp () from /usr/lib/libpthread.so.1
#7  0x0000704cb9e930e0 in ?? () from /usr/lib/libc.so.12
#8  0x00010102464c457f in ?? ()
#9  0x0000000000000000 in ?? ()
(gdb)
```

The process got stuck again after a while, this was the stacktrace I took off the live running process.

fraggerfox:
[patch-src_btop.cpp](https://github.com/user-attachments/files/27564197/patch-src_btop.cpp)

After some experimenting with the semaphores / mutexes, in NetBSD claude was able to come up with a POSIX replacement which for now has not yet hung in my system.

@jsquyres : Let me know if applying this patch builds it correctly in MacOS, since I ran this only in NetBSD.

jsquyres:
@fraggerfox I'll check out your patch, thanks!  Can you run the full `scripts/collect-btop-diagnostics.sh` script that is on my fork?  It gathers several things.

jsquyres:
Ah, I see what you did: it is not merely that C++ destructors may not run. It is also that C++ standard-library synchronization internals may be left in a bad state if the thread is cancelled while inside them.  The cancelled old runner may have left two mutex-like things wedged:

1. `Runner::mtx`
2. `libstdc++`’s internal waiter-pool mutex used by `std::binary_semaphore`

Hence, you replaced `std::binary_semaphore` with POSIX `sem_t`.  Gotcha.

This is going to be a problem on macOS, though.  macOS does not support unnamed POSIX semaphores with `sem_init()` in the normal way. On Darwin, `sem_init()` is effectively unsupported and typically fails with ENOSYS.

Should we look at avoiding `pthread_cancel()` altogether?  I don't know anything about the history of btop, though -- git history makes mention of the runner thread taking longer than 5 seconds, but not any specific reason (e.g., a system call hanging indefinitely).  My Claude is suggesting that we should consider removing `pthread_cancel()` from normal stall recovery. Instead, use a cooperative stop + wake + join; if the runner does not exit cleanly within a timeout, log and quit (or re-exec?) rather than trying to surgically recover from a thread whose state is unknown.

jsquyres:
Just to see what it would look like, Claude and I wrote up the "try to co-operatively restart the thread, and if that fails, re-exec the whole process" approach.  It works, but it got a little messier than I wanted.  But I'm out of time for now.

I didn't want to replace the work on this branch -- especially since it's not clear to me yet which of these two approaches is better -- so I put up a new branch: https://github.com/jsquyres/btop/tree/fix/runner-cooperative-reexec.  See the git commit messages for details.  I didn't make a PR (yet?), but if it would be helpful, we could make a draft PR on this new branch.

fraggerfox:
@jsquyres : Will check this out in this week and get back to you.

jsquyres:
**Summary so far:** The underlying problem is that using `pthread_cancel()` to cancel a stalled runner thread is not guaranteed to clean up all of its state correctly on all platforms -- particularly C++ objects.  Specific problems have been seen on macOS and NetBSD.

I did some more exploration; I think there are now multiple possibilities on how to proceed here:

1. Continue on this PR: find and clean up all the places where C++ runtimes on different OS's react to `pthread_cancel()`.
     * **This seems like a losing game.**
     * We found some problems on macOS, we found some problems on NetBSD.  What other problems will we find?  Given that this is technically undefined behavior, the behavior on the different systems could change over time, too.
2. The https://github.com/jsquyres/btop/commits/fix/runner-cooperative-reexec/ branch: do not use `pthread_cancel()` at all, but instead use a use a co-operative approach to request the runner thread to exit.  If the co-operative approach fails after a short time, re-`exec()` btop itself.
     * **In hindsight, I'm not sure that this is a good approach.**
     * If the runner takes >5 seconds, it's likely because it is stuck in a call somewhere (vs. just being slow), and therefore we're likely to mainly end up re-`exec()`'ing anyway.
3. The https://github.com/jsquyres/btop/tree/fix/minimal-runner-reexec branch: whenever the runner takes >5 seconds, just re-`exec()` (don't try a co-operative approach).
    * **This is safe and reliable.**
    * The one tradeoff, at least as it is currently implemented, is that any run-time selected options will be lost upon re-exec.
4. Fork off each of the the collectors in their own short-lived sub-process (not a thread), which is easy to kill/clean up safely, even if it stalls.  Render the results in the child process and return the rendered string to the parent process.
    * **Not viable** (which is why there's no git branch for it).
    * This was investigated because the serialization/de-serialization requirements between the parent<-->child were minimal (because we're just passing the rendered string).
    * But it turns out that the parent needs the raw data so that it can track the history and render it properly.  Having the child render just the current state is therefore not sufficient.
5. The https://github.com/jsquyres/btop/tree/explore/subprocess-collectors branch: fork off each of the collectors in their own short-lived sub-processes (not a thread), each of which is easy to kill/clean up safely, even if it stalls.  Return the raw data back up to the parent, who handles all the rendering, displaying, and maintaining of historical data.
    * **This is more architecturally clean and reliable.**
    * The child just collects the data and then returns it up to the parent.  The parent can then assemble all the historical data and render it.  
    * But the serialization / de-serialization of the raw data between parent<-->child added a **large** amount of code. 
    * Claude suggests that using a serialization/de-serialization library like [msgpack-c](https://github.com/msgpack/msgpack-c) or [cereal](https://github.com/USCiLab/cereal) would **dramatically** reduce the amount of added code in btop, but would add a dependency.

At this point, I think we might need the project maintainers to chime in on how they'd like this to go.

@aristocratos -- do you have opinions here?

aristocratos:
@jsquyres 
How much C++ experience do you have?
You seem to be relying an awful lot on an LLM.

The 5 listed solutions are over-complicating the issue (probably from tunnel vision by the LLM).

If something is stalling the collection thread, the solution is to find the cause of the stall and fix it. If a fix isn't possible, we create a sub-thread via a future for that particular system call and let it work in the background until the future is ready.

If there is an issue specifically with `pthread_cancel()`, the solution would be to remove it for those specific platforms and instead exit the program with an error if the thread stalls (and possibly increase the timeout).

jsquyres:
@aristocratos I have quite a lot of systems/C experience, less specifically with C++.  Yes, I've used an LLM extensively with this issue, mainly because I don't have detailed knowledge of this code base, and threading problems are quite tricky to find.  I'm quite familiar with the underlying systems issues, though.

 > The 5 listed solutions are over-complicating the issue (probably from tunnel vision by the LLM).

Yes, fair point -- this is why I wanted to ask your opinions / intent here: this is your project, after all.  And I totally agree that at least some of the proposals got way too complicated.  I explored them, and they ended up being just... too much.  I figured they were still worth documenting so that someone else didn't bother exploring them.

> If something is stalling the collection thread, the solution is to find the cause of the stall and fix it. If a fix isn't possible, we create a sub-thread via a future for that particular system call and let it work in the background until the future is ready.

I have a question about that: was the intent of adding the runner thread to be able to handle collection stalls such as OS syscalls that take too long?  

I don't have specific records of this in my notes, but I'm pretty sure the cases we saw this happening was that the the overall runner execution just took longer than 5 seconds; it's not that something in the runner (e.g., a syscall) stalled indefinitely.  Meaning: 

1. The runner started, but for whatever reason, it ran for longer than 5 seconds
2. The main thread then called pthread_cancel()
3. The runner thread did eventually hit a pthread cancelation point and exited -- but on some platforms, this left some C++ runtime state undefined (e.g., not calling destructors).

I guess it comes down to: when a collector syscall takes "too long", do you want btop to allow it to eventually finish (via a thread or future or whatever mechanism), or do you want to kill it and try again at the next iteration?

* Allowing it to finish might be achieved by simply allowing the timeout to be infinite (or removing the runner thread altogether?)
* If a syscall is completely hung, I don't actually know how to safely force it to return (pthread_cancel() only *requests* thread cancelation, after all -- it won't actually interrupt a syscall if it is truly hung/would never have returned) without killing the entire process.

aristocratos:
> was the intent of adding the runner thread to be able to handle collection stalls such as OS syscalls that take too long?

The intent of the runner thread is to not add any latency to UI interactions. (I'm planning on rewriting this eventually since I wrote this many years ago and the design isn't the best...)

> [...] the overall runner execution just took longer than 5 seconds; it's not that something in the runner (e.g., a syscall) stalled indefinitely. 

Well, the total time for collection on MacOS (with a minimally specced M1 mini) is around 50ms. So if the total time takes more than 100x that, there is obviously one (or more) sys calls that's not behaving as expected on the newer Mac arm64 models.

There is the DebugTimer class in the Tools namespace to help find the culprit in such situations. I can't however do this myself since I don't have a Mac with this problem.
https://github.com/aristocratos/btop/blob/6e39144aaf5a6bc01b9f795010b0914431067183/src/btop_tools.hpp#L392-L423


> I guess it comes down to: when a collector syscall takes "too long", do you want btop to allow it to eventually finish (via a thread or future or whatever mechanism) [...]

The idea is to wrap any possibly slow sys calls in static future wrappers, so until we get a response from the sys call we just return the last value (or zero if no value).
With the downside that this particular value will be delayed 1 collection cycle even on systems where it isn't slow.
You can see an example of this in the disk stat collection on linux, added beacuse `statvfs` sys call can be slow.
https://github.com/aristocratos/btop/blob/6e39144aaf5a6bc01b9f795010b0914431067183/src/linux/btop_collect.cpp#L2365-L2412


> [...] or do you want to kill it and try again at the next iteration?

No, if a sys call actually completely stalls, we'll leave it like that, the future will simply never return and 0 (or no value) will be showed in btop. And I would probably consider that to be a problem with the host OS and not something for us to deal with. (It will eventually get cleaned up on exit anyways.)

fraggerfox:
> [patch-src_btop.cpp](https://github.com/user-attachments/files/27564197/patch-src_btop.cpp)
> 
> After some experimenting with the semaphores / mutexes, in NetBSD claude was able to come up with a POSIX replacement which for now has not yet hung in my system.
> 
> @jsquyres : Let me know if applying this patch builds it correctly in MacOS, since I ran this only in NetBSD.

After the above patch I was using. (I got another stall).

```
$ gdb -batch -ex "thread apply all bt" -p 20207
[New LWP 20207 of process 20207]
[Switching to LWP 16641 of process 20207]
0x00007a6200ea8e7a in ___lwp_park60 () from /usr/lib/libc.so.12

Thread 2 (LWP 20207 of process 20207 ""):
#0  0x00007a6200e47dfa in _lwp_wait () from /usr/lib/libc.so.12
#1  0x00007a620140d2a3 in pthread_join () from /usr/lib/libpthread.so.1
#2  0x000000002081c018 in Runner::run (box=..., no_update=false, force_redraw=false) at src/btop.cpp:785
#3  0x000000002081ea3b in btop_main (args=...) at src/btop.cpp:1207
#4  0x00000000208a1a5a in main (argc=2, argv=0x7f7fff773dc8) at src/main.cpp:11

Thread 1 (LWP 16641 of process 20207 ""):
#0  0x00007a6200ea8e7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x00007a6201409f4b in pthread.mutex_lock_slow () from /usr/lib/libpthread.so.1
#2  0x00007a6200f41407 in __getpwuid50 () from /usr/lib/libc.so.12
#3  0x00000000208abef3 in Proc::collect (no_update=false) at src/netbsd/btop_collect.cpp:1251
#4  0x000000002081982c in Runner::_runner () at src/btop.cpp:662
#5  0x00007a620140c89f in pthread.create_tramp () from /usr/lib/libpthread.so.1
#6  0x00007a6200e930e0 in ?? () from /usr/lib/libc.so.12
#7  0x00010102464c457f in ?? ()
#8  0x0000000000000000 in ?? ()
[Inferior 1 (process 20207) detached]
```

I got a stall on `getpwuid()` call (and this runs from the NetBSD specific btop collector), adding a caching call to this `getpwuid()` has resulted in it not getting stuck again. (but I guess this is temporary until some other call gets stuck).
- In this case waiting on a `pthread_join()`  when `getpwuid()` is not a cancellation point, 

I agree with @aristocratos here, I shall also try to use the debug timer to do some checks on NetBSD, usually the stall shows up when the system gets really busy and is very constrained on both CPU and Memory resources (in NetBSD).

aristocratos:
@fraggerfox You might wanna make some customization on the DebugTimer for this use case, to for example only print out the result to the log if it is over a certain number of microseconds, and not have to search through 1000 lines of log if this doesn't occur often.

jsquyres:
@aristocratos Got it.  I can look at this over the next few days.

The challenge is that this happens rarely (at least for me) -- e.g., once every few days of btop running continuously.  I don't know if it has something to do with sleep/wake, or a particular type of the system being busy, etc. (I can't seem to make it happen on demand, even with making all the cores be at 100% busy).  

FWIW, I saw the `DebugTimer` class, but it didn't help in my case because when `pthread_join()` causes the runner to exit early on macOS, the destructors don't fire, so it never printed out the final time elapsed.

Let me add some instrumentation to the macOS collector and see if I can identify which thing(s) is(are) being slow when the problem occurs.  This will give some solid data on where it would be worth emulating the Linux collector futures pattern.  This might take a day or three.

@fraggerfox I don't know if we want to keep pursuing the approach on this thread -- this is undefined behavior, after all, and it could therefore change without warning in the future.  That being said, as part of whatever overall solution is decided upon, we might want to make sure that the 2 deadlocks originally fixed by this PR can't happen.

aristocratos:
@jsquyres To get around the issue with the DebugTimer destructor you can create one timer at the start of a function and then use the `stop_rename_reset` method after each syscall, that way you would still see the timer triggered before the one that got deadlocked by pthread_cancel.

To fix the issue of possible deadlocks from pthread_cancel, feel free to adapt this PR to just remove it for the relevant platforms and instead exiting with an error. I'm guessing this would affect all BSD derived platforms?

A good idea would probably be to extend the timeout to something like 30 seconds to give the system some time to recover, and maybe log a warning if it exceeds 10 seconds.

jsquyres:
@aristocratos Got it.  I've got an instrumented version of btop running since last night; let's see what it turns up in terms of who-is-being-slow during the macOS stalls.  It uses pretty much exactly the strategy you mentioned (extend the timeout so that we can properly log when things go beyond 5 seconds).

jsquyres:
@aristocratos Per your suggestion, I instrumented the macOS collector to find what's actually slow, and then implemented the futures approach you proposed. Here's what I found.

### Root cause 1: Slow IOKit sensor calls

On Apple Silicon Macs, `Cpu::ThermalSensors::getSensors()` makes 77 IOKit HID Mach IPC roundtrips per cycle (on an M4 Pro — the number of sensors varies by model). Each call goes through a Mach IPC roundtrip to the IOKit HID driver in the kernel.

Normal timing: ~131ms mean, ~43ms of actual CPU time, the rest is kernel IPC blocking. But under contention (another process querying HID, memory pressure, CPU frequency ramp-up after idle), I observed up to 703ms. Five such events were seen across 3+ days of continuous monitoring, all with the system fully awake. Four of the five clustered in an 18-minute window, suggesting a transient system condition.

These are well under the 5-second timeout today, but on slower hardware, with more sensors, or under heavier system load, they could cross it. This is the same class of problem @fraggerfox saw with `getpwuid()` stalls on NetBSD.

### Root cause 2: Wall-clock timing across system suspend

`atomic_wait_for()` uses `time_ms()` (`std::chrono::system_clock`) to measure its timeout. On macOS, `system_clock` advances through Sleep, Deep Idle, and DarkWake. If the main thread is already inside `atomic_wait_for()` with `start_time` captured when the system suspends, the elapsed time instantly exceeds the 5-second timeout upon resume — even though the runner only ran for milliseconds of awake time.

I ran btop continuously for 3+ days and captured 38 anomalous runner cycles (>500ms wall time). I instrumented with four simultaneous clocks — wall, monotonic, `CLOCK_UPTIME_RAW` (pauses during sleep on macOS), and `CLOCK_THREAD_CPUTIME_ID` (actual CPU consumed) — and correlated each event against macOS `pmset -g log` power events.

Results:
- **33 of 38** were power events (Sleep/Deep Idle/DarkWake). Signature: wall=2s to 17min, uptime=0.5s, thread CPU=30-40ms, sleep gap matches pmset within seconds.
- **5 of 38** were IOKit contention (system awake, wall=0.5-0.9s, wall ≈ uptime).

No actual stall was triggered in my dataset because both threads typically freeze and resume together — `atomic_wait_for()` captures a fresh `start_time` after resume. But the window exists: if a suspend lands while the main thread is already inside `atomic_wait_for()`, the timeout fires. This narrow window likely explains why the deadlock in #1349 occurs only infrequently.

This does not affect Linux because Linux's `CLOCK_MONOTONIC` pauses during suspend, Linux collectors are much faster (~50ms vs ~180ms), and Linux already wraps its known-slow call (`statvfs`) in a future.

### Why fixing pthread_cancel cleanup doesn't work

When a stall does trigger, `pthread_cancel()` kills the runner and recreates it. But we now know that on macOS and BSDs, `pthread_cancel()` does not invoke clean up all C++ runtime state properly (e.g., on macOS, it does not invoke C++ destructors) — leaving `Runner::mtx` locked, semaphore internals wedged, or CoreFoundation reference counts leaked. The fixes on this PR addressed some of those sites, but @fraggerfox's NetBSD stack traces showed new ones appearing (libstdc++ waiter-pool mutex inside `std::binary_semaphore`, `getpwuid()` holding a libc mutex). This is whack-a-mole against platform-specific C++ runtime internals.

### The fix: futures + uptime clock + clean exit

I implemented the approach you suggested — I've opened https://github.com/aristocratos/btop/pull/1662 with two commits:

1. **Wrap sensor reads in `std::async` futures**, following the existing Linux `statvfs` pattern. Each cycle checks if the previous future is ready (non-blocking), uses cached results if not, and launches a new async. `getSensors()` is self-contained (creates all CF objects, queries sensors, returns plain C++ types, releases everything). Removes `pthread_cancel()` entirely and replaces with 10s warning + 30s clean exit.

2. **Use uptime-based timing in `atomic_wait_for()`**. Uses `CLOCK_UPTIME_RAW` on macOS, `CLOCK_UPTIME` on FreeBSD, `CLOCK_MONOTONIC` on Linux — all pause during system suspend, so suspend time doesn't count toward the stall timeout.

Measured results (44,299 sync-mode cycles over 3+ days vs 1,913 futures-mode cycles):

```
  Metric                sync       futures      change
  mean (ms)            234.9          30.8        -87%
  p50 (ms)             217.6          34.3        -84%
  p95 (ms)             426.5          37.9        -91%
  p99 (ms)             457.1          42.2        -91%
  max (ms)         1073233.1          88.3       -100%

  Outliers >500ms         19             0
  Would-stall >5s         11             0
  Would-exit >30s        n/a             0
```

The synchronous `update_sensors()` path (check future + apply cached result + launch new async) takes mean=0.1ms, max=5.3ms. The actual sensor collection continues in the background at mean=114ms — same as before, but no longer on the runner's critical path.

I think this PR (#1647) can be superseded by the new one (https://github.com/aristocratos/btop/pull/1662), since the futures approach eliminates the need for `pthread_cancel()` altogether. The cleanup fixes here are no longer necessary if `pthread_cancel()` is never called.

jsquyres:
Closing this in favor of #1662.

PR #1648, opened 2026-05-07
fix: gate Apple Silicon GPU support to unbreak intel Mac builds (#1641)

### LINKED ISSUE #1641: Regression: Apple Silicon GPU support was added for macOS unconditionally, which breaks the build
While https://github.com/aristocratos/btop/commit/3ec2aeeb01d0dba12a3bb010531da8ebee0fa5ae commit explicitly states that the code was added for Apple Silicon, it gets included on other archs, which breaks the build.
It should have been done conditionally.

NLagwal:
Solves issue #1641
This commit gates apple silicon GPU monitoring behind arm64 arch checks in the makefile and preprocessor. Unbreaks the build for Intel Macs and other such platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,14 +205,20 @@ endif()
 
 # Other platform dependent flags
 if(APPLE)
-  target_link_libraries(libbtop
-    $<LINK_LIBRARY:FRAMEWORK,CoreFoundation> $<LINK_LIBRARY:FRAMEWORK,IOKit>
-  )
-  if(BTOP_GPU)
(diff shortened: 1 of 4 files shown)


aristocratos:
Great👍
`CMakeLists.txt` needs the check for arm64 also since it currently enables GPU_SUPPORT for MacOs.

NLagwal:
Done! Almost forgot about Cmakelists.txt's check, added it now

aristocratos (approved):

aristocratos:
Thanks for contributing!
Will merge when we it gets confirmed as fixed in #1641

Issue #1042, opened 2025-02-19
[BUG] use-after-free and data races and more


ItsMeSamey:
**Describe the bug**
trying to investigate what causes #1012, i stumbled upon this

`btop` crashes on startup, dumping the following info
```bash
=================================================================
==1876597==ERROR: AddressSanitizer: heap-use-after-free on address 0x52d0000e6473 at pc 0x73835206cbd0 bp 0x7fffbd1a8280 sp 0x7fffbd1a79f8
READ of size 11 at 0x52d0000e6473 thread T0
    #0 0x73835206cbcf in printf_common /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors_format.inc:563
    #1 0x738352086d6e in vsnprintf /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1652
    #2 0x738352089139 in snprintf /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1723
    #3 0x62576d9d4640 in find_intel_gpu_dir (/home/a/Downloads/btop-1.4.0/bin/btop+0x141640) (BuildId: bf958397f3aa09f16b08bbff90b0d8b1961772cd)
    #4 0x62576d97c365 in Gpu::Intel::init() src/linux/btop_collect.cpp:1604
    #5 0x62576d97c365 in Shared::init() src/linux/btop_collect.cpp:300
    #6 0x62576d8c6753 in main src/btop.cpp:1071
    #7 0x738351a35487  (/usr/lib/libc.so.6+0x27487) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    #8 0x738351a3554b in __libc_start_main (/usr/lib/libc.so.6+0x2754b) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    #9 0x62576d8d8ae4 in _start (/home/a/Downloads/btop-1.4.0/bin/btop+0x45ae4) (BuildId: bf958397f3aa09f16b08bbff90b0d8b1961772cd)

0x52d0000e6473 is located 115 bytes inside of 32816-byte region [0x52d0000e6400,0x52d0000ee430)
freed by thread T0 here:
    #0 0x7383520fc102 in free /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:52
[...]

PR #1649, opened 2026-05-07
Fix atomic issues and data races in btop

ItsMeSamey:
This pr every race in #1042 and potentially dozens more.
Original idea was to fix the race on Config::current_preset but after i dug deeper, unless i am somehow horribly wrong, the whole locking mechanism in btop is fundamentally broken.

`atomic_wait`, `atomic_wait_for`, `atomic_lock::atomic_lock` and `atomic_lock::~atomic_lock` used wrong atomic orderings. Here is why:
- `atomic_wait` & `atomic_wait_for` used relaxed ordering for the state of the runner. This would have been file if the state of the runner was _not_ being used for synchronization but use of these function assumed it creates happen befores, the implementation does not.
	eg: `btop_input.cpp::Input::process` gaurds accesses to `Config::current_preset` by using `atomic_wait(Runner::active)` [well now it does] which promises absolutely nothing if relaxed ordering is used.
Now these use acquire ordering and `btop.cpp::Runner::set_active` uses release ordering to guarantee mutual exclusion.
- `atomic_lock::atomic_lock` used `if (wait) while (not this->atom.compare_exchange_strong(this->not_true, true));` which actually sets the `this->not_true` from false to true if it first exchange fails; this is now fixed by repeatedly setting `this->not_true` to false inside of the loop.
- `atomic_lock::atomic_lock` used `this->atom.store(true);` in the else branch which uses seq_cst which is not required here. only release ordering is
- `atomic_lock::~atomic_lock` also used `this->atom.store(false);` which uses sequential consistency as well which is not needed here; changed the ordering to release.

Apart from that, here are the rest of the changes
- `btop.cpp::_signal_handler` uses `clean_quit` in singal handler which eventually calls forbidden functions like malloc which are not allowed inside of signal handlers. Fixed that by not calling `clean_quit`. It eventually gets called as a result of input.interrupt call
- 557fbe5 adds guard before accessing the config to make sure that it is not being accessed by the ui draw call
- 677336f adds guards above 2 more potential accesses in the menu
- 327f695 added free's for 1 more leak detected by valgrind
- 5aaca91 made fail ordering relaxed since no locking is happening in that case.

NOTE: Even tho no code in this pr is written by ai, i did use ai to analyze the tsan dumps and point me to the right place.

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -292,14 +292,9 @@ static void _crash_handler(const int sig) {
 static void _signal_handler(const int sig) {
 	switch (sig) {
 		case SIGINT:
-			if (Runner::active) {
-				Global::should_quit = true;
-				Runner::stopping = true;
-				Input::interrupt();
-			}
-			else {
-				clean_quit(0);
-			}
+			Global::should_quit = true;
+			if (Runner::active) Runner::stopping = true;
+			Input::interrupt();
 			break;
 		case SIGTSTP:
 			if (Runner::active) {
@@ -365,7 +360,7 @@ namespace Runner {
 	atomic<bool> coreNum_reset (false);
 
 	static inline auto set_active(bool value) noexcept {
-		active.store(value, std::memory_order_relaxed);
+		active.store(value, std::memory_order_release);
 		active.notify_all();
 	}
 
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -353,15 +353,14 @@ void init_config(bool low_color, std::optional<std::string>& filter) {
 
 //* Manages secondary thread for collection and drawing of boxes
 namespace Runner {
-	atomic<bool> active (false);
+	atomic_waiting_lock active;
 	atomic<bool> stopping (false);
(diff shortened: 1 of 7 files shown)


deckstose:
Wouldn't it also make sense to replace the lock with a more appropriate synchronisation primitive in the same process? Thinking of a semaphore or a condition variable.

ItsMeSamey:
`atomic_wait` already does futex waitin. `atomic_wait_for` can definitely be better but cpp standard does not expose futex timed waits, so a condition variable is required.
Initially i did not do that to keep the changes small, but since you asked, i wrote a separate class to do waiting with timeouts using condition variable instead of busy looping with sleep.

There is no place a semaphore would be applicable as far as i can see.

aristocratos (dismissed):
Looks good!
Will merge when the conflict is solved.

ItsMeSamey:
Sry for the delay. Had assignments to work on.

aristocratos (approved):

aristocratos:
Thanks for contributing!

PR #1652, opened 2026-05-09
Fix btop with amd gpu on rocm 7 as in Ubuntu 26.04

### LINKED ISSUE #1651: GPU graph doesn't show with AMD GPU on ROCM 7
**Describe the bug**

Run btop on a machine with an AMD gpu and rocm 7 installed.  Currently this is affecting Ubuntu 26.04 

**To Reproduce**

Run btop on a machine with an AMD gpu and rocm 7 installed.  Currently this is affecting Ubuntu 26.04 

**Expected behavior**

Expected to see the GPU graph same as on rocm 6

**Screenshots**
Expected
Image

Actual
Image

 - btop++ version: 1.4.7+6e39144
 - Compiled from source after discovery of issue in 26.04
 - Binary: self compiled and 1.4.6 from ubuntu 26.04
 - Architecture: x86_64
 - Platform: Linux
 - (Linux) Kernel: 7.0.0-15-generic
 - Terminal used: terminator + bash
 - Font used: stock mono-spaced whatever it is.

**Additional context**
just no.  Pull request incoming.

chiluk:
Fixes #1651

The librocm library version incremented to .7, so you need to start looking for that name as well.

### DIFF
--- a/src/linux/btop_collect.cpp
+++ b/src/linux/btop_collect.cpp
@@ -1546,7 +1546,8 @@ namespace Gpu {
 				"librocm_smi64.so",
 				"librocm_smi64.so.5", // fedora
 				"librocm_smi64.so.1.0", // debian
-				"librocm_smi64.so.6"
+				"librocm_smi64.so.6",
+				"librocm_smi64.so.7" // rocm 7 support / Ubuntu 26.04
 			};
 
 			for (const auto& l : libRocAlts) {


aristocratos:
@chiluk 
I think this is a local issue on your machine?

On a Ubuntu 26.04 system with admgpu and rocm from the standard Ubuntu repositories this is symlinked correctly by default:

```bash
lrwxrwxrwx root root  18 B  Wed Apr 15 20:03:16 2026 librocm_smi64.so ⇒ librocm_smi64.so.7
lrwxrwxrwx root root  20 B  Wed Apr 15 20:03:16 2026 librocm_smi64.so.7 ⇒ librocm_smi64.so.7.8
.rw-r--r-- root root 2.1 MB Wed Apr 15 20:03:16 2026 librocm_smi64.so.7.8
```

chiluk:
That's interesting.  This was a machine that was dist-upgraded from 25.10.   I wonder if there's an upgrade bug in the packaging.  Either way I think you should probably be looking for the so.7 as well.

chiluk:
@aristocratos Well I think I partially figured this out.  I somehow have the librocm-smi64-7 package installed, but not the complete rocm metapackage installed.   I bet AI might have detected the dependency in btop and installed the single library package when I asked it to fix btop.

I still think that you should be checking for librocm_smi64.so.7 to handle this odd case.  I'll try to get a fix put together for the ubuntu and debian librocm-smi64-7 packages, so that they ensure that the librocm_smi64.so symlink is created.   Installing the rocm metapackage does have a dependency that creates the librocm_smi64.so symlink, I just haven't figured out which one yet.

mgajda:
Do I understand well that it only works for discrete GPUs? (Related: #1644.)

chiluk:
This fix should apply to discrete and igpus. (I'm also on a  framework desktop).  My system only had the librocm-smi64-7 package installed which didn't create the symlink.  The full room meta package does create the symlink through one of its dependencies.

yinghao-w:
Probably a separate issue but could there be a better way to find the correct library than just listing and iterating through all the possible version suffixes?

mgajda:
Does it mean that version 7 correctly reports the occupancy on iGPU? (Version 6 did not work for my iGPU.)

chiluk:
\

> Probably a separate issue but could there be a better way to find the correct library than just listing and iterating through all the possible version suffixes?

Yes, Ubuntu should've set up the symlink with just the librocm-smi64-7 package installed for
librocm_smi64.so ⇒ librocm_smi64.so.7 which it doesn't

> Does it mean that version 7 correctly reports the occupancy on iGPU? (Version 6 did not work for my iGPU.)

Maybe?  I have a ryzen ai max+ 395 which  uses an igpu works once you install librocm-smi64-7 and create the symlink, I didn't have any issues with btop with rocm 6 though.  So it sounds like you may have something else going on.

chiluk:
@aristocratos 
I did a little more investigation.  After installing the rocm package on ubuntu 
```
/lib/x86_64-linux-gnu$ dpkg -S librocm_smi64.so
librocm-smi-dev: /usr/lib/x86_64-linux-gnu/librocm_smi64.so
```
So you need the librocm-smi-dev package installed for btop to be functional without my submitted patch.  I still believe you should merge my change.

mgajda:
It would be nice to build without `librocm-smi-dev`, but this is maintainer's decision.
I have older 370 HX so was surprised GPU stopped working for me after update to Ubuntu 26.04 (kernel change).

deckstose:
> So you need the librocm-smi-dev package installed for btop to be functional without my submitted patch. I still believe you should merge my change.

This is obviously not desirable. It should just work with the versioned runtime package installed.

The ROCm library versioning is really screwed, and I fear this might surface other bugs as well, but I imagine this is worth the trade off of it working for .7 right now.

@aristocratos LGTM but I'll let have the last word.

araujo-fernando:
Added the change locally on top of version 1.4.7 and it works for both gpus, I will only need to manually rename the GPUs.

```
$ uname -a
Linux shirox-ii 6.17.0-29-generic #29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon May 11 10:30:58 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

It seems a much simpler fix than #1661

Captura de tela de 2026-05-20 16-03-22

chiluk:
> It would be nice to build without `librocm-smi-dev`, but this is maintainer's decision. I have older 370 HX so was surprised GPU stopped working for me after update to Ubuntu 26.04 (kernel change).

This is precisely what broke btop for me as well.  Was working in 25.10, after upgrade to 26.04 it was broken.

chiluk:
Also @aristocratos I'm one of the few community (non-canonical) Ubuntu Core Devs, I'll merge this change into the Ubuntu package as soon as you accept it here.  Ubuntu practice is to push to get approval of developers before backporting fixes into the support stream.

aristocratos (approved):

aristocratos:
Thanks for contributing!

PR #1665, opened 2026-05-18
Allow = to change update ms

jlambert360:
Fix the minor annoyance of having to press shift to raise the update interval.


deckstose (approved):
Great idea, thanks 🚀

deckstose:
Note: we could document this in the help menu (`?`).

PR #1666, opened 2026-05-19
Deduplicate vectors and strings.

max0x7ba:
Namespace-scope `const` objects _defined_ in header files have static linkage. Every translation unit including such a header gets its own private copies of namespace-scope `const` objects _defined_ in the header.

This change deduplicates namespace-scope `const` `string` and `vector<string>` objects by moving their _definitions_ from the .hpp files into corresponding .cpp files.

---

When building on a laptop with latest KDE Neon (Ubuntu 24.04 LTS) using `make -j16 CXX=g++-14 CC=gcc-14`, this change results in:

* ~4.5% faster release-mode build time (0:23.65s before, 0:22.60s after),
* ~20% smaller object file size total (7.6M before, 6.1M after),
* ~16% fewer symbols in `btop` executable (3,538 before, 2,983 after),
* ~45% smaller `.data` section of `btop` executable (21,912 bytes before, 12,056 bytes after),
* ~2.75% smaller `btop` executable size (2,224,728 bytes before, 2,163,568 bytes after),
* ~5% smaller maximum resident set size for `btop -h` invocation (5,616kB before, 5,340kB after).

It ain't much, but honest work. 

---

A more capable and resourceful developer would further remove immutable heap-allocated `string` copies of string literals that have been mmap'ped from `.rodata` section  of `btop` executable.

And, ideally, replace `const vector<string>` objects with `const initializer_list<char const*>` objects to completely eliminate heap-allocated `vector<string>` copies of objects already present in the process virtual memory area that mmap's `.rodata` section  of `btop` executable. Copying `initializer_list<char const*>` objects does not copy its backing array. Unlike copying `array<char const*, N>` objects, which also require specifying the element count in header _declarations_ without _definitions_.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -54,6 +54,33 @@ using namespace Tools;
 using namespace std::literals; // for operator""s
 namespace rng = std::ranges;
 
+const string Symbols::h_line				= "─";
+const string Symbols::v_line				= "│";
+const string Symbols::dotted_v_line		= "╎";
+const string Symbols::left_up			= "┌";
+const string Symbols::right_up			= "┐";
+const string Symbols::left_down			= "└";
+const string Symbols::right_down			= "┘";
+const string Symbols::round_left_up		= "╭";
+const string Symbols::round_right_up		= "╮";
+const string Symbols::round_left_down	= "╰";
+const string Symbols::round_right_down	= "╯";
+const string Symbols::title_left_down	= "┘";
+const string Symbols::title_right_down	= "└";
+const string Symbols::title_left			= "┐";
+const string Symbols::title_right		= "┌";
+const string Symbols::div_right			= "┤";
+const string Symbols::div_left			= "├";
+const string Symbols::div_up				= "┬";
+const string Symbols::div_down			= "┴";
+
+
+const string Symbols::up = "↑";
+const string Symbols::down = "↓";
+const string Symbols::left = "←";
+const string Symbols::right = "→";
+const string Symbols::enter = "↵";
+
 namespace Symbols {
 	const string meter = "■";
 
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -54,24 +54,24 @@ using namespace Tools;
 using namespace std::literals; // for operator""s
 namespace rng = std::ranges;
 
-const string Symbols::h_line				= "─";
-const string Symbols::v_line				= "│";
+const string Symbols::h_line			= "─";
+const string Symbols::v_line			= "│";
 const string Symbols::dotted_v_line		= "╎";
 const string Symbols::left_up			= "┌";
 const string Symbols::right_up			= "┐";
 const string Symbols::left_down			= "└";
-const string Symbols::right_down			= "┘";
+const string Symbols::right_down		= "┘";
 const string Symbols::round_left_up		= "╭";
-const string Symbols::round_right_up		= "╮";
+const string Symbols::round_right_up	= "╮";
 const string Symbols::round_left_down	= "╰";
 const string Symbols::round_right_down	= "╯";
 const string Symbols::title_left_down	= "┘";
 const string Symbols::title_right_down	= "└";
-const string Symbols::title_left			= "┐";
+const string Symbols::title_left		= "┐";
 const string Symbols::title_right		= "┌";
 const string Symbols::div_right			= "┤";
 const string Symbols::div_left			= "├";
-const string Symbols::div_up				= "┬";
+const string Symbols::div_up			= "┬";
 const string Symbols::div_down			= "┴";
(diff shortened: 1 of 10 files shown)


jwakely:
I do the C++ too and this change is good.

image

You could also change the `const std::string` objects to `const std::string_view`, to refer to the string literals instead of allocating storage to copy them into.

For the vectors, you could declare in the header:

```
extern const std::span<const std::string_view> freq_modes;
```

then define them in the .cpp file as:
```
using namespace std::literals;
static const std::array freq_modes_a{ "first"sv, "range"sv, "lowest"sv, "highest"sv, "average"sv };
const std::span<const std::string_view> Config::freq_modes(freq_modes_a);
```
but that's a bit more effort as it needs two variables per vector. Everything would be in static storage though, and initialized at compile-time, not at startup before `main`.

deckstose:
Looks good at first glance, I'll try to get to testing it this week, otherwise feel free to ping me again!

src/btop_draw.cpp
```diff
@@ -54,6 +54,33 @@ using namespace Tools;
...
+const string Symbols::round_left_down	= "╰";
+const string Symbols::round_right_down	= "╯";
+const string Symbols::title_left_down	= "┘";
+const string Symbols::title_right_down	= "└";
+const string Symbols::title_left			= "┐";
+const string Symbols::title_right		= "┌";
+const string Symbols::div_right			= "┤";
+const string Symbols::div_left			= "├";
+const string Symbols::div_up				= "┬";
+const string Symbols::div_down			= "┴";
```
deckstose: The formatting here is a bit messed up, this is what it looks like for me:

Image

Separate them from the equals sign by a single space.

src/btop_shared.cpp
```diff
@@ -329,3 +329,32 @@ auto detect_container() -> std::optional<std::string> {
...
+	{'D', "Waiting"},
+	{'Z', "Zombie"},
+	{'T', "Stopped"},
+	{'t', "Tracing"},
+	{'X', "Dead"},
+	{'x', "Dead"},
+	{'K', "Wakekill"},
+	{'W', "Unknown"},
+	{'P', "Parked"}
+};
```
deckstose, line 360: Please move that into the namespaces that already exist.

deckstose (commented):
LGTM, minus two minor issues 👍

src/btop_shared.cpp
max0x7ba: > Please move that into the namespaces that already exist.

Well, these definitions use _qualified identifiers_ (identifier preceded with the scope resolution operator `::`) to match the definition with its existing declaration in `Proc` namespace. Mismatching the identifier or its type in the definition causes a compiler error.

If these definitions are instead moved into `Proc` namespace, the identifiers would have to become _unqualified_ and, hence, their definitions would also become declarations, that may or may not match any existing header file declarations. Mismatches cause "undefined symbol" linker errors for the header declarations with no matching definitions.

The best practice is for definitions to always use _qualified identifiers_ for compile-time error detection. Re-opening namespace definition should only be done for adding more declarations into the namespace, but never for providing definitions for existing declarations.

For best results, may I rather recommend, humbly, updating all definitions in .cpp files to use _qualified identifiers_ instead, and removing the re-opening namespace definitions? E.g.:

```
// btop_shared.cpp

// 1. btop includes first, to detect btop headers missing includes for what they use.
// 2. Any 3rd-party library includes.
// 3. C++ stdlib includes.
// 4. C stdlib includes.
// 5. Platform specific includes.

// Save typing without re-opening namespace definition with namespace Proc {...}
using namespace Proc;

// Provide definitions for existing declarations. Detect mismatching definitions at compile time.
bool Proc::set_priority(pid_t pid, int priority) {...}
void Proc::proc_sorter(vector<proc_info>& proc_vec, const string& sorting, bool reverse, bool tree) {...}
```

---

While I humbly recommend using the best practices, I do not insist. Are you sure you'd like these definitions moved into `Proc` namespace?

src/btop_draw.cpp
max0x7ba: Done.

max0x7ba:
> Looks good at first glance, I'll try to get to testing it this week, otherwise feel free to ping me again!

FWIW, I've been running `btop` with these changes since before raising this PR -- i roll it, i smoke it 🤷🏻‍♂️💯😀

With some extra Makefile changes to Linux compiler and linker flags for leaner and more efficient machine code in release mode build. They remove the debug mode hardening and stack protection compiler/linker options that instrument the generated machine code with run-time checks. And remove one unnecessary level of indirection from all calls to functions with external linkage (such as btop functions declared in the header files or standard library functions like `strlen` and `memcpy`).

deckstose (approved):

Issue #221, opened 2021-12-21
[REQUEST] Group Processes in the process graph


redxtech:
**Is your feature request related to a problem? Please describe.**
When I am looking at the process list, I would like to be able to know the total memory used by a program. However, by default, there are a number of entries for each process, so I would have to manually add up all the memory usages, cpu usages, etc.

**Describe the solution you'd like**
[bottom by ClementTsang](https://clementtsang.github.io/bottom/nightly/configuration/command-line-flags/) has an option (`-g`/`--group`) that does this. It groups the processes together, so there is only one entry for `Discord`, only one entry for `vivaldi-bin` (my browser) and it shows the total usage of each one.

Optimally, this could be a feature that would be enabled/disabled with an entry in the config file.

**Describe alternatives you've considered**
Maybe it could be a keybind as well?

**Additional context**
Default behaviour:
https://i.imgur.com/lpjIxvx.png

Desired behaviour (btm screenshot):
https://i.imgur.com/AabxK9X.png

DivineMK:
This would be a great addition to btop.

raulpacheco2k:
I was looking for exactly this. It would help a lot with visualization.

DivineMK:
I found out that it is possible if you enable these options in memory tab:
- Tree view
- Aggregate

reitzig:
Can confirm that `proc_aggregate = True` together with `proc_tree = False` and manually collapsing select nodes works for me:

<img width="1052" height="304" alt="Image" src="https://github.com/user-attachments/assets/8c603f14-8336-4571-8b81-dddbfe265621" />

While it would be nice to get automatic aggregation by "application" in some way, I struggle to formulate a general rule for that.

PR #1667, opened 2026-05-19
proc: add proc_tree_auto_collapse option

zzallirog:
## Problem

In tree view, multi-process apps (Chromium / Firefox / Zen / any Electron) spawn dozens of subprocesses — `Isolated Web Content`, `GPU Process`, `Utility Process`, `Socket Process`, `RDD Process`, `WebExtensions`, etc. — that crowd everything else out of the list and make the tree unreadable.

Existing config doesn't cover this:
- `proc_aggregate` only affects resource summing, not visibility
- `proc_tree` / `proc_follow_detailed` toggle the view itself, not which subtrees start open
- there is no hotkey or config to enter tree mode with noisy subtrees pre-collapsed

## Solution

`proc_tree_auto_collapse` — integer, default `0` = disabled.

When `N > 0`, on entering tree mode any process with `≥ N` direct children is marked `collapsed = true`. The user can still expand any of them manually with `Space` / `+`.

### Design notes

- **Per-node, not whole-app.** Threshold is evaluated against each process's *direct* child count. In the screenshots below, `zen-bin` (3 direct children: `forkserver`, `GPU Process`, `plasma-browser-`) stays visible, while its child `forkserver` (20+ direct children) collapses. Logical structure remains readable instead of folding into one opaque `[+] zen-bin` blob.
- **Root + level-1 exempt.** `init` / kernel root and their direct children (`systemd`, `systemd --user`, launchd workers, etc.) are never auto-collapsed, so the top of the tree always shows the system spine.
- **One-shot on tree-mode entry.** Gated on `tree_mode_change`, not every refresh — no per-tick cost during steady tree view.
- **Opt-in.** Default `0` is no-op; existing users see no behaviour change.
- **Validated range.** `0`–`10000` checked in `Config::validate` with descriptive errors on out-of-range.

## Screenshots

**With `proc_tree_auto_collapse = 4`** — `zen-bin` visible, `forkserver` folded to a single line, system spine readable:

image

**Same view, `forkserver` manually expanded to show what's hidden** — `6/521` at the bottom, and this is only the top of the explosion:

image

## Changes

- `src/btop_config.cpp` — config key + default `0` + validation (`0` ≤ value ≤ `10000`)
- `src/btop_menu.cpp` — settings menu entry with description (`← →` to adjust)
- `src/{linux,osx,freebsd,netbsd,openbsd}/btop_collect.cpp` — collapse pass before `_tree_gen`, identical code across all 5 collectors
- `README.md` — config reference

---

*Prepared with AI assistance.*


zzallirog:
Quick context for review — I searched existing issues / PRs before writing this (auto-collapse, tree collapse, hide subprocess, browser tree, many children, etc.) and the only adjacent thread is #221 from 2021. Not a duplicate, but worth linking:

- **#221** asks for *horizontal aggregation* — one entry per app with summed stats, à la `bottom -g`. Process structure hidden entirely.
- **This PR** does *vertical collapse* — parent stays visible, noisy subtree folded behind `[+]`. Structure preserved, expand-on-demand.

Overlapping pain (browser/Electron noise), different shape of fix. In [this comment](https://github.com/aristocratos/btop/issues/221#issuecomment-3136841054) on #221, @reitzig says *"it would be nice to get automatic aggregation by 'application' in some way, I struggle to formulate a general rule for that."* This PR proposes one such rule (`direct children ≥ N`) — not as a replacement for #221, but as a step toward automating what's currently a manual-collapse workflow.

That said — if the 4-year quiet on #221 was a deliberate "we don't actually want this in btop" rather than triage backlog, totally happy to drop this. Just wanted to make the relationship explicit so the connection isn't missed. 🙂

aristocratos (changes requested):
Good idea.
But this should probably live in btop_shared.cpp like the other proc tree related functions instead of duplicated for every platform.

zzallirog:
> Good idea. But this should probably live in btop_shared.cpp like the other proc tree related functions instead of duplicated for every platform.

Done — force-pushed. One moment, completely at your discretion: the config option ended up at the bottom of the proc group. 
Alphabetically + by scope it'd sit cleaner right under proc_tree — feel free to reorder on merge however you like.

aristocratos (approved):
Thanks for contributing!

PR #1676, opened 2026-05-25
[AI generated] macOS: Implement battery wattage collection using IOKit

sharabdin1988:
[AI generated] macOS: Implement battery wattage collection using IOKit

  DESCRIPTION:
  Description
  This PR implements battery wattage collection for macOS. 
  Currently, the macOS implementation of get_battery() returns -1 for the wattage value, leaving the battery power display empty in the UI.

  Changes
   - Added IOKit calls to access AppleSmartBattery service via IORegistry.
   - Implemented calculation of power usage using InstantAmperage (or Amperage) and Voltage.
   - Calculation: abs(amperage * voltage) / 1,000,000 to get Watts.

  Verification
   - Verified on Apple Silicon Mac: battery wattage is now correctly displayed in the top-right corner next to the battery percentage.
   - The project builds successfully with make.

  [AI generated]

### DIFF
--- a/src/osx/btop_collect.cpp
+++ b/src/osx/btop_collect.cpp
@@ -949,6 +949,7 @@ namespace Cpu {
 
 		uint32_t percent = -1;
 		long seconds = -1;
+		float watts = -1;
 		string status = "discharging";
 		IOPSInfo_Wrap ps_info{};
 		if (ps_info()) {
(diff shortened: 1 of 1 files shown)


georgebastille:
Hi @sharabdin1988 — I've opened #1799 covering the same ground, and I'd rather flag that here myself than have it turn up unannounced.

Short version: your approach is sound and I've adopted it. #1799 reads the SMC `PSTR` key while the machine is running on battery, and falls back to `AppleSmartBattery` `InstantAmperage` × `Voltage` — your change, essentially — on AC, while charging, and on hardware without the key. It's credited in the commit message.

I checked your arithmetic before writing any of this, because `ioreg` displays `InstantAmperage` as `18446744073709551383` and I assumed that was a bug waiting to happen. It isn't — that's −233 in two's complement and `CFNumberGetValue` with `kCFNumberLongLongType` reads it back correctly. Your figure cross-checks exactly against the gauge's own `BatteryData.BatteryPower` field (−2713 mW against your 2.71 W).

The only reason it's the fallback rather than the primary source is refresh rate. Sampling both every 2 s on an M5:

```
   0.2s  PSTR=2.55W | AppleSmartBattery=2.73W (amp=-221 mA volt=12363 mV)
  20.2s  PSTR=4.67W | AppleSmartBattery=4.67W (amp=-378 mA volt=12349 mV)
  80.3s  PSTR=2.45W | AppleSmartBattery=2.45W (amp=-198 mA volt=12359 mV)
 140.3s  PSTR=2.12W | AppleSmartBattery=2.12W (amp=-171 mA volt=12379 mV)
```

Only the lines where the gauge changed are shown; `PSTR` moved continuously throughout. `AppleSmartBatteryManager` repolls those properties about once a minute — [documented on the developer forums](https://developer.apple.com/forums/thread/107914), matching the 60 s gaps above — so at btop's default 2000 ms update interval the gauge on its own holds the same figure for about thirty consecutive frames. The two agree closely when sampled at the same instant, so this is one quantity at two refresh rates rather than a disagreement about what's being measured.

Two things I hit that are worth folding into your PR regardless of which one lands:

1. **rdar://15394253** — the gauge sometimes sign-extends the current to 16 bits instead of 64, returning around 65000 mA. With `abs()` and no range check that renders as several hundred watts. #1799 discards anything above 200 W.
2. **`PSTR` on AC measures the whole machine, not the battery.** On my M5 idling on AC it read 16–25 W against 5.07 W actually moving through the battery. Your approach doesn't have this problem at all, which is precisely why it's the one doing the work there.

I'm not trying to elbow this out. If the maintainers would rather take yours, #1799's third commit is substantially your change and I'd be glad to see it land either way — or I'm happy to fold them into a single PR with you as co-author if that's easier to review. #1799 also fixes the time-remaining estimate, which was separately broken on macOS and is independent of the wattage question.

> Disclosure: like your PR, mine is AI generated and marked as such, and this comment was drafted the same way. Every measurement above was taken on my own hardware.

PR #1683, opened 2026-05-29
feat: add mellow theme [AI generated]

PepeRoConde:


deckstose:
Is this ai generated?

You don't need to duplicate the hex codes in comments right above the actual code.

PepeRoConde:
Yes this was indeed AI generated, sorry for not indicating it as you explicitly ask for it, and you are also right that the comments are sloppy.

PR #1691, opened 2026-06-02
Pull Request: Add Docker-based Compilation and Execution with NVIDIA GPU Support [AI generated]

GabrieleSantangelo:
## PR Description

This Pull Request introduces a multi-stage Docker build system and execution helper for btop on Linux systems. 

Historically, compiling btop++ from source requires a modern compiler (GCC 14+ or Clang 19+) which is not yet standard or easily accessible on many current Linux distributions (e.g., Ubuntu 24.04 ships with GCC 13 by default). Using Docker solves this dependency issue completely.

### Key Features

1. **Zero-Dependency Compilation**: Build btop using the gcc:14-bookworm container. There is no need to install compilers or development headers on the host machine. It is completely independent of the host Linux version.
2. **Minimal Setup for NVIDIA GPU Support**: Since NVIDIA monitoring relies on dynamic loading of NVML (libnvidia-ml.so), this Docker configuration is optimized to load NVIDIA libraries at runtime without linking static drivers or requiring complex Nvidia SDK setups at compile time.
3. **Double Workflow**:
   - **Exporting (Recommended)**: Build inside Docker and output the native btop binary directly to the host (./bin/btop).
   - **In-Container Execution**: Run btop containerized with host namespaces (--pid host, --net host, etc.) and NVIDIA GPU access passing.
4. **Makefile Integration Suggestion**: This is a minimal, standalone setup (Dockerfile + helper script). In the future, these commands could be integrated directly into the root Makefile (e.g., make docker-build, make docker-export) for an even cleaner workflow.

---

### Changes Introduced

1. **Dockerfile**: A multi-stage Docker file containing compilation (builder), binary extraction (binary), and runtime (runtime) stages.
2. **docker-build.sh**: A wrapper utility to run the Docker steps, handle compilation, manage container lifecycles, and automatically clean up temporary containers and images (docker rm / docker rmi) to keep the user's workspace tidy.
3. **src/linux/intel_gpu_top/intel_gpu_top.c**: Added #ifndef _GNU_SOURCE #define _GNU_SOURCE #endif to resolve implicit declaration of asprintf under GCC 14.
4. **README.md**: Added a details block under "Compilation Linux" explaining the new Docker workflow.

---

### Testing Done

- Verified compiling inside container on host using Ubuntu 24.04 and Docker 27.2.0.
- Verified exported binary on host runs correctly and identifies NVIDIA GPU using NVML.
- Verified container-based execution with --gpus all --pid host --net host --privileged runs successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,63 @@
+# ==============================================================================
+# STAGE 1: Build stage (using a compiler image with GCC 14+)
+# ==============================================================================
+FROM gcc:14-bookworm AS builder
+
+# Install build dependencies
+# lowdown is used for generating the man page, but it's optional. 
+# Installing it ensures we get a complete build with man pages.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    coreutils \
+    sed \
+    git \
+    make \
+    lowdown \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+# Copy the entire workspace into the build directory
+COPY . .
+
+# Clean any existing artifacts and build btop
+# GPU_SUPPORT=true is the default on Linux x86_64 but explicitly setting it guarantees it.
+# We do not use STATIC=true because dynamic loading of NVML library (libnvidia-ml.so)
+# is required for Nvidia GPU monitoring.
+RUN make clean && make GPU_SUPPORT=true
+
+# Create an install structure inside /install
+RUN make install DESTDIR=/install PREFIX=/usr
+
+# ==============================================================================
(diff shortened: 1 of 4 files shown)


deckstose:
> In-Container Execution: Run btop containerized with host namespaces (--pid host, --net host, etc.) and NVIDIA GPU access passing.

I heavily discourage from doing that.

Our toolchain baselines are recent operating systems, not ones from over two years ago. I think that this is a reasonable choice. This just adds the maintenance burden upon us for something we don't really need.

PR #1704, opened 2026-06-09
fix: UTF-8 locale detection

### LINKED ISSUE #1703: [BUG] UTF-8 locale detection
**Describe the bug**

On certain operating systems (such as NixOS) with specific locales, the locale name does not explicitly include the `UTF-8` suffix, even though the underlying codeset is indeed UTF-8. This causes `btop` to fail to detect UTF-8 support.

**To Reproduce**

1. Configure the NixOS system locale to `vi_VN` (and the language accordingly), which corresponds to `vi_VN/UTF-8` in the GUI installer.
2. Run `btop`.

**Expected behavior**

`btop` should detect the UTF-8 codeset correctly and start rendering normally without errors.

**Screenshots**

none

 - btop++ version: 

```
btop version: 1.4.7
Compiled with: g++ (15.2.0)
Configured with: cmake -DBTOP_STATIC=FALSE -DBTOP_GPU=ON
```
(from nixpkgs unstable)

 - Binary: self compiled
 - Architecture: x86_64
 - Platform: Linux
 - Linux Kernel: 7.0.11
 - Terminal used: kitty
 - Font used: FiraCode Nerd Font

**Additional context**

```
[maachue@maachu3-vm:~/nixos]$ locale
LANG=vi_VN
LC_CTYPE="vi_VN"
LC_NUMERIC=vi_VN
LC_TIME=vi_VN
LC_COLLATE="vi_VN"
LC_MONETARY=vi_VN
LC_MESSAGES="vi_VN"
LC_PAPER=vi_VN
LC_NAME=vi_VN
LC_ADDRESS=vi_VN
LC_TELEPHONE=vi_VN
LC_MEASUREMENT=vi_VN
LC_IDENTIFICATION=vi_VN
LC_ALL=

[maachue@maachu3-vm:~]$ locale -a
C
C.utf8
en_US.utf8
POSIX
vi_VN
vi_VN.utf8

[maachue@maachu3-vm:~/nixos]$ locale charmap
UTF-8

[maachue@maachu3-vm:~/nixos]$ btop
ERROR: No UTF-8 locale detected!
Use --force-utf argument to force start if you're sure your terminal can handle it.
```

As a temporary workaround, I have to run it with the `--force-utf` flag.

The UTF-8 locale check is implemented in `src/btop.cpp` (lines 939 to 1004). Instead of parsing the locale name string, I suggest using `nl_langinfo`/`nl_langinfo_l` from the `<langinfo.h>` header (POSIX.1 / POSIX.1-2008).
(macOS and potentially older BSD systems require both `<langinfo.h>` and `<xlocale.h>` to use `nl_langinfo_l`, `newlocale`, `freelocale`)

**GDB Backtrace**

none

maachue:
Fix #1703. See that issue for more detailed.

### DIFF
--- a/src/btop.cpp
+++ b/src/btop.cpp
@@ -27,6 +27,7 @@ tab-size = 4
 #include <mutex>
 #include <optional>
 #include <pthread.h>
+#include <langinfo.h>
 #include <span>
 #include <string_view>
 #ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


deckstose:
This will break systems. `nl_langinfo(CODESET)` returns a completely unrelated string for me.

maachue:
> This will break systems. `nl_langinfo(CODESET)` returns a completely unrelated string for me.

As I tested, FreeBSD 15, Fedora 44, NixOS (unstable), `nl_langinfo(CODESET)` returned `UTF-8` as expected.
Could you provide me with your OS and return value of `nl_langinfo(CODESET)`?
if your OS is macOS, I'm sorry for that, I don't have any Macbook, iMac or Mac mini to test, feel free to close this PR.

deckstose:
`ANSI_X3.4-1968` on Arch. The second item when you run `locale -m`.

maachue:
> `ANSI_X3.4-1968` on Arch. The second item when you run `locale -m`.

Did you happen to call `setlocale(LC_ALL, "")` before invoking `nl_langinfo(CODESET)`?
Test code:
```
// C++23 
#include <langinfo.h>

#include <clocale>
#include <print>
#include <string_view>

int main() {
  std::println("PRE-SETLOCALE: {}", std::setlocale(LC_ALL, nullptr));
  std::println("CODESET0: {}", nl_langinfo(CODESET));
  const auto *a = std::setlocale(LC_ALL, "");
  if (a != nullptr) {
    std::println("{}\nCODESET2: {}", a, nl_langinfo(CODESET));
    return 0;
  }

  std::println("failed");

  return 0;
}
```
Output:
```
PRE-SETLOCALE: C
CODESET0: ANSI_X3.4-1968
vi_VN
CODESET2: UTF-8
```
in my PR changes, I ensured that `setlocale` is executed before `nl_langinfo`.

deckstose:
> Did you happen to call setlocale(LC_ALL, "") before invoking nl_langinfo(CODESET)?

I didn't. Works wonderfully! However, we need to check that this works on macOS as well.

maachue:
wrong push, sorry

maachue:
Changes:
- Clean up locale logic: Completely removed the fallback logic. I think relying on `nl_langinfo` should be enough for charmap/encoding detection. Note that the fallback in the macOS branch remains untouched because I'm unfamiliar with macOS internals.
- Set global C++ locale: This change fixes an issue where logs were incorrectly displaying which locale the program uses because the global C++ locale wasn't being updated (`std::locale()` queries global C++ locale, not C). The global C and C++ locales operate independently; `std::setlocale` only updates the C-side locale despite being in the std namespace.

src/btop.cpp
```diff
@@ -936,72 +937,38 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
 	init_config(cli.low_color, cli.filter);
 
 	//? Try to find and set a UTF-8 locale
-	if (std::setlocale(LC_ALL, "") != nullptr and not std::string_view { std::setlocale(LC_ALL, "") }.contains(";")
-	and str_to_upper(s_replace((string)std::setlocale(LC_ALL, ""), "-", "")).ends_with("UTF8")) {
+	if (const auto *loc = std::setlocale(LC_ALL, "");
+	loc != nullptr and str_to_upper(s_replace((string)nl_langinfo(CODESET), "-", "")).ends_with("UTF8")) {
```
deckstose: I think you can remove the uppercasing and replacing logic. Something like

```cpp
if (nl_langinfo(CODESET) == "UTF-8") {
```

should work, no?

src/btop.cpp
```diff
@@ -936,72 +937,38 @@ static auto configure_tty_mode(std::optional<bool> force_tty) {
...
+		} else if (std::setlocale(LC_ALL, "en_US.UTF-8") != nullptr) {
+			std::locale::global(std::locale("en_US.UTF-8"));
+			Logger::debug("Setting LC_ALL=en_US.UTF-8");
+		} else {
+			Logger::warning("Failed to set macos locale, continuing anyway.");
 		}
 	#else
-		if (found.empty() and cli.force_utf) {
+		if (cli.force_utf) {
 			Logger::warning("No UTF-8 locale detected! Forcing start with --force-utf argument.");
```
deckstose: The log message is misleading then. The locale _might_ be detected properly but UTF is forced anyway.

maachue: This is the fallback logic when `nl_langinfo` fail. Can you add a bit details please?

src/btop.cpp
maachue: It works. But POSIX hasn't never defined what `nl_langinfo(CODESET)` will return (but defined that `char` pointer is always valid). But on most modern systems, it is `UTF-8`. I think we need to use `strcasecmp` from `<strings.h>` (POSIX header), and compare the returned value with `utf-8` and `utf8`
(see new commit)

maachue:
Sorry for the force push and the previous git history in this PR has been cleaned up.
Last time, I accidentally merged instead of rebasing to update the branch, so this force push fixes that.
(and I sorry for abandoned this PR for a while)

This commit contains a change:
- Used a custom case-insensitive ASCII-only comparision to avoid locale-specific issues (POSIX `strncasecmp`, `std::tolower` depend on the active locale which overridden by `std::setlocale(LC_ALL, "")`).

Ready for review. I won't make further changes without feedback.

Issue #1685, opened 2026-05-30
update timer cannot be changed (using -/+ or mouse) when show processes as tree


xzmeng:
this behavior is observed in 1.4.7

1.4.6 works fine

PR #1709, opened 2026-06-14
Fix -/+ key bindings

ebignumber:
Fixes issue #1685 while applying the inactive theme when -/+ buttons are used for expanding or contracting trees. The latter can be removed if need be.

### DIFF
--- a/src/btop_draw.cpp
+++ b/src/btop_draw.cpp
@@ -603,10 +603,10 @@ namespace Cpu {
 				+ (!Config::current_preset.has_value() ? "*" : to_string(Config::current_preset.value())) + Fx::ub + title_right;
 			Input::mouse_mappings["p"] = {button_y, x + 17, 1, 8};
 			const string update = to_string(Config::getI("update_ms")) + "ms";
-			out += fmt::format("{}{}{}", 
-				Mv::to(button_y, x + width - update.size() - 8) + title_left + Fx::b + Theme::c(is_proc_focused_and_in_tree ? "inactive_fg" : "hi_fg") + "- ",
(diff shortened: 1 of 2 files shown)


deckstose:
LGTM, could you make use of `fmt::format` for the line you edited and combine the check of `is_proc_focused` (rename it to that) and `Config::getB("proc_tree")` into another variable, please?

ebignumber:
> LGTM, could you make use of `fmt::format` for the line you edited and combine the check of `is_proc_focused` (rename it to that) and `Config::getB("proc_tree")` into another variable, please?

@deckstose I have implemented at least two of the changes you wanted; I renamed the variable to is_proc_focused and created a new variable named is_proc_focused_and_in_tree. I'm a bit confused by what you mean by "make use of fmt::format as I have never used it before". I did make use of it in my changes, but I'm not sure if this is what you want or not.

deckstose:
Not exactly, it's

```cpp
fmt::format_to(std::back_inserter(out), "{}{}{}...{}", args);
```

where every arg that was previously combined with `operator+` is a format arg.

So `Mv::to(...) + title_left + Fx::b` becomes `format_to(std::back_inserter(out), "{}{}{}", Mv::to(...), title_left, Fx::b);`

ebignumber:
I think I got it now.

deckstose (approved):
Thanks a lot!

ebignumber:
> Thanks a lot!

No, thank you and all the others for making the codebase so easy to read and contribute to that a total C++ newb such as myself could contribute. That is very impressive.

Issue #1718, opened 2026-06-20
IO graph scale breaks when IO Graph Speed value set to anything above 2047


feihtthief:
**Describe the bug**

When setting the `io_graph_speeds` value for a drive to anything above 2047 the graph scales as if the max value where a very small number

**To Reproduce**

Set `io_graph_speeds` for a drive to to 2048 and produce some IO

**Expected behavior**

Graph should scale based on provided value

**Screenshots**

Example with `io_graph_speeds` set to `/:2047`
<img width="397" height="380" alt="Image" src="https://github.com/user-attachments/assets/e23d23a7-58d5-4815-a637-4d8ccfd7e3aa" />

Example with `io_graph_speeds` set to `/:2048`
<img width="395" height="381" alt="Image" src="https://github.com/user-attachments/assets/c36b4545-7bf6-425a-8216-2309e1f3c332" />

**Info (please complete the following information):**
- btop++ version:
  - Self-compiled: `btop version: 1.4.7+d03b839`
  - `1.4.7` on Fedora Linux 44 running under WSL
- Binary: self compiled & static binary from WSL distro
- Architecture: `x86_64`
- Platform: Linux under WSL
- (Linux) Kernel: `6.6.114.1-microsoft-standard-WSL2`
- (OSX/FreeBSD) OS release version: N/A
- Terminal used: Default
- Font used: Default

**Additional context**

I suspect something is overflowing in `btop_draw.cpp` when left shifting by 20. (I'll investigate and try to provide a fix, but what I know about C++ fits in a matchbox and leaves room for the matches.)

PR #1719, opened 2026-06-20
Fix IO graph scaling for values above 2047

feihtthief:
Issue #1718


feihtthief:
I don't know C++ but I've seen this approach around the rest of the codebase so I hope it's ok.

deckstose (approved):

PR #1728, opened 2026-06-25
freebsd: improving ram detection including ARC cache and fixing overflow bug

crocidb:
Hi everyone!

I initially noticed that cached memory wasn't displayed at all in FreeBSD. Tested in FreeBSD 15.1, 15.0, 14.3 and 13.5 in systems with both ZFS and UFS filesystems. 

**EDIT**: I wrote about the issues and my proposed solution [in my blog](https://crocidb.com/post/freebsd-ate-my-ram/).

This PR fixes:

## Disk cache is not being counted

That's because it's drawing information from `vm.stats.vm.v_cache_count`, but the kernel doesn't use cache pages since FreeBSD 12. It's still there for compatibility, but returns 0 all the time. 

My solution is using `vfs.bufspace`, which gets metadata buffer cache from the filesystem and the usage of the ARC cache. ARC is the caching system for ZFS, it completely bypasses the vfs buffers. The thing is that if you're running on ZFS, `vfs.bufspace` will be zero. If you're on UFS, ARC info will always be zero. But you could be running two drivers with different filesystems.

Another thing I change slightly how the _used_ memory is calculated. Following the behavior on Linux where it considers disk cache as available memory, so now I subtract the cache bytes from the used bytes, since cache is contained in `wired` pages. `htop` does it too.

In fact, I submitted a [similar PR](https://github.com/htop-dev/htop/pull/2033) to `htop` simplifying their cache view. After this PR, `btop` and `htop` still have different heuristic due to them considering `laundry` pages as used memory.

## Memory Values Overflow

During the tests I noticed some very weird behavior when usage would sometimes cycle to lower numbers after rising. It turns out that all the memory information was being processed in an `u_int`, which is 32bit precision. This caused overflows when values went above 4GiB, displaying a very low value instead. So I switched for `uint64_t`.

---

I made no usage of AI to generate code, although some AI is part of my research process to understand how FreeBSD manages their virtual memory.

Have a good day!

### DIFF
--- a/src/freebsd/btop_collect.cpp
+++ b/src/freebsd/btop_collect.cpp
@@ -599,34 +599,61 @@ namespace Mem {
 		auto &mem = current_mem;
 		static bool snapped = (getenv("BTOP_SNAPPED") != nullptr);
 
-		int mib[4];
-		u_int memActive, memWire, cachedMem, freeMem;
+		int mib[5];
+		u_int memActive = 0, memWire = 0, freeMem = 0;
+		uint64_t cachedBytes = 0;
 		size_t len;
 
+		// active
    		len = 4; sysctlnametomib("vm.stats.vm.v_active_count", mib, &len);
 		len = sizeof(memActive);
 		sysctl(mib, 4, &(memActive), &len, nullptr, 0);
-		memActive *= Shared::pageSize;
 
+		// wired
 		len = 4; sysctlnametomib("vm.stats.vm.v_wire_count", mib, &len);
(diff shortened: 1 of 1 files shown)


jdebp:
The OpenBSD code also uses `uint_t` instead of `uint64_t` to calculate memory size, and requires the same kind of fix for 64-bit machines.  The NetBSD code was changed to use 64-bit calculations in e6d96662d8028038c507fd5ab47c187657447f7a .

PR #1736, opened 2026-06-29
README: fix typos

ruby-R53:
fix table border at the bottom, add space before parenthesis in config file explanation

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,7 +519,7 @@ See [GPU compatibility](#gpu-compatibility) section for more about compiling wit
    | `GPU_SUPPORT=<true\|false>`     | Enable/disable GPU support (Enabled by default on X86_64 Linux)         |
    | `RSMI_STATIC=true`              | To statically link the ROCm SMI library used for querying AMDGPU        |
    | `ADDFLAGS=<flags>`              | For appending flags to both compiler and linker                         |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1414,7 +1414,7 @@ proc_info_smaps = false
 #* Show proc box on left side of screen instead of right.
 proc_left = false
 
-#* (Linux) Filter processes tied to the Linux kernel(similar behavior to htop).
```
deckstose, line 1417: This is autogenerated from the config descriptions in `src/btop_config.cpp` and should also be updated.

README.md
```diff
@@ -1491,7 +1491,8 @@ background_update = true
 custom_cpu_name = ""
 
 #* Optional filter for shown disks, should be full path of a mountpoint, separate multiple values with whitespace " ".
-#* Only disks matching the filter will be shown. Prepend exclude= to only show disks not matching the filter. Examples: disk_filter="/boot /home/user", disks_filter="exclude=/boot /home/user"
```
deckstose, line 1494: As this is auto generated this doesn't make sense to format here.

ruby-R53: generated from _where_, then?

deckstose: `btop --default-config`, and as above, the source is in `src/btop_config.cpp`.

PR #1739, opened 2026-07-02
feat: show zswap usage

### LINKED ISSUE #1633: zRam & zSwap usage

When using zRam or zSwap, you dedicate an amount of memory for swap usage, that the system can compress and use for swapping before sending data to the slowere disks swap.

It would be nice too know how much of your memory Swap is actually beeing used, and what the compression ratio is of the data.  i belive the data can be found in /proc and /sys without needing sudo on most systems.  only write_back monitoring of zSwap i belive needs sudo atleast on my fedora 44 system as that data is in kernel debug area i belive.

tepo4ka:
This implements support for zswap monitoring on Linux.

Partially closes #1633.

This addes new optional entry "Zswap" to the Swap memory box which displays compressed size stored in RAM. If this entry is shown, then "Used" field uses the formula `swap_used - swap_zswap_original`, where `swap_zswap_original` is the uncompressed size stored in zswap. In other words, "Used" shows how much is stored on disk, while "Zswap" --- in RAM.

It is based on the Zswap and Zswapped entries exposed in [/proc/meminfo](https://www.kernel.org/doc/html/latest/filesystems/proc.html#meminfo). For some reason these are not documented in the corresponding manpage. Note that Linux 5.19 is required ([kernel commit](https://github.com/torvalds/linux/commit/f6498b776d280b30a4614d8261840961e993c2c8)), as previously the only way to obtain this data was through debugfs. Here is the [discussion](https://lwn.net/ml/linux-kernel/20220510152847.230957-1-hannes@cmpxchg.org/) which led to mentioned patch.

I am reusing color gradient "cached" for this entry. This feels a bit hacky, but introducing new gradient would require changing every color scheme.

Note that because we store both compressed and uncompressed size it is possible to display the compression ratio like "Zswap (2.5x)". I decided not to do it because the compression ratio is fairly stable and because I'm not sure if title is the right place for it.

Here's how it looks:
image image

### DIFF
--- a/src/btop_config.cpp
+++ b/src/btop_config.cpp
@@ -190,7 +190,9 @@ namespace Config {
 		{"zfs_arc_cached",		"#* Count ZFS ARC in cached and available memory."},
 
 		{"show_swap", 			"#* If swap memory should be shown in memory box."},
-
+	#ifdef __linux__
+		{"show_zswap", 			"#* If zswap usage should be shown in memory box."},
+	#endif
 		{"swap_disk", 			"#* Show swap as a disk, ignores show_swap value above, inserts itself after first disk."},
 
 		{"show_disks", 			"#* If mem box should be split to also show disks info."},
@@ -323,6 +325,9 @@ namespace Config {
 		{"mem_below_net", false},
 		{"zfs_arc_cached", true},
 		{"show_swap", true},
+	#ifdef __linux__
+		{"show_zswap", true},
+	#endif
 		{"swap_disk", true},
 		{"show_disks", true},
 		{"only_physical", true},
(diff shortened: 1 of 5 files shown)


Coca162:
Trying this out myself it crashes with `ERROR: Exception in runner thread -> Mem:: -> unordered_map::at` being thrown. Seems to only occur due to https://github.com/aristocratos/btop/pull/1739/commits/6a3a3382f7999a90f7c271dcdf9be8f10a7ca2b2 as the previous commit was fine.

Edit: the same error does occur as well but just at a larger terminal size, so I didn't notice it.

05126619z:
Bring glory to this pr, it should be worked on.

tepo4ka:
Sorry, I've been very busy last couple of months and am going to look into it.

I am daily driving this PR and have experienced the described crash only once. My guess is that it may be related to the theme -- I am using `gruvbox_dark_v2`. It would be helpful if someone could triage this and perhaps get a stacktrace.

PR #1787, opened 2026-08-10
fix: battery rate and time reading in mW and minutes on freebsd

SamlonDev:
this was done according to ahci_battery(4)

additionally added garbage guards since sysctl sometimes returns garbage mid-state when plugging in or unplugging the laptop.

**AI was used to research the issue, fix was done by me**


SamlonDev:
altough the garbage is returned by the battery itself, and maybe it should be instead handled by the kernel instead of making btop handle it?
I'd like to know what someone else thinks about this

Issue #1653, opened 2026-05-12
macOS: GPU temperature not displayed on Apple Silicon M5 (SMC `flt ` data type not parsed)


necoli1822:
## Summary

On Apple Silicon M5 (MacBook Air M5, macOS 26.5), the GPU box in btop shows `0 °C` for temperature, while utilisation, power, and memory all read correctly. CPU temperature is displayed correctly via the `PMU tdie` IOHID fallback already in HEAD.

Investigation shows two related causes:

1. **`getSMCTemp` only parses the `sp78` data type.** Apple Silicon GPU thermal sensors are exposed as `flt ` (4-byte IEEE float) via SMC. The legacy `sp78` parser drops these silently and returns `-1`, so every SMC GPU key returns `-1` regardless of whether it exists on the chip.
2. **No SMC-based GPU temperature reader.** The existing GPU temperature path in `btop_collect.cpp::get_gpu_temp_iohid` filters IOHID sensors by name (`"GPU"` substring or `PMU TP*g`). On M5 the IOHID sensors are named generically (`PMU tdie1`...`PMU tdie14`, `PMU2 tdie1`...`PMU2 tdie10`), so none match, and there is no fallback to SMC.

## Reproduction

- MacBook Air M5 (2026), macOS 26.5, btop built from `v1.4.7` (commit `6e39144`).
- Launch btop. GPU box header shows utilisation/power/RAM values but the temperature column is blank or `0`.
- Verified the IOHID sensor set on M5 via [Issac-Lopez/mac_temp_sensor](https://github.com/Issac-Lopez/mac_temp_sensor)'s probe binary: the only thermal sensor names are `PMU tdie1`...`PMU tdie14`, `PMU2 tdie1`...`PMU2 tdie10`, `PMU tdev1-8`, `PMU2 tdev1-5`, `NAND CH0 temp`, `PMU tcal`, `PMU2 tcal` (calibration constants), and `gas gauge battery` entries. None contain `"GPU"` or end in `"g"`.
[...]

PR #1541, opened 2026-02-14
feat: Add Apple Silicon GPU support via IOReport API

gneitzke:
Add GPU monitoring for Apple Silicon Macs using the IOReport framework and IOHIDEventSystem for temperature readings.

Metrics supported:
- GPU utilization (from GPU Performance States residency)
- GPU power consumption (from Energy Model channel)
- GPU temperature (from IOHIDEventSystem thermal sensors)
- GPU clock speed (from DVFS frequency table + residency weighting)
- Memory usage (unified memory architecture, reports system RAM)

Implementation follows the existing Nvml/Rsmi/Intel backend pattern with init/shutdown/collect<is_init> template functions in a new AppleSilicon namespace under Gpu.

Build system changes:
- Enable GPU_SUPPORT on macOS in both CMakeLists.txt and Makefile
- Link IOReport library on macOS
- Add 'apple' to shown_gpus vendor filter

Closes #955

PR #1788, opened 2026-08-10
[AI generated] fix(osx): report GPU temperature on M3 and later Apple Silicon

### LINKED ISSUE #1688: on M4 no gpu temps
**Describe the bug**

Btop shows gpu temps as 0 degrees.

**To Reproduce**

run btop.

**Expected behavior**

To display temperatures on the gpu sections.

**Screenshots**

Image

 - btop++ version: 1.4.7
 - Binary: Compiled with: clang++ (21.0.0)
 - Architecture: aarch64
 - Platform: OsX
 - (OSX/FreeBSD) Os release version: OSX 26.5
 - Terminal used: Terminal

AwesomenessReborn:
Fixes #1688
Refs #1653 (same two defects, should resolve it — see the hardware caveat under Testing)

## The problem

On Apple Silicon from M3 onwards the GPU box reports a temperature of 0. Reported
independently on M3 (#1688), M4, M4 Pro and M5 (#1653). M1 is unaffected, which is
the clue to what is going on. 

Issue is reproducible, see comment below.

## AI disclosure

Issue is identified by me as a Human. I have used AI to identify tools / SMC tools that are able to detect and display temperature readings as needed for display. The code in this PR was AI generated, hence the `[AI generated]` tag. 

I used AI code agent tools to automate direct experimentation on the affected hardware, enumerating the
IOHID sensor list, differential GPU-versus-CPU load testing to rule out the `tdie`
sensors, probing the SMC keyspace, and benchmarking probe against enumeration.

I have reviewed and verified every line. No code was copied from [mactop](https://github.com/metaspartan/mactop); it is cited
below only as prior art for the 10 °C floor and its rationale. 

Issue reproduced on my system (M4 PRO Laptop).

## Root cause 

There are two independent defects. Fixing either one alone still leaves the reading
at 0.

**1. Sensor matching by name finds nothing (`btop_collect.cpp`)**

`get_gpu_temp_iohid()` identifies GPU sensors by their IOHID `Product` string,
accepting either a `"GPU"` substring or the `PMU TP*g` pattern added in #1601 for
M2 Pro. Newer chips expose neither. Enumerating all 77 thermal sensors on an M4 Pro
gives:

```
PMU tdie1 … PMU tdie14      PMU tdev1 … PMU tdev8
PMU tcal                    NAND CH0 temp        gas gauge battery
```

Nothing matches, the function returns `-1`, and no value is ever pushed.

This filter cannot be rescued by widening it. Loading the GPU (Metal compute) and the
CPU separately and diffing the sensors shows every `tdie` sensor rises *more* under CPU
load than GPU load — `tdie1` moved +19 °C under CPU load versus +9.5 °C under GPU load.
They are SoC die sensors; there is no GPU sensor in the IOHID list to match at any
name pattern. A second source is required.

**2. The SMC decoder cannot read the GPU data type (`smc.cpp`)**

`getSMCTemp()` decodes only `sp78` and returns `-1` for everything else. Apple Silicon
exposes its GPU thermal sensors as 4-byte IEEE floats of type `flt `, so reading those
keys returns `-1` even once you know where to look. This half of the diagnosis is
necoli1822's, from #1653.

## The change

* Decode `flt ` in `getSMCTemp()`, alongside the existing `sp78` path.
* Add `SMCConnection::getTempByKey()` to read an arbitrary sensor by key.
* When `get_gpu_temp_iohid()` finds nothing, fall back to reading the GPU SMC keys.

Design notes:

* **Keys are discovered, not hardcoded.** The set differs per chip, which is exactly
  why the hardcoded name match from #1601 broke on later hardware. The
  `Tg<cluster><sensor>` keyspace is probed once at init and the responders cached.
* **Probing rather than full enumeration.** Walking every key via `#KEY` finds the
  identical 22 sensors but costs ~0.5 s on this machine (3246 keys exposed); probing
  the GPU keyspace costs ~0.07 s. Measured both on an M4 Pro.
* **Discovery is gated behind IOHID failing**, so M1/M2 machines pay nothing.
* **`temp_info` now reflects what was actually found.** If neither source works, the
  GPU temperature is reported as unsupported rather than displayed as a fake 0 — which
  is the user-visible bug in this report.
* **Readings at or below 10 °C are rejected.** Some sensors transiently report a tiny
  uninitialized value right after boot before tracking real die temps. Prior art and
  rationale from mactop (MIT, Carsen Klock), which documents the same behaviour and its
  cause — see `kSiliconMinTempC` in metaspartan/mactop.
* **Averaged, not max**, to stay consistent with how the existing IOHID path reports.

The change is purely additive: 81 insertions, 0 deletions. No existing code path is
modified, so the `sp78` decode and the IOHID path behave exactly as before.

## Testing

Verified on an M4 Pro (Mac16,8, 12-core CPU / 16-core GPU) running macOS 26.5,
built with Apple clang 21.0.0.

Discovery finds 22 GPU sensors:

```
Tg04 Tg05 Tg0d Tg0e Tg0j Tg0k Tg0y Tg0z Tg0K Tg0L Tg0R
Tg0S Tg0X Tg0Y Tg1c Tg1d Tg1k Tg1l Tg1E Tg1F Tg1U Tg1V
```

This is an exact match against an independent full `#KEY` enumeration filtered on
`dataType == 'flt '`, so the probe range is not missing sensors. Values track GPU
load and read plausibly at idle. Before the change this path returned `-1` and the
box showed 0.

I do not have M1/M2/M3/M5 hardware. The IOHID path is untouched and the SMC discovery
only runs when IOHID returns nothing, so M1/M2 behaviour should be unchanged, but
confirmation on other chips would be welcome.

## Not included

#1753 also proposed changing CPU cluster temperature from average to max. That is an
unrelated concern and belongs in its own PR, per the contributing guidelines.

## Credit

* #1653 (necoli1822) — identified both the `flt ` decode gap and the `PMU tdie*`
  naming change on M4/M5. That analysis was never opened as a PR.
* #1753 (char-papadopoulos) — earlier attempt at this bug, closed by its author.
* #1541 (gneitzke) and #1601 (RockBacon9922) — the Apple Silicon GPU and temperature
  paths this builds on.


AwesomenessReborn:
Current release build (sourced via brew)

image

Built binary via https://github.com/AwesomenessReborn/btop/tree/fix/osx-gpu-temp-smc-flt

image

Verified, compared and took inspiration from SMC methods from https://github.com/metaspartan/mactop.

AwesomenessReborn:
https://github.com/aristocratos/btop/issues/1688#issuecomment-5274676776

Independent confirmation on M3, author who originally reported #1688.

Current tested platforms: 
- M4 Pro (mine, screenshots above)
- M3 (@shaozk, linked above)

necoli1822:
Confirmed working on **Apple M5** (macOS 26.5.1, arm64), built from `fix/osx-gpu-temp-smc-flt` @ 7088387 with a plain `make` — no extra deps, no sudo.

**The GPU box reports a temperature, not 0:**

- idle: `32°C` @ 338 MHz, 0.09 W
- under a Metal FMA compute load: `63°C` @ 1578 MHz, 10.8 W, 100% util

Some supporting detail in case it's useful:

**1. The IOHID premise holds on M5.** I dumped every AppleVendor / TemperatureSensor IOHID service — 47 sensors, and **zero** match btop's GPU name filter, so `get_gpu_temp_iohid()` returns -1 and the SMC path is the only source. M5 exposes the same `PMU tdie1..14` / `PMU tdev1..8` / `PMU tcal` / `NAND CH0 temp` set you saw on M4 Pro, plus an additional `PMU2 tdie1..10` / `PMU2 tdev1..5` / `PMU2 tcal` block. Still nothing GPU-named.

**2. Key discovery finds 23 sensors** (vs 22 on your M4 Pro), 84.5 ms over the 496-key probe:

```
Tg04 Tg0C Tg0G Tg0K Tg0O Tg0R Tg0U Tg0X Tg0d Tg0g Tg0j Tg0m Tg0p
Tg12 Tg16 Tg1A Tg1I Tg1M Tg1Y Tg1c Tg1g Tg1o Tg1s
```

Two clusters. Without the `flt ` decode every one of these reads -1, so both halves of the fix are needed here.

**3. They track the GPU, not the SoC.** Sampled across a 45 s Metal burn: 32 °C idle → 45 → 52 → 57 → 61 → 64 °C at the end of the load → 44 → 41 → 39 °C during cooldown. Real GPU die sensors, not the tdie sensors that followed CPU load.

One follow-up, not a blocker: `get_gpu_temp_smc()` opens a fresh `SMCConnection` and closes it on every poll. Timed here that's **8.40 ms per tick** for the 23 keys; holding the connection open across polls instead measures **4.67 ms**. Worth doing — the open/close is most of the cost.

AwesomenessReborn:
Thank you for thorough M5 testing @necoli1822 . Appreciate the connection-reuse measurements and optimization suggestion. Will be looking into it.

The independent platforms that are validated now: 

Current tested platforms:

M4 Pro (mine, screenshots above)
M3 (above first comment above)
M5 (above second comment, with suggestions for improvement.

AwesomenessReborn:
@aristocratos, when you have a chance, would you mind taking a look at this?

The branch is now rebased onto current main. The fix has been validated across M3, M4 Pro, and M5 hardware, including independent M3 and M5 reports.

The M5 testing also identified a possible non-blocking optimization—reusing the SMC connection across polls—which I’m happy to address here or separately if you prefer. 

I'd wish to dive on this issue separately in the future once this issue is closed.

AwesomenessReborn:
@deckstose — could you take a look at this when you have a chance? I originally pinged @aristocratos but I see you've been handling merges recently.

Summary for a quick review:

- **Fixes #1688, resolves #1653** — GPU temp shows `0` on M3 and later Apple Silicon. Users on #1653 have been asking for an official fix since May.
- **Purely additive**: 81 insertions, 0 deletions. The existing `sp78` decode and IOHID path are untouched; the SMC fallback only runs when IOHID finds no GPU sensor, so M1/M2 behaviour is unchanged.
- **Independently confirmed on three chips**: M3 ([shaozk, #1688](https://github.com/aristocratos/btop/issues/1688#issuecomment-5274676776)), M4 Pro (mine, screenshots above), M5 ([necoli1822, above](https://github.com/aristocratos/btop/pull/1788#issuecomment-5274780373), with full sensor dump and load/cooldown curve).
- Rebased onto current `main` today (no conflicts, patch unchanged — `range-diff` is `=`); CI green on macOS (make + CMake) under the newly-combined workflows.

One open question: necoli1822's M5 testing suggested reusing the SMC connection across polls instead of reopening it. It's a small, non-blocking optimization — happy to fold it into this PR or send it separately, whichever you prefer.

Happy to adjust anything else.

PR #1824, opened 2026-09-07
fix: resource inflation on filter input

### LINKED ISSUE #1822: [BUG] Proc tree + aggregate: process resources inflate on every filter change
**Describe the bug**

With proc_tree and proc_aggregate enabled, every modification of filter string adds the children's resource values to the parent's totals  (it seems so because only processes with children grow). CPU%, MEM and THREADS grow per keystroke and print nonsense, for example systemd at hundreds of terabytes of memory.

Removing a character counts too.
The fields are reset on the next interval update, so it looks like a minor issue.

**To Reproduce**
1. Enable proc_tree and proc_aggregate.
2. Hit key 'f', then type. Watch CPU%/MEM/THREADS on tree parents grow per keystroke. Alternating with backspace keeps processes visible, increasing update interval buys more time for a screenshot.

**Expected behavior**

Filtering recomputes the values or doesn't change them at all.

**Screenshots**
(CPU is at 0 in this screenshot because it was genuinely 0 across all processes - nothing to sum; when values are > 0.0, it explodes too)
Image

 - btop++ version: 1.4.7
 - Binary: 
Compiled with: g++ (15.2.1)
Configured with: make STATIC= GPU_SUPPORT=true RSMI_STATIC= ;
installed from pacman - cachyos-extra-znver4 and cachyos-extra-v3
 - Architecture: x86_64
 - Platform: Linux
 - Kernel: 7.2.3-1-cachyos && 7.2.2-1-cachyos (verified on two machines)
 - Terminal used: ghostty && tty
 - Font used: NotoSansMono-Regular.ttf

**Additional context**

Reproduced on two machines: Ryzen 9950X3D/Radeon and Core Ultra 265H/Intel Arc/NVidia, both CachyOS.
ONLY reproducible in case (proc_tree && proc_aggregate) == true.

src/btop_shared.cpp#L252-L257 seems to be the culprit. NOT TESTED, this claim was generated by an LLM.

**Related issues**

#817 
#526

ychwch:
closes #1822


deckstose (approved):
Thanks!

PR #1825, opened 2026-09-08
fix(docs): fixing wrong singular in disks_filter examples

outofrange:
I just ran into this, wondering why my config copy/pasted from README examples doesn't work - it's `disks_filter`, not `disk_filter` :)


deckstose (approved):
Thank you 💯

PR #1845, opened 2026-09-20
[AI generated] illumos/Solaris platform support

agorouvein-rgb:
[AI generated] — drafted with Claude, reviewed, tested, and debugged by me on real hardware/VMs at every step. See the "How this was built" note at the bottom for full transparency.

## Summary

Adds a new `src/sunos/` backend (originally `src/openindiana/`, renamed since the backend is generic across illumos distributions and Solaris, not OpenIndiana-specific) implementing `Shared`, `Cpu`, `Mem`, `Net`, and `Proc` data collection against illumos-native APIs, with optional support for Solaris 10 x86. This backend started as a copy of `src/freebsd/`, with every FreeBSD-specific mechanism (`kvm(3)`, `sysctl(3)`, `devstat(3)`) replaced by illumos equivalents:

- **CPU stats**: `kstat(3KSTAT)` `cpu_stat` per-core counters, discovered by walking the live kstat chain rather than probing instance numbers 0..N-1 (CPU ids are not guaranteed contiguous or zero-based — confirmed on SPARC, where a 64-CPU system reported ids 64-127)
- **Memory**: `kstat unix:system_pages` + ZFS ARC size (`kstat zfs:arcstats`), used to derive used/available/cached the way Linux/FreeBSD treat page cache
- **Swap**: native `swapctl(2)` — no `kvm` handle required
- **Filesystem enumeration**: `getmntent(3C)` reading `/etc/mnttab`, filtering out pseudo-mounts (`vmblock`, `vmhgfs`, etc.)
- **Disk I/O**: `kstat_io_t` enumeration by `ks_class=="disk"`
- **Network**: `kstat` per-link `rbytes64`/`obytes64`, looked up directly by interface name (module `"link"`, class `"net"`). Interface enumeration uses `SIOCGLIFCONF` rather than `getifaddrs()`, which doesn't exist on Solaris 10.
- **Process enumeration**: `/proc/<pid>/psinfo` (`psinfo_t`), replacing `kvm_getprocs()`/`kvm_getargv()`

Also includes portability fixes to shared, non-platform files:

- `src/btop.cpp`: extends the existing `pthread_join()` fallback (already used for `__APPLE__`/`__OpenBSD__`/`__NetBSD__`) to `__sun`, since `pthread_timedjoin_np()` is a GNU/glibc extension unavailable on illumos.
- `src/btop.cpp`: fixes a crash on terminal resize. `signal()` has one-shot (System V) semantics on illumos/Solaris — handlers reset to default after their first delivery. btop uses `SIGWINCH`/`SIGUSR1` to interrupt its input loop, so a second resize (or a second Ctrl-C/Ctrl-Z) during a session could kill the process outright, with no core dump or log entry. Signal handlers on this platform are now installed via `sigaction()` for persistent, BSD-style semantics. This affected every illumos/Solaris build, not just Solaris 10.
- `src/sunos/btop_collect.cpp`: the file-local `namespace fs = std::filesystem;` alias is renamed to `btop_fs`, since native Solaris's `<sys/buf.h>` (pulled in transitively by some of the headers this backend uses) declares its own `fs` at file scope and collides with it. illumos doesn't hit this, which is why it wasn't caught earlier.
- `Makefile`: adds a `sunos` `PLATFORM_DIR` block matching the existing per-platform pattern, linking `-lkstat -lsocket -lnsl`. Source-file discovery now uses GNU Make's built-in `$(wildcard ...)` instead of shelling out to `find -maxdepth`, which native Solaris's `find` doesn't support (this also simplifies the shared, non-platform-specific part of the build). Thread count detection now tries `psrinfo` first, falling back to `getconf NPROCESSORS_ONLN` only if that's empty, since the latter alone has been reported as unrecognized on at least one native Solaris configuration (silently serializing the build). `-static-libstdc++` is no longer forced by default — Oracle Solaris ships no static libstdc++ at all — and is now opt-in via `S10=true` (see below), which is the only case that actually needs it.

### Solaris 10 x86 support (optional, `S10=true`)

Solaris 10 has no C++23-capable compiler, so binaries are cross-built on an illumos system with a modern GCC and run on Solaris 10 itself. That requires a small number of libc functions Solaris 10 lacks (mostly POSIX 2008 locale functions, plus a few small helpers like `clock_gettime`/`nanosleep`, which live in `librt` rather than libc there). These are provided by an optional compatibility shim, `src/sunos/s10compat.c`, compiled and linked in only with `S10=true`; the default `sunos` build doesn't touch it. It's inert unless explicitly requested.

## Known limitations

- ZFS per-dataset disk I/O is not separated; all physical disk I/O is attributed to the root (`/`) mountpoint, since illumos ZFS mounts report the dataset name rather than a physical device path in `/etc/mnttab`. True per-dataset attribution likely requires illumos's per-dataset ZFS kstats, not yet investigated.
- `pr_psargs` truncates long command lines to `PRARGSZ` (~80 chars); full argv reconstruction via `libproc`'s `proc_arg_grab()` is a possible follow-up.
- CPU temperature, frequency, and battery reporting are stubbed (return empty/false/no-op), not yet implemented.
- `get_cpuName()` currently returns an empty string.
- Mouse input requires a terminal that reports the SGR (1006) mouse protocol; older terminals (including some shipped with Solaris 10) only speak the legacy X10 protocol, so mouse clicks/scroll won't register there. Keyboard navigation is unaffected.

## Testing

Built and tested in a VirtualBox VM running OpenIndiana Hipster 2026.04, GCC 14.3.0, and on bare metal (Acer Aspire V5-571, same OI release, using the built-in Realtek RTL8111 Ethernet as `rge0`). Bare-metal testing included running btop continuously in a MATE Terminal session while connecting in over SSH from a separate OI VM and running a package install (`pkg install fastfetch`), exercising real CPU, disk, and network activity end-to-end rather than synthetic load. (Aside, for anyone else with this model: the built-in Atheros AR9462 wifi is not supported by illumos's `arn` driver, which only covers AR9280/9281/9285 — Ethernet works fine.)

Solaris 10 (1/13, x86, 64-bit kernel) VMWare VM tested separately, cross-built from the OpenIndiana host per the `S10=true` process above.

All of Peter Tribble's review feedback (thank you again) has been addressed and tested, including a retest on his own hardware:
- CPU panel on SPARC (contiguous-id assumption) — fixed by walking the live kstat chain instead of probing instance numbers; confirmed working by Peter on a native, natively-compiled Solaris 11.4 SPARC system (T8-1).
- `fs`/`sys/buf.h` collision on native Solaris — fixed, renamed to `btop_fs`; confirmed by the same native Solaris 11.4 build.
- `find`/`getconf NPROCESSORS_ONLN` build portability — fixed; confirmed by the same native Solaris 11.4 build.
- `-static-libstdc++` forced unconditionally — now opt-in, off by default, only added for the Solaris 10 cross-build (`S10=true`); dropping it by default is what let the native Solaris 11.4 build succeed.

Still not tested: other illumos distributions (SmartOS, etc.), and native (non-cross-built) compilation on Solaris 10, which has no C++23-capable compiler and so isn't expected to support a native build at all.

## How this was built

I don't have a C++ background, but I am a software tester with 30+ years of experience in the software industry. I worked through this interactively with Claude (Anthropic's AI) over several sessions: I ran every command and test on my actual VMs and hardware, verified every kstat/swapctl/psinfo field against real system output before any of it went into the actual source file, and traced/fixed real bugs along the way (e.g. the ZFS-dataset I/O attribution limitation above was a design decision I made once we hit it, not something guessed at; the resize crash was found and root-caused by reproducing it with `truss`, tracing it to `signal()`'s platform-specific semantics, and confirming the fix against the actual signal delivery behaviour). Happy to answer questions about any part of the implementation, and glad to keep iterating based on review feedback.

OpenIndiana-btop
OpenIndiana-fastfetch
image


mebenn:
btop_on_oi

agorouvein-rgb:
> btop_on_oi

Thanks for trying it out — good to see it working cleanly on your setup too!

mebenn:
Hey, really nice'; cheers. Compiled directly using gmake. I've got OI running on Linux using VirtualBox .
I've noticed a few niggles though. The CPU info box does not display info on the CPU Hardware. I see you've got the same problem.  (Use psrinfo -vp). I'm sure which C++ methods are used by Illumos to retrieve this info and how this uis implemented in btop to do this for Linux. Probably require an #ifdef SunOS or something.
But this is really small stuff that can be fixed later.
If you examine a process by pressing return, the IO/R and IO/W are empty. Have you got the same behaviour?
The number of threads (called LWP on Illumos) seems correct (prstat -p <pid>). I've only checked a few.

I've also tried running it though valgrind; but had to stop as it took for ever. But, in general, btop is very responsive which is good.

Would be nice to test it on bare-metal, i.e, ssh on to a sparc machine.  I'll do further testing tomorrow evening and check memory and have a look at the code.

agorouvein-rgb:
Hey @mebenn 

Thanks for the detailed testing, especially catching the IO/R and IO/W issue — that one wasn't actually documented anywhere, so genuinely good find.

Both are fixed and pushed:

**CPU hardware name**: now pulled from the `cpu_info` kstat (`brand` on x86, falling back to `implementation` on SPARC/wherever `brand` isn't present), then cleaned up the same way upstream's Linux backend does — strips the boilerplate (`Intel(R)`, `Core(TM)`, `CPU @ x.xxGHz`, etc.) down to just the model, e.g. `i7-6700K`. One limitation: I don't have AMD or SPARC hardware to verify against, so while it should work reasonably on those too, the exact cosmetic output (spacing, leftover artifacts on unusual brand-string shapes) is untested there — low risk, but flagging it.

**IO/R and IO/W**: these were dead code left over from a macOS-derived stub, now implemented via `/proc/<pid>/usage`. Initially I used `pr_inblk`/`pr_oublk` (the seemingly obvious fields), but real testing against sustained raw-device I/O showed they stay at 0 regardless of actual volume — turns out this is a long-documented Solaris/illumos quirk, not specific to this port (Brendan Gregg wrote about it back in 2005, and `psutil`'s own Solaris backend avoids those fields for the same reason). Switched to `pr_ioch` instead, which does track correctly. The real limitation: `pr_ioch` is a single combined read+write counter, with no per-direction split available on this platform. IO/R now shows that combined total; IO/W is left blank (not `0 Byte`) to signal "not available" rather than falsely claiming zero write activity.

Both tested and confirmed working on OpenIndiana. More broadly (not specific to this latest commit, but worth mentioning for context): the port has also been confirmed on bare metal — an old Acer laptop running current OI — and independently by a mailing-list contributor who compiled and ran it on a T8-1 with 64 SPARC cores.

